PR #11, opened 2015-05-06
Rearrange Lapack to be like BLAS. Implement cholesky decomposition

btracey:


kortschak:
You'll need to edit .travis.yaml to point to the correct location at line 42 `s/clapack/cgo/`.

lapack64/lapack64.go
```diff
@@ -0,0 +1,42 @@
+// Copyright ©2015 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// lapack64 provides a set of convenient wrapper functions for LAPACK calls, as
```
kortschak: Prefix with "Package".

native/dpotf2.go
```diff
@@ -0,0 +1,66 @@
...
+
+	"github.com/gonum/blas"
+	"github.com/gonum/blas/blas64"
+)
+
+// Dpotrf computes the cholesky decomposition of the symmetric positive definite
+// matrix a. If ul == blas.Upper, then a is stored as an upper-triangular matrix,
+// and a = U^T U is stored in place into a. If ul == blas.Lower, then a = L L^T
+// is computed and stored in-place into a. This is the unblocked version of the
+// algorithm. Dpotrf returns 0 if the decomposition is successful, otherwise
```
kortschak: I think we should be returning `ok bool` rather than a status integer.

lapack64/lapack64.go
```diff
@@ -0,0 +1,42 @@
...
+// The default implementation is native.Implementation.
+func Use(l lapack.Float64) {
+	lapack64 = l
+}
+
+// Dpotrf computes the cholesky factorization of a.
+//  A = U^T * U if ul == blas.Upper
+//  A = L * L^T if ul == blas.Lower
+// On entry, a is a symmetric matrix, and at conclusion it is the U or L matrix.
+func Dpotrf(ul blas.Uplo, n int, a []float64, lda int) int {
```
kortschak: We are in a position to us the blas64 types so that this becomes `Dpotrf(a blas64.Symmetric) (blas64.Triangular, error)`*. The documentation then only needs to say that the backing data is used in the returned value.

\* see comment about return value below.

native/dpotf2.go
btracey: We could make this change in lapack64, but shouldn't the actual "native" try to mimic the existing lapack as much as possible? I suppose the cblas wrapper could make the same transformations, but we would need to keep them in sync.

lapack64/lapack64.go
btracey: Yes, it should be Potrf.

I'm not sure about your suggestion for return types. Thinking about how I would want to use this in mat64
````
func (t *TriDense) Chol(a Symmetric) {
     // front matter
    // Copy data from SymDense into t
     lapack64.Potrf(t.mat.Uplo, t.mat.N, t.mat.Data, t.mat.Stride)
}
// And done because the data is now in-place into t
````
If we take a Symmetric in, we would need to copy the data into t, then construct a new blas64.Symmetric. This isn't costly, but I don't think it's any more clear for the code reader who needs to infer that the data is shared between the symmetric and the triangular.

File: native/dpotf2.go (hunk 2 above)
Comment (kortschak) [inline native/dpotf2.go]: We make essentially the same degree of changes in cgo/blas with the rot procedures. I don't see an issue with that.

This is pretty much equivalent to how the standard library wraps ERRNO.

File: lapack64/lapack64.go (hunk 3 above)
Comment (kortschak) [inline lapack64/lapack64.go]: The current Cholesky method only takes a SymDense, why do,we,need to accept a Symmetric?

Comment (btracey) [inline lapack64/lapack64.go]: This ties in with the discussion on gonum-dev. One can compute the Cholesky decomposition of a (positive definite) symmetric matrix. Why would we only support SymDense and not all Symmetric? We support arbitrary types for other operations.

Comment (kortschak) [inline lapack64/lapack64.go]: Mainly because it's easy to say `t.Cholesky(SymDenseCopyOf(s))` if needed.

File: lapack64/lapack64.go (hunk 1 above)
Comment (btracey) [inline lapack64/lapack64.go]: Done.

File: native/dpotf2.go (hunk 2 above)
Comment (btracey) [inline native/dpotf2.go]: Making it ok bool

**Comment by btracey** [comment]
PTAL

**Comment by kortschak** [comment]
.travis.yaml needs to be updated still.

File: native/dpotf2.go (hunk 2 above)
Comment (kortschak) [inline native/dpotf2.go]: Sorry to be difficult on this, but is the indormation we are discarding here likely to be useful to someone?

Making it an error allows that to continue to be used if needed.

**Comment by btracey** [comment]
Travis was updated to swap clapack for cgo. It's still failing (obviously). Before it was giving a "pushd clapack" failure (or something like that). Now it's failing with 
````
The command "go test -x -v ./..." exited with 2.
0.86s
$ diff <(gofmt -d .) <("")
/home/travis/build.sh: line 41: : command not found
````
There are a bunch of lapack.Job warnings as well (not sure what they mean). Is the problem that there are no go files in the cgo folder (just headers and a perl script?). Those have been there for a while, but I've never tried to integrate a c lapack installation.

File: native/dpotf2.go (hunk 2 above)
Comment (btracey) [inline native/dpotf2.go]: > On May 7, 2015, at 2:05 PM, Dan Kortschak <notifications@github.com> wrote:
> 
> In native/dpotf2.go <https://github.com/gonum/lapack/pull/11#discussion_r29895939>:
> 
> > @@ -0,0 +1,66 @@
> > +package native
> > +
> > +import (
> > +	"math"
> > +
> > +	"github.com/gonum/blas"
> > +	"github.com/gonum/blas/blas64"
> > +)
> > +
> > +// Dpotrf computes the cholesky decomposition of the symmetric positive definite
> > +// matrix a. If ul == blas.Upper, then a is stored as an upper-triangular matrix,
> > +// and a = U^T U is stored in place into a. If ul == blas.Lower, then a = L L^T
> > +// is computed and stored in-place into a. This is the unblocked version of the
> > +// algorithm. Dpotrf returns 0 if the decomposition is successful, otherwise
> Sorry to be difficult on this, but is the indormation we are discarding here likely to be useful to someone? Making it an error allows that to continue to be used if needed.

Personally, I don’t know for sure, but I don’t really see how. If A is positive definite, then all of the eigenvalues are > 0. A not being positive definite implies one of them is not. However, you can reorder the rows and such of a matrix without changing the eigenvalues (only the order). Thus, it seems to me that the info can’t give you the eigenvalue that’s negative, as there isn’t any real order to the eigenvaules, they just are. Possibly you could “augment” the diagonal of the matrix to make it positive definite, but again, it’s not clear to me why you’d change the first entry that failed. The value on the diagonal is already different from when it started. The integer is also weird in this case, because it’s one-indexed in normal lapack.

I did some quick google searching and couldn’t find an example where the returned integer was used. For example, https://stat.duke.edu/~cr173/Sta523_Fa14/blas_lapack.html <https://stat.duke.edu/~cr173/Sta523_Fa14/blas_lapack.html>, doesn’t even say that that’s the return.

**Comment by kortschak** [comment]
I'll have a look later today.

File: native/dpotf2.go (hunk 2 above)
Comment (kortschak) [inline native/dpotf2.go]: I guess I'm wondering if it would be useful for debugging.

File: lapack.go [hunk 4]
```diff
@@ -1,39 +1,9 @@
 package lapack
 
-import (
-	"github.com/gonum/blas"
-)
-
-const None = 'N'
-
-type Job byte
```
kortschak: This was needed by cgo/lapack.

lapack.go
```diff
@@ -1,39 +1,9 @@
...
-
-const (
-	All       (Job) = 'A'
-	Slim      (Job) = 'S'
-	Overwrite (Job) = 'O'
-)
-
-type CompSV byte
```
kortschak: Same here.

kortschak:
OK, figured out why it wasn't building: There are types defined in lapack/lapack.go that are required by lapack/cgo/lapack.go

native/dpotf2.go
btracey: Maybe? If I had a matrix I thought was positive definite, and Chol told me it wasn’t, my next step would be to look at the eigenvalues to see if they’re all positive. If they aren’t and I wanted to dig further, I’m not sure how to do that other than to intrusively modify the go cholesky routine to do some print statements. I’m not sure what I’d learn from the number itself.

kortschak: OK. We'll leave as bool.

btracey: I don’t feel particularly strongly, except that I do think the mat64 version has a nice API. I can put it back if you prefer.

kortschak: No, bool is fine - I prefer it over int. I'm just looking through the genLapack to figure out how to get the return to be a bool.

btracey:
Okay, cgo now compiles on my machine. I hacked the perl script in a way that's not pretty, but we'll need to update the script anyway at some point to use go generate, and have the right return types, etc.

kortschak:
Are all the in returns in LAPACK 0=success !0=failure?

kortschak:
Changes for genLapack.pl

```
diff --git a/cgo/genLapack.pl b/cgo/genLapack.pl
index 8c15042..07de5ef 100755
--- a/cgo/genLapack.pl
+++ b/cgo/genLapack.pl
@@ -57,12 +57,14 @@ const (
        rowMajor order = 101 + iota
        colMajor
 )
- 
+
 func init() {
          _ = lapack.Complex128(Lapack{})
          _ = lapack.Float64(Lapack{})
 }
 
+func isZero(ret C.int) bool { return ret == 0 }
+
 EOH
 
 $/ = undef;
@@ -110,11 +112,18 @@ our %typeConv = (
        "LAPACK_C_SELECT2" => "Select2Complex64",
        "LAPACK_Z_SELECT1" => "Select1Complex128",
        "LAPACK_Z_SELECT2" => "Select2Complex128",
-       "void" => ""
+       "void" => "",
+
+       "lapack_int_return_type" => "bool",
+       "lapack_int_return" => "isZero",
+       "float_return_type" => "float32",
+       "float_return" => "float32",
+       "double_return_type" => "float64",
+       "double_return" => "float64"
 );
 
 foreach my $line (@lines) {
-       process($line); 
+       process($line);
 }
 
 close($golapack);
@@ -162,14 +171,15 @@ sub processProto {
                $gofunc = ucfirst $func;
        }
 
-       my $GoRet = $typeConv{$ret};
+       my $GoRet = $typeConv{$ret."_return"};
+       my $GoRetType = $typeConv{$ret."_return_type"};
        my $complexType = $func;
        $complexType =~ s/.*_[isd]?([zc]).*/$1/;
        my ($params,$bp) = processParamToGo($func, $paramList, $complexType);
        if ($params eq "") {
                return
        }
-       print $golapack "func (Lapack) ".$gofunc."(".$params.") ".$GoRet."{\n";
+       print $golapack "func (Lapack) ".$gofunc."(".$params.") ".$GoRetType."{\n";
        print $golapack "\t";
        if ($ret ne 'void') {
                print $golapack "\n".$bp."\n"."return ".$GoRet."(";
```

btracey:
Sorry, how do I apply that patch? I reverted back to the original genLapack.pl. I copied the text in the block above, saved it in cgo as patch.diff, and ran ````git apply -v patch.diff````. I get the following output:

````
brendan:~/Documents/mygo/src/github.com/gonum/lapack/cgo$ git apply -v patch.diff 
Checking patch cgo/genLapack.pl...
error: while searching for:
       rowMajor order = 101 + iota
       colMajor
)
 
func init() {
         _ = lapack.Complex128(Lapack{})
         _ = lapack.Float64(Lapack{})
}

EOH

$/ = undef;

error: patch failed: cgo/genLapack.pl:57
error: cgo/genLapack.pl: patch does not apply
````
I don't understand the "error searching for" because those lines are in the file...

btracey:
I'm not familiar with all of Lapack (learning as I go), but so far all of the returns are zero for success.

btracey:
Okay, when I say "restored", I actually mean I copied and pasted in the old file. I wonder if that's part of the problem (where git is using a file index rather than just the characters?)

kortschak:
Yeah, that will lead to a conflict.

kortschak:
Those failures aren't real AFAICS. Travis is getting pretty crappy recently.

kortschak:
This last failure is due to removal of the test-coverage.sh file.

btracey:
I thought we had removed all of those files. Do I remember incorrectly?

kortschak:
Yes, you have. You should not have.

The .travis.yaml calls the coverage script. This fails with 127 (mot found).

btracey:
There we go. Sorry for making the process difficult.

kortschak:
No worries, I'm learning about travis in the process - this is a good thing.

btracey:
PTAL

kortschak:
LGTM

PR #13, opened 2015-06-06
cleanup .travis.yml and associated files

jonlawlor:
As explained in
https://groups.google.com/d/msg/gonum-dev/GKJjAtDxR9I/n0FD7WvvN5AJ

Removed notifications and go tip, simplified code, added some comments,
put all travis related code (except for .travis.yml) into .travis/

A note for discussion: I named the build that relies on gonum/blas "gonum," where previously we have called it native, because native is typically used to describe something that is processor dependent and not portable, while the gonum/blas library is the opposite.  I also fiddled around a little with getting the ATLAS build to work, but I was having issues linking the lapack functions so its install script is not complete.  It is not enabled for builds, although it would be nice because it takes much less time to install.

PTAL

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,60 +1,36 @@
+sudo: required
+
 language: go
 
 env:
-  global:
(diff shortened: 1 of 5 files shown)


kortschak:
Nice clean up.

LGTM

PR #14, opened 2015-06-09
Put lapack.go in gitignore

joshk0:
Otherwise, this directory is always dirty after a Go build when
gonum/lapack is consumed as a submodule.

### DIFF
--- a/cgo/.gitignore
+++ b/cgo/.gitignore
@@ -1,0 +1,1 @@
+lapack.go


kortschak:
NOT LGTM

Code generation is not a client-side operation in the Go ecosystem.

joshk0:
So what's the point of making the genLapack.pl script available to users, and advertising its presence in README.md?

kortschak:
When the package is mature, a reasonable default will be committed. It the moment, the package is in flux; @btracey is doing work to implement basic Go LAPACK functions.

If we add lapack.go to .gitignore, we lose tracking of both github.com/gonum/lapack/lapack.go (not generated and needed for the Go implementations that do exist) and we don't track github.com/gonum/lapack/cgo/lapack.go (which is what you are aiming at). When it comes time to commit the latter we then spend time trying to figure out why it didn't get committed. Those are both significant downsides for us.

The benefit that *you* get is that the repo seems to be in a clean state. Why is that important to you?

You are correct, we do commit the genLapack.pl code. The generation code is there primarily for our use, but since it's there it makes sense to also have instructions on how to use it for users that do not want to use OpenBLAS (which is what we will likely default to). I would personally rather not have users do this.

joshk0:
The .gitignore is in the cgo/ directory, so it is scoped only to that directory. It does not affect files named lapack.go which live in other directories.

Our Go project has our dependencies pinned as submodules, including gonum/lapack. During the build we must run genLapack.pl to get the symbols we need available in Go. This creates cgo/lapack.go. In development, this means that if we are working on files in the outer repository, the gonum/lapack submodule is always marked as dirty in git status because of the presence of cgo/lapack.go.

Therefore, this change is merely to convince our tooling to play nice by not making the submodule tree dirty every time we do a build. So not a functional problem, but regarding day-to-day development with it, it would be nice to fix.

kortschak:
Thank you for explaining more. It may be worthwhile use then doing the code generation now. @btracey are you OK with committing that?

btracey:
I haven't looked at the cgo side of lapack. Is it hard to just provide the lapack.go call like we do in blas/cgo, or is that too much work for something we'll be fixing anyway soon?

@joshk0 Which lapack functions do you use?

joshk0:
ssyev is the only one I am using right now.

btracey:
Interesting. I didn't realize there were problems that single precision eigenvalues are still really useful. 

Dsyevr was my next target for native, but there are a lot of steps to get there.

joshk0:
I'm using it for [this](https://en.wikipedia.org/wiki/Graph_partition#Spectral_partitioning_and_spectral_bisection) and find it to be fairly useful for my needs since I only care about the sign of the eigenvalue. But I'm also a novice at linear algebra, so you tell me whether i should switch to double precision!

btracey:
Are you making a Go version of METIS? That would be exciting.

joshk0:
You're right, looks like I could bind to METIS instead, and they say that their method is better than spectral partitioning...

kortschak:
>Is it hard to just provide the lapack.go call like we do in blas/cgo, or is that too much work for something we'll be fixing anyway soon?

That's essentially what I was suggesting. I think that is pretty much what we had before.

@joshk0 What are you building?

jgraettinger:
This is being applied to a graph that's continuously built and churned. We index connected components of the graph together, and for performance reasons need to be able to bound how large a clique can potentially get.

It's perhaps an odd use case because we're not looking to partitioning to necessarily *understand* the graph, or even because we believe it *should* be partitioned, but as a principled way of making a least-bad cut at a given moment, when we decide that a clique is over-threshold and needs to be broken up. Thus spectral partitioning seems to make sense, since we're only ever making one cut (as after that, it's under-threshold).

I'm not familiar with METIS. It certainly does look interesting, but would it be suited to an online application of this kind?

btracey:
The use of METIS around me is to split a computational mesh into roughly equal partitions to then apply MPI-style parallelism with each partition on one core.

PR #15, opened 2015-06-25
Adddlarf

btracey:


btracey:
This routine is used as a kernel for many other Lapack routines. It works toward linear solve and symmetric eigenvalues. 

I thought about changing iladlr and iladlc to return the 1-indexed row which would make the code cleaner. However, according to the call graph at http://www.netlib.org/lapack/explore-html/d5/d87/iladlr_8f.html#a98b05d45e20605b554f65db26470d079 , this routine is only used by dlarf. For the single case of improvement, it didn't seem worth the confusion in behavior difference.

lapack.go
```diff
@@ -1,3 +1,7 @@
+// Copyright ©2015 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
 package lapack
```
kortschak: Do we want to add a package-level doc comment at this stage?

native/dlarf.go
```diff
@@ -0,0 +1,70 @@
...
+// license that can be found in the LICENSE file.
+
+package native
+
+import (
+	"github.com/gonum/blas"
+	"github.com/gonum/blas/blas64"
+)
+
+// Dlarf applies an elementary reflector to a general rectangular matrix c
```
kortschak: s/$/./

native/dlarf.go
```diff
@@ -0,0 +1,70 @@
...
+// This computes
+//  c = h * c if side == Left
+//  c = c * h if side == right
+// where
+//  h = 1 - tau * v * v^T
+// and c is an m * n matrix.
+//
+// Work is pre-allocated memory of size at least m if side == Left and at least
+// n if side == Right.
```
kortschak: Note that Dlarf will panic if this is not satisfied.

native/dlarf.go
```diff
@@ -0,0 +1,70 @@
...
+		var i int
+		if applyleft {
+			lastv = m
+		} else {
+			lastv = n
+		}
+		if incv > 0 {
+			i = (lastv - 1) * incv
+		}
+		// Look for the last non-zero row in v
```
kortschak: s/$/./

native/dlarf.go
```diff
@@ -0,0 +1,70 @@
...
+		for lastv >= 0 && v[i] == 0 {
+			lastv--
+			i -= incv
+		}
+		if applyleft {
+			// Scan for the last non-zero column in C(0:lastv, :)
```
kortschak: Make this not look like a smiley face?

native/iladlc.go
```diff
@@ -0,0 +1,29 @@
...
+func (Implementation) Iladlc(m, n int, a []float64, lda int) int {
+	if n == 0 {
+		return n - 1
+	}
+	// Test common case where corner is non-zero
+	if a[n-1] != 0 || a[(m-1)*lda+(n-1)] != 0 {
+		return n - 1
+	}
+
+	// Scan each row tracking the highest column seen
```
kortschak: s/$/./

native/iladlc.go
```diff
@@ -0,0 +1,29 @@
...
+	highest := -1
+	for i := 0; i < m; i++ {
+		for j := n - 1; j >= 0; j-- {
+			if a[i*lda+j] != 0 {
+				highest = max(highest, j)
```
kortschak: return max(highest, j)

native/iladlc.go
```diff
@@ -0,0 +1,29 @@
...
+				break
+			}
+		}
+	}
+	return highest
```
kortschak: return -1

kortschak:
LGTM with very minor comments.

lapack.go
btracey: Yes. I think that's a good idea. I'll make a new commit for it though.

native/dlarf.go
btracey: It is matlab notation, but it is also the notation in the FORTRAN comments. I'll change them to brackets, as that is also more go-like.

PR #17, opened 2015-07-03
Working implementation of blocked QR

btracey:


btracey:
NOT READY FOR SUBMISSION. Lots of comments to improve and small things to clean up.

Sorry for the large PR. It took a while to understand what was actually going on with the individual routines.

btracey:
Okay. Comments fixed, and tests fully added. 

PTAL.

btracey:
Don't review this yet. I have been working on some new functions and the tests will be restructured a bit. At that point we can find the best way to merge the changes.

native/dgeqr2.go
```diff
@@ -0,0 +1,50 @@
...
+// the
+//  v[j] = 0           j < i
+//  v[j] = i           j == i
+//  v[j] = a[i*lda+j]  j > i
+// and computing h_i = I - tau[i] * v * v^T.
+//
+// The orthonormal matrix Q can be constucted from a product of these elementary
+// reflectors, Q = H(1)H(2) ... H(k), where k = min(m,n).
+//
+// Work is a temporary storage matrix of at least len(n).
```
kortschak: Dgeqr will panic if work is too short.

btracey:
Alright. Found some bugs. Full implementation of Dgels! Working on better documentation/testing now.

btracey:
Ok! Comments fixed and tests added. The tests will fail right now because of gonum/blas#130. 

PTAL

I know this is a lot. Let me know if I should split it up. Adding the extra functionality allowed me to better understand what was going on and provide more cross tests.

lapack.go
```diff
@@ -30,3 +30,34 @@ type Float64 interface {
 }
 
 type Complex128 interface{}
+
+// Direct specifies the direction of the multiplication for the Householder matrix.
+type Direct byte
+
+const (
+	Forward  (Direct) = 'F'
```
kortschak: Don't need the parentheses on types (also below).

kortschak:
I'm not entirely sure where to start with this. Can you direct me?

btracey:
Start with Dgeqr2. This will lead you to Dlarf and Dlarfg.  Once done, you can move to Dgeqrf, which does the same thing as Dgeqr2, but is the blocked version, which will lead you to Dlarfb and Dlarft. With those in place, Dgelq2 and Dgelqf are the same idea. Then go to Dormqr and Dormlq which are a small step after that. Then look at Dlange, which will seem like a bit of a non-sequitur. However, this puts all the major pieces in place to examine Dgels. This will lead you to a few more simple auxiliary routines .

native/general.go
```diff
@@ -33,3 +76,30 @@ func max(a, b int) int {
 	}
 	return b
 }
+
+func sign(a, b float64) float64 {
```
kortschak: Use math.Copysign.

native/dgelq2.go
```diff
@@ -0,0 +1,40 @@
...
+	}
+	if len(work) < m {
+		panic(badWork)
+	}
+	for i := 0; i < k; i++ {
+		tau[i] = impl.Dlarfg(n-i, &a[i*lda+i], a[i*lda+min(i+1, n-1):], 1)
+		if i < m-1 {
+			aii := a[i*lda+i]
+			a[i*lda+i] = 1
+			impl.Dlarf(blas.Right, m-i-1, n-i, a[i*lda+i:], 1, tau[i], a[(i+1)*lda+i:], lda, work)
```
kortschak: I'd prefer these on conceptually blocked lines where possible (also all below).

btracey: Do you mean the inputs to Dlarf, that is, add newlines in between the function arguments where it makes sense?

native/general.go
btracey: Done.

Do you anything about the history of this function? It's useful here, but it seems like a weird thing to have in the math package (and yet a bunch of languages seem to have it easily accessible). Is this a lot more useful than I think?

native/dlange.go
```diff
@@ -0,0 +1,76 @@
...
+	"github.com/gonum/lapack"
+)
+
+// Dlange computes the matrix norm of the general m x n matrix a. The input norm
+// specifies the norm computed.
+//  lapack.MaxAbs: the absolute value of the maximum element.
+//  lapack.MaxColumnSum: the maximum column sum of the absolute values of the entries.
+//  lapack.MaxRowSum: the maximum row sum of the absolute values of the entries.
+//  lapack.Frobenius: the square root of the sum of the squares of the entries.
+// If norm == lapack.MaxColumnSum, work must be of length n. There are no restrictions
```
kortschak: Note that this will panic.

native/dlarfb.go
```diff
@@ -0,0 +1,341 @@
...
+//  c = c * h if side == Right and trans == NoTrans
+//  c = h^T * c if side == Left and trans == Trans
+//  c = c * h^t if side == Right and trans == NoTrans
+// h is a product of elementary reflectors. direct sets the direction of multiplication
+//  h = h_1 * h_2 * ... * h_k if direct == Forward
+//  h = h_k * h_k-1 * ... * h_1 if direct == Backward
+// The combination of direct and store defines the orientation of the elementary
+// reflectors. In all cases the ones on the diagonal are implicitly represented.
+//
+//  If direct == lapack.Forward and store == lapack.Columnwise
```
kortschak: These "If..." lines should not be indented.

native/dlarf.go
```diff
@@ -43,10 +54,10 @@ func (impl Implementation) Dlarf(side blas.Side, m, n int, v []float64, incv int
 		}
 		if applyleft {
 			// Scan for the last non-zero column in C[0:lastv, :]
-			lastc = impl.Iladlc(lastv, n, c, ldc)
+			lastc = impl.Iladlc(lastv+1, n, c, ldc)
```
kortschak: I take it this and the call to Iladr below are optimisations (they are not in netlib)?

native/dgels.go
```diff
@@ -0,0 +1,180 @@
...
+//  2. If m < n and trans == blas.NoTrans, Dgels finds the minimum norm solution of
+//  A * X = B.
+//  3. If m >= n and trans == blas.Trans, Dgels finds the minimum norm solution of
+//  A^T * X = B.
+//  4. If m < n and trans == blas.Trans, Dgels finds X such that || A*X - B||_2
+//  is minimized.
+// Note that the least-squares solutions (cases 1 and 3) perform the minimization
+// per column of B. This is not the same as finding the minimum-norm matrix.
+//
+// The matrix a is a general matrix of size m x n and is modified during this call.
```
kortschak: I think we use '×' without adjacent spaces instead of 'x' with in matrix-dimensional descriptions.

native/dlarfb.go
```diff
@@ -0,0 +1,341 @@
...
+//  V = (  1       )
+//      ( v1  1    )
+//      ( v1 v2  1 )
+//      ( v1 v2 v3 )
+//      ( v1 v2 v3 )
+//  If direct == lapack.Forward and store == lapack.Rowwise
```
kortschak: s/Rowwise/RowWise/g
s/Columnwise/ColumnWise/g

These are hyphenated so camels will rule.

native/dlarft.go
```diff
@@ -0,0 +1,137 @@
...
+			}
+			var lastv int
+			if store == lapack.Columnwise {
+				// skip trailing zeros
+				for lastv = n - 1; lastv >= i+1; lastv-- {
+					if v[lastv*ldv+i] != 0 {
+						break
+					}
+				}
+				for j := 0; j < i; j++ {
```
kortschak: Can you explain the difference here from the netlib code?

native/dgelq2.go
kortschak: Yes please.

native/general.go
kortschak: I use it. I can't add anything beyond that.

native/dlarf.go
btracey: http://www.netlib.org/lapack/explore-html/d3/dd4/dlarf_8f_source.html
Lines 184 and 186

native/dlarft.go
btracey: I'm not sure what you mean. This looks like lines 222-224 of http://www.netlib.org/lapack/explore-html/dd/daa/dlarft_8f_source.html

native/dormqr.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+	if nb < nbmin || nb >= k {
+		// Call unblocked code
+		impl.Dorm2r(side, trans, m, n, k, a, lda, tau, c, ldc, work)
+		return
+	}
+	ldt := nb
+	t := make([]float64, nb*ldt)
+
+	switch {
```
kortschak: Note that the netlib code is simpler here.

native/dlarf.go
kortschak: Interesting: http://www.netlib.org/lapack/explore-3.1.1-html/dlarf.f.html (I read this because I hate the interface the other provides).

native/dlarft.go
kortschak: I'm reading the 3.1.1 code directly. See above.

native/dormqr.go
kortschak: Again - 3.1.1.

btracey: It's simpler, but the netlib code takes advantage of the fortran for loop construct. It seemed cleaner to write out the loop, but maybe I'm missing an obvious transformation. The problem is that the ">=" sign switches with a "<" sign.

kortschak: Factoring this to the commonality makes it easier to see the patterns and ensure that errors are not case-specific, but we can do it later. I don't understand the inequality issue.

kortschak:
It was pretty hard to concentrate through dlarfb, so if we could generate that it would be good - I can look again later if you want.

It would be good to have the new tests passing with the cgo code if possible.

LGTM with trepidation.

kortschak:
@vladimir-ch would you please also take a look.

native/dormqr.go
btracey: So, netlib sets the three loop indices, i1, i2, i3. This is "start at i1, add i3 until you go past i2". If i3 is positive, this is loop until the index becomes greater than or equal to i2. If i3 is negative, this is loop until i1 is less than or equal to i2. I don't know how to make a general transformation of this construct. What I could do instead is set i1 and i3 as the netlib code does, and then have
````
for {
     ib = min(nb, k-i) 
     // common code
    i += i3
    if k-i <= 0{
        break
    }
}
````

I wasn't sure if it was more legible to have the common code or to have the clear loop.

**Comment by btracey** [comment]
What do you mean "If we could generate that part it would be good"?

File: native/dormqr.go (hunk 11 above)
Comment (btracey) [inline native/dormqr.go]: I think the online code is version 3.5.0, which may explain the discrepancy.

**Comment by kortschak** [comment]
I'm just wondering if there is some simpler way of representing the structure here (either in the code itself - probably not - or in some code generator) that reduces the difficulty of reviewing and maintaining it.

File: native/dormqr.go (hunk 11 above)
Comment (kortschak) [inline native/dormqr.go]: Yes, the version explains it (see comment above).

I would prefer the commonality to be emphasised over the loop explicitness, but wait for Vladimir's input.

File: native/dlabad.go [hunk 12]
```diff
@@ -0,0 +1,11 @@
+// Copyright ©2015 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package native
+
+func (impl Implementation) Dlabad(small, large float64) {
```
kortschak: Add doc comment.

native/general.go
```diff
@@ -4,22 +4,65 @@
...
-import "github.com/gonum/lapack"
+import (
+	"math"
+
+	"github.com/gonum/lapack"
+)
 
 type Implementation struct{}
 
+type Blah struct {
```
kortschak: Why is this here?

native/general.go
```diff
@@ -4,22 +4,65 @@
...
 package native
 
-import "github.com/gonum/lapack"
+import (
+	"math"
+
+	"github.com/gonum/lapack"
+)
 
 type Implementation struct{}
```
kortschak: Add doc comment.

lapack.go
```diff
@@ -30,3 +30,34 @@ type Float64 interface {
 }
 
 type Complex128 interface{}
```
kortschak: What is this?

lapack.go
```diff
@@ -30,3 +30,34 @@ type Float64 interface {
...
+const (
+	Forward  (Direct) = 'F'
+	Backward (Direct) = 'B'
+)
+
+// StoreV indicates the storage direction of elementary reflectors.
+type StoreV byte
+
+const (
+	Columnwise (StoreV) = 'C'
```
kortschak: Add doc comments.

lapack.go
```diff
@@ -30,3 +30,34 @@ type Float64 interface {
...
+	MaxAbs       (MatrixNorm) = 'M' // max(abs(A(i,j)))  ('M')
+	MaxColumnSum (MatrixNorm) = 'O' // Maximum column sum (one norm) ('1', 'O')
+	MaxRowSum    (MatrixNorm) = 'I' // Maximum row sum (infinity norm) ('I', 'i')
+	NormFrob     (MatrixNorm) = 'F' // Frobenium norm (sqrt of sum of squares) ('F', 'f', E, 'e')
+)
+
+type MatrixType byte
+
+const (
+	General (MatrixType) = 'G'
```
kortschak: Add doc comment.

lapack.go
```diff
@@ -30,3 +30,34 @@ type Float64 interface {
...
+type MatrixNorm byte
+
+const (
+	MaxAbs       (MatrixNorm) = 'M' // max(abs(A(i,j)))  ('M')
+	MaxColumnSum (MatrixNorm) = 'O' // Maximum column sum (one norm) ('1', 'O')
+	MaxRowSum    (MatrixNorm) = 'I' // Maximum row sum (infinity norm) ('I', 'i')
+	NormFrob     (MatrixNorm) = 'F' // Frobenium norm (sqrt of sum of squares) ('F', 'f', E, 'e')
+)
+
+type MatrixType byte
```
kortschak: Add doc.

lapack.go
```diff
@@ -30,3 +30,34 @@ type Float64 interface {
...
+
+// StoreV indicates the storage direction of elementary reflectors.
+type StoreV byte
+
+const (
+	Columnwise (StoreV) = 'C'
+	Rowwise    (StoreV) = 'R'
+)
+
+type MatrixNorm byte
```
kortschak: Add doc.

native/dgels.go
```diff
@@ -0,0 +1,180 @@
+// Copyright ©2015 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package native
+
+import (
+	"github.com/gonum/lapack"
+
```
kortschak: Delete blank line.

native/dlarfg.go
```diff
@@ -0,0 +1,64 @@
...
+//      (    x)   (   0)
+//  H^T * H = I
+// H is represented in the form
+//  H = 1 - tau * (1; v) * (1 v^T)
+// where tau is a real scalar.
+//
+// On entry, x contains the vector x, on exit it contains v.
+//
+// On entry, c contains alpha. On exit it contains beta.
+func (impl Implementation) Dlarfg(n int, c *float64, x []float64, incX int) (tau float64) {
```
kortschak: Is there a reason we can't return beta instead of using c?

native/dlassq.go
```diff
@@ -0,0 +1,25 @@
+// Copyright ©2015 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package native
+
+import "math"
+
+// Dlassq updates a sum of squares in scaled form.
+func (impl Implementation) Dlassq(n int, x []float64, incx int, scale *float64, sumsq *float64) {
```
kortschak: Remove redundant types. Add explanation of parameters. Can we instead return scale and sumsq?

native/dtrtrs.go
```diff
@@ -0,0 +1,31 @@
...
+	"github.com/gonum/blas/blas64"
+)
+
+// Dtrtrs solves a triangular system of the form a * x = b or a^T * x = b. Dtrtrs
+// checks for singularity in a. If a is singular, false is returned and no solve
+// is performed. True is returned otherwise.
+func (impl Implementation) Dtrtrs(uplo blas.Uplo, trans blas.Transpose, diag blas.Diag, n, nrhs int, a []float64, lda int, b []float64, ldb int) (ok bool) {
+	nounit := diag == blas.NonUnit
+	if n == 0 {
+		return
```
kortschak: explicit false.

native/general.go
```diff
@@ -4,22 +4,65 @@
...
+	badTrans      = "lapack: bad trans"
+	badUplo       = "lapack: illegal triangle"
+	badWork       = "lapack: insufficient working memory"
+	badWorkStride = "lapack: insufficient working array stride"
+	negDimension  = "lapack: negative matrix dimension"
+	nLT0          = "lapack: n < 0"
+	shortWork     = "lapack: working array shorter than declared"
 )
 
 func blockSize() int {
```
kortschak: Add a comment for this - not exported, but it will be useful to know why 64 is chosen for example.

native/ilaenv.go
```diff
@@ -0,0 +1,375 @@
...
+//  4: The number of shifts.
+//  5: The minumum column dimension for blocking to be used.
+//  6: The crossover point for SVD (to use QR factorization or not).
+//  7: The number of processors.
+//  8: The crossover point for multishift in QR and QZ methods for nonsymmetric eigenvalue problems.
+//  9: Maximum size of the subproblems in divide-and-conquer algorithms.
+//  10: ieee NaN arithmetic can be trusted not to trap.
+//  11: infinity arithmetic can be trusted not to trap.
+func (Implementation) Ilaenv(ispec int, s string, opts string, n1, n2, n3, n4 int) int {
+	// TODO(btracey): Relpace this with a constant lookup? A list of constants?
```
kortschak: I was thinking about this. Ilaenv is horrible, but do we need to keep it to allow compatibility with the cgo code?

btracey:
As to auto-generating go code from Fortran, I think it would be hard. Since everything is by reference, it's hard to tell if something is supposed to be a number or a matrix. The next main issue are the rules between 0 indexing and 1-indexing. Loop counters need to change, but number of rows counters don't. Therefore, row - loop needs to change, but loop -loop and row -row doesn't Finally, everything is passed by reference in Fortran, so it's hard to tell (purely from code) what is modified and what isn't.  If I were better at parsers and lexers, autogeneration is  probably possible, but certain issues make it pretty tricky to do without significant guidance.

native/ilaenv.go
btracey: My goal for this PR was to get something working, and part of that was to keep as much of the structure as reasonable in order to ensure I had a working implentation. Personally, I don't think we ever want to expose Ilaenv as part of an interface. mat64 will never need it, as it will call the blocked algorithms directly. A user should never want to call it, as they will also just want to call the blocked algorithm. Cgo may expose the function (as this does now), but I can't see why the user would ever want it.

native/dlarfg.go
btracey: Not particularly. I chose this way because in usage the value is modified in place. For the householder constants (the tau values), the slice contains whatever, and is populated with the correct values. For alpha/beta, the value always comes from the diagonal of the input array (i.e. in dgeqrf) and its modified in place to contain the new value. Thus, it both feels like a pointer, since the value is modified in place, and it made it easier to implement the initial code.

native/general.go
btracey: No good reason. Sorry. Will remove.

lapack.go
btracey: This is still a tie-over from before I started looking at lapack.

kortschak:
I wasn't thinking of autogenerating from fortran, just autogenerating the particularly repetitive portion of dlarfb. It was really just an idea.

native/dlarfg.go
kortschak: The calls then become

```
tau[i], a[i*lda+i] = impl.Dlarfg(n-i, a[i*lda+i], a[i*lda+min(i+1, n-1):], 1)
```

and

```
tau[i], a[i*lda+i] = impl.Dlarfg(m-i, a[i*lda+i], a[min((i+1), m-1)*lda+i:], lda)
```

I'll leave it up to you.

native/dgelq2.go
```diff
@@ -0,0 +1,40 @@
...
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package native
+
+import "github.com/gonum/blas"
+
+// Dgelq2 computes the LQ factorization of the m x n matrix a.
+//
+// During Dgeqr2, a is modified to contain the information to construct Q and L.
```
vladimir-ch: Dgelq2?

native/dgelqf.go
```diff
@@ -0,0 +1,73 @@
...
+)
+
+// Dgelqf computes the LQ factorization of the m x n matrix a using a blocked
+// algorithm. Please see the documentation for Dgelq2 for a description of the
+// parameters at entry and exit.
+//
+// work is vector of temporary storage with length at least m, and will panic if
+// work is shorter. Lwork is a parameter specifying the accessible length of the
+// work slice. Lwork restricts the amount of blocking that may be used.
+// In the special case where lwork == -1, no check will be made on the length of
```
vladimir-ch: I think this does not explain well what happens when lwork == -1.

native/dlarfg.go
```diff
@@ -0,0 +1,64 @@
...
+
+package native
+
+import (
+	"math"
+
+	"github.com/gonum/blas/blas64"
+)
+
+// Dlarfg generates an elementary reflector for a household matrix. It creates
```
vladimir-ch: household -> Householder

native/dormqr.go
vladimir-ch: I also prefer the netlib/common code way, there is a clear setup part and a clear loop part. In fact, the for-loop above is a rather nice translation of the Fortran loop into Go.

lapack.go
btracey: Done.

lapack.go
btracey: Done.

lapack.go
btracey: Done.

native/dgelq2.go
btracey: Done.

native/dgelq2.go
btracey: Alright. I split up a bunch of lines. I tried to use good judgement. If it was short enough to read on one line I left it, and in things like Dgemm where it is alpha * a * b + beta * c , I put the alpha * a * b term on one line.

native/dgelqf.go
btracey: Updated. Let me know if it's better.

native/dgels.go
btracey: Done.

native/dgels.go
btracey: Tried to get them all.

native/dlabad.go
btracey: What would you like me to say? Here is the original comment

````
DLABAD takes as input the values computed by DLAMCH for underflow and
overflow, and returns the square root of each of these values if the
log of LARGE is sufficiently large.  This subroutine is intended to
identify machines with a large exponent range, such as the Crays, and
redefine the underflow and overflow limits to be the square roots of
the values computed by DLAMCH.  This subroutine is needed because
DLAMCH does not compensate for poor arithmetic in the upper half of
the exponent range, as is found on a Cray.
````

File: native/dlange.go (hunk 5 above)
Comment (btracey) [inline native/dlange.go]: Done.

File: native/dlarfb.go (hunk 6 above)
Comment (btracey) [inline native/dlarfb.go]: Okay, done.

File: native/dlarfb.go (hunk 9 above)
Comment (btracey) [inline native/dlarfb.go]: Done.

Maybe we should work on eliminating hyphenated words. I'm not looking forward to camel rule.

File: native/dlarfg.go (hunk 28 above)
Comment (btracey) [inline native/dlarfg.go]: Done.

File: native/dlarfg.go (hunk 21 above)
Comment (btracey) [inline native/dlarfg.go]: Yea, I don't know what the right balance is in terms of API changes. I'll change it to not use a pointer.

File: native/dlassq.go (hunk 22 above)
Comment (btracey) [inline native/dlassq.go]: Done.

File: native/dormqr.go (hunk 11 above)
Comment (btracey) [inline native/dormqr.go]: Alright, I changed the code around, but I really don't think it's simpler. The different cases loop in different directions in the different cases (top vs bottom, forward vs backward). This is disguised by looking common in the netlib code. It's easy to see in the separate case that, for example, n is fixed in left && notran, rather than having so see that ni is never modified.

File: native/general.go (hunk 14 above)
Comment (btracey) [inline native/general.go]: Done.

File: native/general.go (hunk 24 above)
Comment (btracey) [inline native/general.go]: Nope. Deleted and replaced with a Ilaenv call in Dpotrf.

**Comment by btracey** [comment]
I don't understand why it's failing travis. It's saying that it can't find lapack.Complex128, which is true as it doesn't exist. However, cgo/lapack.go doesn't exist as a file either, and I don't see anything in the travis.yml file telling it to run the perl script.

**Comment by kortschak** [comment]
https://github.com/gonum/lapack/blob/master/.travis/linux/OpenBLAS/install.sh is calling genLapack.pl.

File: lapack.go [hunk 29]
```diff
@@ -29,4 +29,35 @@ type Float64 interface {
 	Dpotrf(ul blas.Uplo, n int, a []float64, lda int) (ok bool)
 }
 
-type Complex128 interface{}
```
kortschak: It seems we need to keep this - just with documentation saying that it is the set of LAPACK functions we support.

native/dlabad.go
kortschak: SGTM except that it doesn't return anything - do you mean record?

lapack.go
btracey: Okay, thanks for checking. What do we mean by "support" in this package? I know we're not really sure, but support is pretty vague. My plan had been to add those functions needed by matrix, unless we explicitly need something.

Another possible option is to add all of the Go functions we have. However, there are a number of partial LAPACK implementations. Atlas and OpenBLAS both have a couple of functions, and I believe Eigen does as well. Keeping the interfaces minimal makes it easier for outside implementations to fully support them.

native/dlassq.go
```diff
@@ -0,0 +1,26 @@
+// Copyright ©2015 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package native
+
+import "math"
+
+// Dlassq updates a sum of squares in scaled form.
+func (impl Implementation) Dlassq(n int, x []float64, incx int, scale float64, sumsq float64) (float64, float64) {
```
kortschak: Name the return parameters - tricky, sorry.

btracey: Is there any sort of convention for these kind of names? scaleAfter or something?

lapack.go
kortschak: >My plan had been to add those functions needed by matrix, unless we explicitly need something.

SGTM - or by request via issues.

native/dlabad.go
btracey: That was the netlib comment for the original LAPACK routine. Since parameters are passed by reference, "return" really means "overwrite". This doesn't apply since Go guarantees IEEE-754. We could also delete it entirely.

lapack.go
btracey: Agreed.

native/dlassq.go
kortschak: From the fortran docs, "scl" and "smsq". Horrible, but with precedence.

native/dlabad.go
kortschak: I'll leave it up to you as to whether it's kept, but it seems to me that if it is kept, either you accept pointers (something we avoid) or return the values unaltered.

btracey: Deleted.

kortschak: Nice work!

native/dlassq.go
btracey: Added the returns.

native/dlabad.go
btracey: It was unclear to me how closely to follow the original code. I'm all for deleting code if we think that's okay.

btracey:
Responded to comments.

kortschak:
LGTM but wait for @vladimir-ch if he has time.

lapack.go
```diff
@@ -11,22 +11,57 @@ const None = 'N'
 type Job byte
```
vladimir-ch: Should `Job` and `CompSV` below also have a comment?

lapack.go
```diff
@@ -11,22 +11,57 @@ const None = 'N'
...
 type CompSV byte
 
 const (
-	Compact  (CompSV) = 'P'
-	Explicit (CompSV) = 'I'
+	Compact  CompSV = 'P'
+	Explicit CompSV = 'I'
 )
 
+// Complex128 defines the complex128 interface for the Lapack function. This interface
```
vladimir-ch: Perhaps it is some nuance with definite articles in English, but what is _the_ Lapack function? Also below.

lapack.go
```diff
@@ -11,22 +11,57 @@ const None = 'N'
...
+	ColumnWise StoreV = 'C' // Reflector stored in a column of the matrix.
+	RowWise    StoreV = 'R' // Reflector stored in a row of the matrix.
+)
+
+// MatrixNorm represents the kind of matrix norm to compute.
+type MatrixNorm byte
+
+const (
+	MaxAbs       MatrixNorm = 'M' // max(abs(A(i,j)))  ('M')
+	MaxColumnSum MatrixNorm = 'O' // Maximum column sum (one norm) ('1', 'O')
```
vladimir-ch: What are the characters '1', 'i', 'f', ... in parenthesis at the end here and below? Are these alternative characters that original Lapack also accepts?

lapack.go
kortschak: In this case it appears to be referring to a generic lapack function described by one of the interface methods. It could be reworded to be clearer - I think interface is being overloaded here (the method of the Go interface is describing the relevant LAPACK function interface).

Maybe:

```
Complex128 defines the public complex128 LAPACK API supported by gonum.
```

lapack.go
kortschak: Yes.

native/dgelq2.go
```diff
@@ -0,0 +1,44 @@
...
+// (not including the diagonal) contain the elementary reflectors. Tau is modified
+// to contain the reflector scales. Tau must have length of at least k = min(m,n)
+// and will panic otherwise.
+//
+// See Dgeqr2 for a description of the elementary reflectors and orthonormal
+// matrix Q. Q is constructed as a product of these elementary reflectors,
+// Q = H(k) ... H(2) H(1).
+//
+// Work is a temporary storage slice of length at least m and will panic otherwise.
+func (impl Implementation) Dgelq2(m, n int, a []float64, lda int, tau, work []float64) {
```
vladimir-ch: Should `lda` be also documented? Probably aplicable at many places throughout this PR.

kortschak: This probably would be best done in a package-level comment because it is a LAPACK/BLAS standard approach. I don't think we do it for blas (we probably should).

native/dgelqf.go
```diff
@@ -0,0 +1,81 @@
...
+
+// Dgelqf computes the LQ factorization of the m×n matrix a using a blocked
+// algorithm. Please see the documentation for Dgelq2 for a description of the
+// parameters at entry and exit.
+//
+// work is vector of temporary storage with length at least m, and will panic if
+// work is shorter. Lwork is a parameter specifying the accessible length of the
+// work slice. Dgelqf is a blocked LQ factorization, but the block size is limited
+// by the temporary space available. If lwork == -1, instead of performing Dgelqf,
+// instead the optimal work length will stored into work[0].
```
vladimir-ch: Yes, this is much better. But the wording is strange. It seems to me that 'instead' can be deleted and 'will stored' should be 'will be stored'.

From the previous reading I remember that this comment clarification could be also done for other functions below. Sorry for not poiting it out earlier.

native/dgels.go
```diff
@@ -0,0 +1,204 @@
...
+package native
+
+import (
+	"github.com/gonum/blas"
+	"github.com/gonum/lapack"
+)
+
+// Dgels finds a minimum-norm solution based on the matrices a and b using the
+// QR or LQ factorization to find the solution. Dgels returns false if the matrix
```
vladimir-ch: I feel like nitpicking, but the `to find the solution` seems unnecessary to me.

native/dgels.go
```diff
@@ -0,0 +1,204 @@
...
+// A is singular, and true if this solution was successfully found.
```
vladimir-ch: Is it possible that A is not singular and still the solution could not be found?

native/dgels.go
```diff
@@ -0,0 +1,204 @@
...
+//
+// The minimization problems solved depends on the input parameters.
```
vladimir-ch: problems -> problem

native/dgels.go
```diff
@@ -0,0 +1,204 @@
...
+//  is minimized.
+// Note that the least-squares solutions (cases 1 and 3) perform the minimization
+// per column of B. This is not the same as finding the minimum-norm matrix.
+//
+// The matrix a is a general matrix of size m×n and is modified during this call.
+// The input matrix b is of size max(m,n)×nrhs, and serves two purposes. On entry,
+// the elements of b specify the input matrix B. B has size m×nrhs if
+// trans == blas.NoTrans, and n×nrhs if trans == blas.Trans. On exit, the
+// leading submatrix of b contains the solution vectors X. If trans == blas.NoTrans,
+// this submatrix is of size n×nrhs, and in m×nrhs otherwise.
```
vladimir-ch: in -> of or delete or something else?

lapack.go
vladimir-ch: > Complex128 defines the public complex128 LAPACK API supported by gonum/lapack.

I like that.

native/dgelq2.go
vladimir-ch: > This probably would be best done in a package-level comment because it is a LAPACK/BLAS standard approach.

That's what I thought. A package-level comment seems like the best place for it.

native/dgeqr2.go
```diff
@@ -0,0 +1,57 @@
+// Copyright ©2015 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package native
+
+import "github.com/gonum/blas"
+
+// Dgeqr2 computes a QR factorization of the m x n matrix a.
```
vladimir-ch: m x n -> m×n

native/dgelq2.go
```diff
@@ -0,0 +1,44 @@
...
+//
+// During Dgelq2, a is modified to contain the information to construct Q and L.
+// The lower triangle of a contains the matrix L. The upper triangular elements
+// (not including the diagonal) contain the elementary reflectors. Tau is modified
+// to contain the reflector scales. Tau must have length of at least k = min(m,n)
+// and will panic otherwise.
+//
+// See Dgeqr2 for a description of the elementary reflectors and orthonormal
+// matrix Q. Q is constructed as a product of these elementary reflectors,
+// Q = H(k) ... H(2) H(1).
```
vladimir-ch: The notation for elementary reflectors should be unified, now they are denoted by H(k), H_k or h_k. I like the H_k version best, but it is up to you.

native/dgeqr2.go
```diff
@@ -0,0 +1,57 @@
...
+// the
+//  v[j] = 0           j < i
+//  v[j] = i           j == i
+//  v[j] = a[i*lda+j]  j > i
+// and computing h_i = I - tau[i] * v * v^T.
+//
+// The orthonormal matrix Q can be constucted from a product of these elementary
+// reflectors, Q = H(1)H(2) ... H(k), where k = min(m,n).
+//
+// Work is a temporary storage matrix of at least len(n), and will panic otherwise.
```
vladimir-ch: Work is a temporary storage of length at least n, and Dgeqr2 will panic otherwise?

vladimir-ch: I think that if these `Work is ...` comments had the same structure, they could be easier to parse visually to extract the minimum required length. For example, now work is described as slice/vector/matrix of/with length ...

native/dgeqrf.go
```diff
@@ -0,0 +1,93 @@
...
+import (
+	"github.com/gonum/blas"
+	"github.com/gonum/lapack"
+)
+
+// Dgeqrf computes the QR factorization of the m×n matrix a using a blocked
+// algorithm. Please see the documentation for Dgeqr2 for a description of the
+// parameters at entry and exit.
+//
+// work is vector of temporary storage with length at least n, and will panic
```
vladimir-ch: An example of `Work is ...` comment. At least Dgeqrf should be inserted before will, but also see above.

native/dgeqrf.go
```diff
@@ -0,0 +1,93 @@
...
+// otherwise. The length of work limits the ability for performing optimal blocking.
+// Running Dgeqrf with lwork == -1 will store the optimal work length into work[0].
+//
+// tau must be at least len min(m,n), and will panic otherwise.
```
vladimir-ch: tau must have length at least min(m,n), and Dgeqrf will panic otherwise?

native/dlange.go
```diff
@@ -0,0 +1,76 @@
...
+
+import (
+	"math"
+
+	"github.com/gonum/lapack"
+)
+
+// Dlange computes the matrix norm of the general m×n matrix a. The input norm
+// specifies the norm computed.
+//  lapack.MaxAbs: the absolute value of the maximum element.
```
vladimir-ch: the maximum absolute value of the entries/elements?

native/dlapy2.go
```diff
@@ -0,0 +1,12 @@
+// Copyright ©2015 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package native
+
+import "math"
+
+// Dlapy2 is the lapack version of math.Hypot.
```
vladimir-ch: lapack -> LAPACK?

native/dlarfb.go
```diff
@@ -0,0 +1,422 @@
...
+//  V = ( v1 v1  1       )
+//      ( v2 v2 v2  1    )
+//      ( v3 v3 v3 v3  1 )
+// An elementary reflector can be explicitly constructed by extracting the
+// corresponding elements of v, placing a 1 where the diagonal would be, and
+// placing zeros in the remaining elements.
+//
+// t is a k×k matrix containing the block reflector. See Dlarft for more information.
+//
+// Work contains space for temporary memory used by Dlarfb. Work must be of size
```
vladimir-ch: Another version of `Work is ...`

native/dlarfb.go
```diff
@@ -0,0 +1,422 @@
...
+	"github.com/gonum/lapack"
+)
+
+// Dlarfb applies a block reflector to a matrix.
+//
+// In the call to Dlarfb, c is multiplied by the implicitly defined matrix h.
+//  c = h * c if side == Left and trans == NoTrans
+//  c = c * h if side == Right and trans == NoTrans
+//  c = h^T * c if side == Left and trans == Trans
+//  c = c * h^t if side == Right and trans == NoTrans
```
vladimir-ch: NoTrans -> Trans

native/dlarfb.go
```diff
@@ -0,0 +1,422 @@
...
+
+import (
+	"github.com/gonum/blas"
+	"github.com/gonum/blas/blas64"
+	"github.com/gonum/lapack"
+)
+
+// Dlarfb applies a block reflector to a matrix.
+//
+// In the call to Dlarfb, c is multiplied by the implicitly defined matrix h.
```
vladimir-ch: What about modifying it to: `... the mxn matrix c is ... matrix h as follows:` ? Otherwise the size of c is not explicitly stated and this seems like a good place.

native/dlarfg.go
```diff
@@ -0,0 +1,60 @@
...
+				break
+			}
+		}
+		xnorm = bi.Dnrm2(n-1, x, incX)
+		beta = -math.Copysign(impl.Dlapy2(alpha, xnorm), alpha)
+	}
+	tau = (beta - alpha) / beta
+	bi.Dscal(n-1, 1/(alpha-beta), x, incX)
+	for j := 0; j < knt; j++ {
+		beta = beta * safmin
```
vladimir-ch: *=

native/dlascl.go
```diff
@@ -0,0 +1,72 @@
...
+
+package native
+
+import (
+	"math"
+
+	"github.com/gonum/lapack"
+)
+
+// Dlascl multiplies a rectangular matrix by a scalar.
```
vladimir-ch: This could probably be expanded a bit to explain the various arguments.

native/dlascl.go
```diff
@@ -0,0 +1,72 @@
...
+func (impl Implementation) Dlascl(kind lapack.MatrixType, kl, ku int, cfrom, cto float64, m, n int, a []float64, lda int) {
```
vladimir-ch: What are `kl` and `ku`? They are not used.

native/dlassq.go
```diff
@@ -0,0 +1,26 @@
...
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package native
+
+import "math"
+
+// Dlassq updates a sum of squares in scaled form.
+func (impl Implementation) Dlassq(n int, x []float64, incx int, scale float64, sumsq float64) (scl, smsq float64) {
+	if n < 0 {
```
vladimir-ch: `<=` ?

native/dlassq.go
```diff
@@ -0,0 +1,26 @@
+// Copyright ©2015 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package native
+
+import "math"
+
+// Dlassq updates a sum of squares in scaled form.
```
vladimir-ch: This comment is also rather minimalistic.

native/general.go
```diff
@@ -4,20 +4,57 @@
...
 
-import "github.com/gonum/lapack"
+import (
+	"math"
 
+	"github.com/gonum/lapack"
+)
+
+// Implementation is the native Go implementation of LAPACK routines. It
+// is built on top of calls to the return of blas64.Implementation, so while
```
vladimir-ch: `... to the return of ...`?

testlapack/general.go
```diff
@@ -4,9 +4,289 @@
...
+			blas64.Gemm(blas.NoTrans, blas.NoTrans, 1, hcopy, hi, 0, h)
+		} else {
+			// H = H_I * H in backward mode
+			blas64.Gemm(blas.NoTrans, blas.NoTrans, 1, hi, hcopy, 0, h)
+		}
+	}
+	return h
+}
+
+// constructQ_QR constructs the Q matrix from the result of dgeqrf and dgeqr2
```
vladimir-ch: constuctQ_QR -> constructQ

vladimir-ch:
Sorry for the flood of nitpicking. I will try to give it a second pass to look more at the code itself in the afternoon when I have recovered some strength to concentrate.

native/dormqr.go
vladimir-ch: Yes, it looks worse now. There really are four different cases, a fact which is obfuscated in the current code. If you feel like changing it back, I agree.

native/dgelqf.go
```diff
@@ -0,0 +1,81 @@
...
+//
+// tau must have length at least min(m,n).
```
vladimir-ch: This condition does not seem to be checked.

native/dgels.go
```diff
@@ -0,0 +1,204 @@
...
+// work is a slice for temporary storage, and lwork specifies the usable length of
+// the slice. At minimum, lwork >= max(m,n) + max(m,n,nrhs), and Dgels will panic
+// if this is not met. A longer work will enable blocked algorithms to be called.
+// In the special case that lwork == -1, work[0] will be set to the optimal working
+// length.
+func (impl Implementation) Dgels(trans blas.Transpose, m, n, nrhs int, a []float64, lda int, b []float64, ldb int, work []float64, lwork int) bool {
+	notran := trans == blas.NoTrans
+	checkMatrix(m, n, a, lda)
+	mn := min(m, n)
+	if notran {
```
vladimir-ch: Both branches do the same thing.

native/dgeqr2.go
```diff
@@ -0,0 +1,57 @@
...
+func (impl Implementation) Dgeqr2(m, n int, a []float64, lda int, tau []float64, work []float64) {
```
vladimir-ch: Remove type for tau.

native/dgeqrf.go
vladimir-ch: This condition is not checked.

vladimir-ch:
I think it LGTM after addressing my comments. Great work!

kortschak:
Both of you deserved a round of applause for this behemoth.

lapack.go
btracey: I wasn't sure what they do, as they were there before I started editing and they are not used in gonum/native. I did what anyone does and googled it. gonum/lapack (and forks) is 4 of the top 7 results.

It seems as if Job does not mean something consistent. Ddisna, for example, has 3 choices for the Job char that are completely unrelated to the ones that are there now. I'd like to leave these uncommented for now. We can change the OpenBLAS build and delete them.

lapack.go
btracey: Agreed. Thanks. Done.

lapack.go
btracey: Yes. I found it useful when looking at fortran Lapack code. The actual lapack code has things like "if norm == '1'", and it's nice to have a reference what that means. I thought it would be useful to others with fortran experience, but I can delete if you'd prefer.

native/dgelq2.go
btracey: The data types (so far) are identical to BLAS. blas/native has a large doc comment about the data layout http://godoc.org/github.com/gonum/blas/native . There is a comment in the lapack package level documentation to this effect: "Slice function arguments frequently represent vectors and matrices. The data layout is identical to that found in https://godoc.org/github.com/gonum/blas/native. "
I don't think it's worth doubling the documentation on layout information. Users will hopefully be using mat64, and if not then hopefully at least blas64 and lapack64. If we can modify the lapack package documentation to be more clear, I'm all for it.

native/dgelqf.go
btracey: Done.

native/dgels.go
btracey: Nitpicks encouraged. Done.

native/dgels.go
btracey: I don't think so (barring floating point issues). Dormqr cannot fail. Only the triangular solve can fail, and that only fails if the diagonal entry is zero. If the matrix is really badly conditioned you could have some Infs from floating point issues, but ok would still be true because it is not detected in this function.

native/dgels.go
btracey: Done.

native/dgels.go
btracey: Done.

native/dgeqr2.go
btracey: Thanks. Done.

native/dgelq2.go
btracey: Done.

native/dgeqr2.go
btracey: One of the issues is that sometimes work is treated as if it's a matrix and other times as if it is a slice. I like "is temporary storage of length", avoids the issue.

btracey: Oh, and sometimes work is just an array, but other times it's an array with a length (work and lwork)

lapack.go
kortschak: CompSV can though. It is pretty consistently (half a dozen checked) mean compute singular values.

Job could have its constants deleted since they are pretty broad - and the meaning of it seems to be "What job should I do?"

native/dgeqr2.go
btracey: Attempted to structure similarly.

native/dgeqrf.go
btracey: Done.

native/dlange.go
btracey: Done.

native/dlapy2.go
btracey: Done.

native/dlarfb.go
btracey: Done.

native/dlarfb.go
btracey: Done.

native/dlarfb.go
btracey: Done.

native/dlarfg.go
btracey: Done.

native/dlascl.go
btracey: They are for the banded types that have not yet been implemented in this function. This is a stupid function, especially by LAPACK standards. It should be Dgescl or something like that rather than switch over all of the matrix types. The input parameters have a variety of meanings depending on the matrix type.

native/dlassq.go
btracey: Done.

native/dlassq.go
btracey: Done.

native/general.go
btracey: Yes. I added parentheses, but it's the type returned by http://godoc.org/github.com/gonum/blas/blas64#Implementation

testlapack/general.go
btracey: Done.

native/dgelqf.go
btracey: Done.

native/dgels.go
btracey: Done.

native/dgeqr2.go
btracey: Done.

native/dlascl.go
vladimir-ch: That also explains my confusion about cfrom and cto :-)

btracey:
Thank you as well Dan. Sorry again this was so large. I'll keep them smaller in the future even if I need to implement a few more for understanding.

native/dgels.go
vladimir-ch: Don't tempt me :-)

PR #22, opened 2015-07-30
Add Dgetf2 (unblocked LU algorithm) and tests

btracey:


btracey:
PTAL @kortschak

native/dgetf2.go
```diff
@@ -0,0 +1,64 @@
...
+
+import (
+	"math"
+
+	"github.com/gonum/blas/blas64"
+)
+
+// Dgetf2 computes the LU decomposition of the m×n matrix a.
+// The LU decomposition is a factorization of a into
+//  a = P * L * U
```
kortschak: I don't know if we do this elsewhere, but we probably should if not. "a" here should be "A" since we are not actually describing the abstract matrix in the parameter a as we do in mat64; m, n, a and lda are required to do this. So, leave the parameter as "a" and change the doc to "A" where appropriate. Probably we need to vet all the docs at some point to ensure we are doing this consistently.

btracey:
Sorry, I forgot I cheated on the swapping. Fixing now.

kortschak:
Add tests for the cgo implementation.

btracey:
mat64 will use the blocked version of the algorithm (dgetrf). Do we want cgo tests even though mat64 won't import dgetf2?

kortschak:
Yes, please.

native/dgetf2.go
```diff
@@ -0,0 +1,63 @@
...
+			// Swap the rows if necessary.
+			if jp != j {
+				bi.Dswap(n, a[j*lda:], 1, a[jp*lda:], 1)
+			}
+			if j < m-1 {
+				aj := a[j*lda+j]
+				if math.Abs(aj) >= sfmin {
+					bi.Dscal(m-j-1, 1/aj, a[(j+1)*lda+j:], lda)
+				} else {
+					for i := 0; i < m-j+1; i++ {
```
kortschak: Should this `+1` be here?

kortschak:
The issue of blas/native v blas/cgo and their interaction with lapack/native v lapack/cgo I think needs to be sorted out. Proviously the configuration was simple since we didn't use lapack, now there are two knobs where one is ignored in some states of the other. This needs at least to be documented in mat64.

btracey:
Added cgo tests. Agreed on the need for mat64 documentation.

native/dgetf2.go
btracey: Assuming you mean -1, yes. Nice catch.

kortschak: The loop in fortran is `1,m-j` which is `1 <= i <= m-j`; shifted to 0-based `0 <= i <= m-j-1` which is `0 <= i < m-j`. So no adjustment at all. Am I missing something?

btracey: Yes. j is 1-indexed, so to shift from Fortran j to Go j you also need to subtract 1.

btracey: For example, for the first column, in fortran it's 1 <= i <= m -1.

kortschak: Ah, OK. Thanks.

native/dgetf2.go
```diff
@@ -0,0 +1,63 @@
...
+//  A = P * L * U
+// where P is a permutation matrix, L is a unit lower triangular matrix, and
+// U is a (usually) non-unit upper triangular matrix. On exit, L and U are stored
+// in place into a.
+//
+// ipiv is a permutation vector. It indicates that row i of the matrix was
+// changed with ipiv[i]. ipiv must have length at least min(m,n), and will panic
+// otherwise.
+//
+// Dgetf2 returns whether the matrix a is singular. The LU decomposition will
```
kortschak: add the name ok to the return - at the moment the documentation make it look like you are returning (isSingular bool).

kortschak:
I'm not sure why the cgo code is failing on <go1.5. The OpenBLAS code is [here](https://github.com/xianyi/OpenBLAS/blob/develop/lapack/getf2/getf2_k.c) and it looks like it should be doing the right things wrt the return boolean.

kortschak:
What's more, it passes on my machine.

btracey:
Mine too.

native/dgetf2.go
btracey: I think it is returning isSingular. The factorization doesn't stop when a singularity is found.

kortschak:
Would you make an exploratory change to clapack.Dgetf2 to print the return value before calling isZero so we can see what is happening on the travis machine? Also add `echo OpenBLAS $(git rev-parse HEAD)` just below https://github.com/gonum/lapack/blob/master/.travis/linux/OpenBLAS/install.sh#L10 (we might want to keep that there - but for another PR). This will all be reset after we see what's going on.

native/dgetf2.go
kortschak: In the OpenBLAS code info is set to != 0 when it's singular and the fortran specifies info == 0 indicates OK. Here you set ok = false when you see a singularity. And you call it "ok".

btracey:
I'll make the change in a few minutes.

native/dgetf2.go
btracey: Oh, I see your point. Yes.

kortschak:
I suggested wrong line in install.sh - I've pushed the correct version now.

btracey:
So I shouldn't add the cgo test lines? Sorry, I'm also helping a new gopher at the moment.

kortschak:
Yes, add them, and the other things (see #23 but wait for it to succeed).

btracey:
Pushed. Here's what it prints on my machine
````
brendan:~/Documents/mygo/src/github.com/gonum/lapack/cgo$ go test
0
0
0
0
0
0
2
2
2
2
PASS
ok  	github.com/gonum/lapack/cgo	1.156s
````

**Comment by kortschak** [comment]
So the last two give different results. Unfortunately the rev-parse is on the wrong line (note where it is on #23 - I was wrong in my original instructions), so we can't see which OpenBLAS is being used.

**Comment by kortschak** [comment]
I suspect that this is a bug in OpenBLAS, but we need to the HEAD sha to report - I imagine that both of use are using an older version of OpenBLAS than the current default HEAD.

**Comment by kortschak** [comment]
OK, I'm officially bewildered. I've built https://github.com/xianyi/OpenBLAS/commit/3f1b57668e3d73944981d7bda194ca133f4660b7 which is the sha that was reported in the #23 tests and so presumably is what was built here also, and then tested that on my machine.

I get pass and the same output as you.

**Comment by kortschak** [comment]
@xianyi Do you have any ideas why OpenBLAS Dgetf2 might be returning singular matrices as not singular, but only on travis hardware?

**Comment by vladimir-ch** [comment]
I'm not sure if it is related, but in optimize we also had problems with tests failing only on Travis, they were running differently. As for the reason, I don't know.

**Comment by btracey** [comment]
I assume I should revert the cgo changes, or are they still useful?

**Comment by kortschak** [comment]
Don't revert them, reset --hard e86d740 and push -f.

File: cgo/lapack.go [hunk 4]
```diff
@@ -50,3 +75,32 @@ func (impl Implementation) Dpotrf(ul blas.Uplo, n int, a []float64, lda int) (ok
...
+// will occur if the false is returned and the result is used to solve a
+// system of equations.
+func (Implementation) Dgetf2(m, n int, a []float64, lda int, ipiv []int) (ok bool) {
+	mn := min(m, n)
+	checkMatrix(m, n, a, lda)
+	if len(ipiv) < mn {
+		panic(badIpiv)
+	}
+	ipiv32 := make([]int32, len(ipiv))
+	ok := clapack.Dgetf2(m, n, a, lda, ipiv32)
```
kortschak: Woops

btracey:
Oh, oops.

btracey:
What is the status of this PR? Are we waiting until we have a resolution on the OpenBLAS error?

kortschak:
Press ahead, neither of us can reproduce the failure, one at least with the same ref for OpenBLAS. I guess we can open an issue with xianyi/OpenBLAS; it might go better if we can write a C reproducer to have travis build - this should be relatively straight forward.

PR #23, opened 2015-07-30
travis: update go1.3beta2 -> go1.5beta3

kortschak:
Also output OpenBLAS sha during build.

Merging because it fixes the build.

TBR @jonlawlor


btracey:
LGTM

PR #24, opened 2015-08-01
Add dgetrf and tests

btracey:


btracey:
Wrong about something. Fixing.

btracey:
Okay. Fixed. PTAL @kortschak

btracey:
OpenBLAS tests pass on my machine.

cgo/lapack.go
```diff
@@ -13,6 +13,7 @@ import (
 
 // Copied from lapack/native. Keep in sync.
```
kortschak: Can we reflect this comment in the native location as well - I should have done that.

cgo/lapack.go
```diff
@@ -104,3 +105,34 @@ func (Implementation) Dgetf2(m, n int, a []float64, lda int, ipiv []int) (ok boo
 	}
 	return ok
 }
+
+// Dgetrf computes the LU decomposition of the m×n matrix a.
```
kortschak: Upper case "A" in most cases here.

cgo/lapack.go
```diff
@@ -104,3 +105,34 @@ func (Implementation) Dgetf2(m, n int, a []float64, lda int, ipiv []int) (ok boo
...
+// The LU decomposition is a factorization of a into
+//  A = P * L * U
+// where P is a permutation matrix, L is a unit lower triangular matrix, and
+// U is a (usually) non-unit upper triangular matrix. On exit, L and U are stored
+// in place into a.
+//
+// ipiv is a permutation vector. It indicates that row i of the matrix was
```
kortschak: Can we note that ipiv is zero-based here. It may be surprising that we don't do one-based when everyone (?) else does.

cgo/lapack.go
```diff
@@ -104,3 +105,34 @@ func (Implementation) Dgetf2(m, n int, a []float64, lda int, ipiv []int) (ok boo
...
+func (impl Implementation) Dgetrf(m, n int, a []float64, lda int, ipiv []int) (ok bool) {
+	mn := min(m, n)
+	checkMatrix(m, n, a, lda)
+	if len(ipiv) < mn {
+		panic(badIpiv)
+	}
+	ipiv32 := make([]int32, len(ipiv))
+	ok = clapack.Dgetrf(m, n, a, lda, ipiv32)
+	for i, v := range ipiv32 {
+		ipiv[i] = int(v) - 1 // OpenBLAS returns one indexed.
```
kortschak: Is it only OpenBLAS or all cblas? If it's only OpenBLAS we should do nothing and note that the behaviour is implementation dependent and if it's everyone, we should not mention OpenBLAS by name.

native/dgetrf.go
```diff
@@ -0,0 +1,66 @@
+package native
+
+import (
+	"github.com/gonum/blas"
+	"github.com/gonum/blas/blas64"
+)
+
+// Dgetrf computes the LU decomposition of the m×n matrix a.
```
kortschak: Capitalisation of "A".

native/dgetrf.go
```diff
@@ -0,0 +1,66 @@
...
+// The LU decomposition is a factorization of a into
+//  A = P * L * U
+// where P is a permutation matrix, L is a unit lower triangular matrix, and
+// U is a (usually) non-unit upper triangular matrix. On exit, L and U are stored
+// in place into a.
+//
+// ipiv is a permutation vector. It indicates that row i of the matrix was
```
kortschak: Note that ipiv is zero-based (probably in other places also - maybe another PR?).

native/dlaswp.go
```diff
@@ -0,0 +1,24 @@
+package native
+
+import "github.com/gonum/blas/blas64"
+
+// Dlaswp swaps the rows k1 to k2 of a according to the indices in ipiv.
+// a is a matrix with n columns and stride lda. incX is the increment for ipiv.
+// k1 and k2 are zero-indexed. If incX is negative, then
```
kortschak: Join these lines.

native/general.go
```diff
@@ -18,6 +18,7 @@ type Implementation struct{}
 var _ lapack.Float64 = Implementation{}
 
 const (
```
kortschak: Sync comment.

testlapack/dgetf2.go
```diff
@@ -173,3 +98,93 @@ func Dgetf2Test(t *testing.T, impl Dgetf2er) {
...
+		Rows:   m,
+		Cols:   mn,
+		Stride: ldl,
+		Data:   l,
+	}
+	blas64.Gemm(blas.NoTrans, blas.NoTrans, 1, L, U, 0, LU)
+
+	p := make([]float64, m*m)
+	ldp := m
+	for i := 0; i < m; i++ {
```
kortschak: This seems unnecessarily complicated. Why not just set the ones in the correct place from the outset?

kortschak:
LGTM with minor comments.

testlapack/dgetf2.go
btracey: Is there an easy way to do that? A single row may be swapped multiple times.

kortschak: Making a pi�ata right now, but I'll post code in a while.

btracey: Sounds like fun! I'm heading out for a run anyway.

cgo/lapack.go
btracey: I don't know. I only have OpenBLAS installed on my machine. Last time I tried (admittedly a few months ago) ATLAS didn't compile on OSX (and I'm not sure if ATLAS even has a full lapack implementation). I think both ATLAS and Netlib state that there is no official c API for Lapack.

testlapack/dgetf2.go
kortschak: Sorry, the idea I had doesn't work - this is the right way to do it.

cgo/lapack.go
btracey: Ah, oops. I had this in at one point as well.

cgo/lapack.go
btracey: Oops, sorry.

cgo/lapack.go

cgo/lapack.go
btracey: Changed wording.

native/dgetrf.go
btracey: Added the comment for the LU cases.

native/dlaswp.go
btracey: Done.

PR #26, opened 2015-08-03
testlapack: be more tolerant of singularity misses

kortschak:
Also remove Errorf in favour of Error where possible.

See https://github.com/xianyi/OpenBLAS/issues/610#issuecomment-127095782

@btracey Please take a look.


btracey:
LGTM. We'll need to find a solution to the full problem (testing for == 0), and that solution probably lives at the mat64 level. I don't know if it's to look at the condition number or to test the diagonal elements or nothing.

PR #30, opened 2015-08-10
Fix Gels to return singular result

btracey:


kortschak:
LGTM are there others at need this to happen?

btracey:
No. I checked all the others.

PR #31, opened 2015-08-10
Fix Dgels to check the correct dimension

btracey:


kortschak:
Are you sure? Netlib defines MN = MIN( M, N ).

btracey:
It does? From Netlib Dgels: "The leading dimension of the array B. LDB >= MAX(1,M,N)."

B is an M x NRHS matrix. X is an N x NRHS matrix. Both of these matrices have to fit into B, so the number of rows has to be at least max(m,n)

kortschak:
The issue is that NETLIB makes the check of ldb against max(m, n), but uses mn = min(m, n) for checking against lwork.

```
      INFO = 0
      MN = MIN( M, N )
      LQUERY = ( LWORK.EQ.-1 )
      IF( .NOT.( LSAME( TRANS, 'N' ) .OR. LSAME( TRANS, 'T' ) ) ) THEN
         INFO = -1
      ELSE IF( M.LT.0 ) THEN
         INFO = -2
      ELSE IF( N.LT.0 ) THEN
         INFO = -3
      ELSE IF( NRHS.LT.0 ) THEN
         INFO = -4
      ELSE IF( LDA.LT.MAX( 1, M ) ) THEN
         INFO = -6
      ELSE IF( LDB.LT.MAX( 1, M, N ) ) THEN
         INFO = -8
      ELSE IF( LWORK.LT.MAX( 1, MN+MAX( MN, NRHS ) ) .AND. .NOT.LQUERY ) THEN
         INFO = -10
      END IF
```

cgo/lapack.go
```diff
@@ -228,7 +228,7 @@ func (impl Implementation) Dgeqrf(m, n int, a []float64, lda int, tau, work []fl
 // with native, lwork == -1 will not run Dgeqrf but will instead write the minimum
 // work necessary to work[0]. If len(work) < lwork, Dgeqrf will panic.
 func (impl Implementation) Dgels(trans blas.Transpose, m, n, nrhs int, a []float64, lda int, b []float64, ldb int, work []float64, lwork int) bool {
-	mn := min(m, n)
+	mn := max(m, n)
```
kortschak: It looks to me like the real fix is to change the line below (no comment available there) to `checkMatrix(max(m, n), nrhs, b, ldb)`, and revert this.

native/dgels.go
```diff
@@ -41,7 +41,7 @@ import (
 func (impl Implementation) Dgels(trans blas.Transpose, m, n, nrhs int, a []float64, lda int, b []float64, ldb int, work []float64, lwork int) bool {
 	notran := trans == blas.NoTrans
 	checkMatrix(m, n, a, lda)
-	mn := min(m, n)
+	mn := max(m, n)
```
kortschak: Same here.

btracey:
Yep. You're totally right. Sorry. Fixed.

btracey:
Again, I don't understand, one fails on 1.4, the other on 1.5rc1.

btracey:
Ah, I see what you mean by "real failure". It's a time out test failure, and doesn't seem to be consistent as to which test it fails on. Maybe we should increase the test timeout to 20minutes?

kortschak:
What concerns me is the fact that there is such a discrepancy in the amount of time it takes to complete cgo tests on travis compared to a local machine - it's seconds on my laptop and ~ten minutes on travis. This really indicates to me something is not right.

kortschak:
LGTM

btracey:
I agree that it's weird. It seems like it has to be a cgo problem or a travis problem. If it's a travis problem there's nothing we can do about it. I don't know how to test if it's a cgo problem. We're running on different operating systems and it's fine 100% of the time (so far) on my machine.

kortschak:
I think Seb's answer is probably the reality.

btracey:
Seb's answer?

kortschak:
https://groups.google.com/d/topic/golang-nuts/MYBwEq2Sg_c/discussion

PR #33, opened 2015-08-11
Remove unnecessary inputs to Trtrs in lapack64

btracey:


btracey:
PTAL @kortschak 

From what I can tell, travis is flukey. The failures have been in different releases.

kortschak:
It's odd that those were left in there.

LGTM - I've restarted the tests, so left them finish.

kortschak:
That failure on go1.5 looks like a real gc bug. We may want to try to replicate it on a local machine.

kortschak:
Passes go1.5rc1 on local machine.

btracey:
How do you tell it's real? I've restarted the tests a couple of times on this PR and the last, and it's varied which of the releases failed. Sometimes it's 1.5rc1, sometimes it's 1.3 and 1.4, sometimes it's all of them. They all fail with a timeout (like this one does). I don't know how to distinguish GC bug from just travis being flaky.

kortschak:
I've done many restarts and go1.5rc1 never passes. I have run tests on my machine here and it always passes with the same OpenBLAS and lapacke refs. I think it's more than flakiness - the cgo call never returns.

btracey:
Passes on 1.5rc1 on my machine. Seems to have passed this time too. If it's a gc bug I don't know how to debug it.

PR #36, opened 2015-08-20
Add dlarts

btracey:


btracey:
Not ready for submission. 

Dlarts is necessary for Dgecon (Compute condition number from QR facotrization) and Dtrcon (compute condition of triangular matrix, which can be used from LU to compute the condition number). 

In the normal case, this function mostly just calls Dtrsv. In the case where the matrix is badly conditioned, it does all sorts of crazy things. Any ideas on testing? For the badly conditioned cases in BLAS I created the ill conditioned cases by hand (for the Givens rotation cases). Here though, there are many possible code paths, so I can't construct a suffient set by hand. Additionally, it's meant to protect against bad scaling, so we can't compare the output to Dtrsv (because it shouldn't be the same), and I'm worried about testing that A * x = b, because presumably we would need also need to be careful with scaling. At the least, I'll need to do a second pass with visual inspection.

kortschak:
You should be able to use dvykov's go-fuzz to build a minimal corpus that tests all paths with some fiddling.

jonlawlor:
If you're going to try out fuzz, I think you'll first want to write byte encoders / decoders for the various blas types.

kortschak:
We have one for Dense already, the rest can be copied from a created Dense.

btracey:
> We have one

One what? A go-fuzz example, or a comparison for triangular solve?

kortschak:
We have a binary encoder. This means that a fuzzer can be written with a corpus of encoded matrices from the original handwritten test cases. The other matrix types can be created from the decoded Dense.

btracey:
Okay, I ran a fuzzer against it for a while so I think it's free of panics. Unfortunately, there is no cgo version of the algorithm so I cannot compare outputs.

PTAL

kortschak:
Pity Dlatrs doesn't exist in the cgo. I'll have a look some time over the weekend.

kortschak:
Change the PR title to reflect the function name?

native/dlatrs.go
```diff
@@ -0,0 +1,334 @@
...
+		panic(badUplo)
+	}
+	if trans != blas.Trans && trans != blas.NoTrans {
+		panic(badTrans)
+	}
+	if diag != blas.Unit && diag != blas.NonUnit {
+		panic(badDiag)
+	}
+	upper := uplo == blas.Upper
+	notrans := trans == blas.NoTrans
```
kortschak: s/notrans/noTrans/g
s/nonunit/nonUnit/g

native/dlatrs.go
```diff
@@ -0,0 +1,334 @@
...
+	nonunit := diag == blas.NonUnit
+
+	if n < 0 {
+		panic(nLT0)
+	}
+	checkMatrix(n, n, a, lda)
+	// Why incX isn't specified like every other blas/lapcak function I have no idea.
```
kortschak: s/lapcak/lapack/

Diatribe has merit, but could be more informative - make the concern clearer.

native/dlatrs.go
```diff
@@ -0,0 +1,334 @@
...
+	checkVector(n, cnorm, 1)
+
+	if n == 0 {
+		return
+	}
+	scale = 1
+	bi := blas64.Implementation()
+	if !normin {
+		if upper {
+			for j := 0; j < n; j++ {
```
kortschak: I take it the lda here is because we are row major.

native/dlatrs.go
```diff
@@ -0,0 +1,334 @@
...
+			jinc = -1
+		} else {
+			jfirst = 0
+			jlast = n - 1
+			jinc = 1
+		}
+		// Compute the growth in A * x = b.
+		if tscal != 1 {
+			grow = 0
+			goto Out
```
kortschak: s/Out/finish/g or some near equivalent.

native/dlatrs.go
```diff
@@ -0,0 +1,334 @@
...
+			for j := jfirst; j != jlast; j += jinc {
+				if grow <= smlnum {
+					goto Out
+				}
+				xj := 1 + cnorm[j]
+				grow /= xj
+			}
+		}
+	}
+Out:
```
kortschak: Newline before this.

native/dlatrs.go
```diff
@@ -0,0 +1,334 @@
...
+						bi.Dscal(n, rec, x, 1)
+						scale *= rec
+					}
+				} else if xj*cnorm[j] > bignum-xmax {
+					bi.Dscal(n, 0.5, x, 1)
+					scale *= 0.5
+				}
+				if upper {
+					if j > 0 {
+						bi.Daxpy(j, -x[j]*tscal, a[j:], lda, x, 1)
```
kortschak: Should this be a[:j]?

kortschak:
LGTM with minor comments.

native/dlatrs.go
btracey: Done.

native/dlatrs.go
btracey: Just removed it.

native/dlatrs.go
btracey: Yep. It's computing the sum over the column.

native/dlatrs.go
btracey: Done.

native/dlatrs.go
btracey: Done.

kortschak:
s/dlarts/dlatrs/ in name, but doesn't matter.

PR #37, opened 2015-09-04
Add drscl and tests, and fix the dlamch parameters while here.

btracey:


btracey:
The tests reach all three code paths in the for loop.

native/general.go
```diff
@@ -78,18 +78,15 @@ func max(a, b int) int {
 // fixed values. Probably a way to get them as constants.
 // TODO(btracey): Is there a better way to find the smallest number such that 1+E > 1
 
-var dlamchE, dlamchS, dlamchP float64
+var (
+	// dlamchE is the machine epsilon. For IEEE this is 2^-53.
+	dlamchE = math.Float64frombits(0x3ca0000000000000)
```
kortschak: It's a pity the binary exponent form was removed from the language.

native/drscl.go
```diff
@@ -0,0 +1,40 @@
...
+	cden := a
+	cnum := 1.0
+	smlnum := dlamchS
+	bignum := 1 / smlnum
+	for {
+		cden1 := cden * smlnum
+		cnum1 := cnum / bignum
+		var mul float64
+		var done bool
+		if math.Abs(cden1) > math.Abs(cnum) && cnum != 0 {
```
kortschak: Make this whole block a switch and move cnum != 0 to the front of this case.

kortschak:
LGTM with minor comments.

native/general.go
btracey: This is the first I've heard about it ever being in the language.

native/drscl.go
btracey: Done.

native/general.go
kortschak: It's not anymore. It was removed for 1.5. It's used by the compiler and not specified, so they removed it.

PR #38, opened 2015-09-04
Add dlacn2

btracey:


btracey:
This function is very iterative and so hard to test directly, and there's no cgo function to compare to. However, when this is merged along with the other two current PRs, they will all be called by Dgecon which we can test against cgo.

kortschak:
Ugh, that fortran spaghetti is horrible.

LGTM

btracey:
Yep. The original FORTRAN was worse.

kortschak:
I meant the NETLIB 3.5 code - yours is good.

kortschak:
Oops, isave should be a *[3]int or be returned, the mutations you are doing are not visible externally. Sorry I missed that.

PR #44, opened 2015-09-10
Add Dtrcon and test

btracey:


btracey:
I'm not fully sure what to do about this. The couple of tests I checked that failed gave the same answer in cgo and native (that is, the same different answer from Dgecon). I'm not sure what's wrong with the algorithm either, it looks pretty similar to the Dgecon algorithm.

btracey:
The reason to merge would be that this is how one checks the condition number from a QR factorization, and if it matches cgo at least it gives similar behavior.

testlapack/dtrcon.go
```diff
@@ -0,0 +1,183 @@
...
+			t.Errorf("Inf norm mismatch. Want %v, got %v.", test.condInf, condInf)
+		}
+		if !floats.Equal(aCopy, test.a) {
+			t.Errorf("a modified during call")
+		}
+	}
+
+	// Dtrcon does not match the Dgecon output in many cases. See
+	// https://github.com/xianyi/OpenBLAS/issues/636
+	// TODO(btracey): Reimplement this when
```
kortschak: Finish this sentence.

native/dtrcon.go
```diff
@@ -0,0 +1,79 @@
...
+		panic(badWork)
+	}
+	if len(iwork) < n {
+		panic(badWork)
+	}
+	if n == 0 {
+		return 1
+	}
+	bi := blas64.Implementation()
+	onenrm := norm == lapack.MaxColumnSum
```
kortschak: This is only used in one place. Just do the test there.

kortschak:
LGTM - maybe clean up the commented out code though first.

kortschak:
It seems when they disagree, the disagreement in in the anorm value when getting the 1-norm cond and in the ainvnorm when getting the inf-norm cond.

btracey:
Interesting, especially weird in the anorm part. I'll look into it when I get a chance (and get OpenBLAS compiled again...)

btracey:
At least in the one norm cases it has something to do with being sensitive to the values in work.

btracey:
Got it. Fix coming.

btracey:
That fixes the one norm cases. In the inf norm cases, it works every time where unit == true and diag == true. Otherwise it fails with low probability.

kortschak:
By how much?

btracey:
More than numeric noise. One such case is 
````
dtrcon.go:171: Inf norm mismatch. Upper = true, unit = false, want 0.06492931414228323, got 0.08734855675106225
````
Tyically it's wrong in the first or second significant digit.

btracey:
The fact that it's right a lot of the time suggests that none of the functions are grossly wrong. The fact that it always works for Upper and Unit suggests the function that is wrong has these. This implies to me that Dlatrs is wrong somewhere, which isn't that big of a surprise.

btracey:
Though, Dtrcon seems to match the cgo output (I'll have to check this), so it suggests something wrong with the algorithm and not the coding)

PR #45, opened 2015-09-10
Add Dpocon and tests

btracey:


btracey:
This one at least matches Dgecon. No deterministic at the start because the matrix multiplied by itself is singular.

cgo/lapack.go
```diff
@@ -524,6 +524,29 @@ func (impl Implementation) Dormqr(side blas.Side, trans blas.Transpose, m, n, k
...
+	if uplo != blas.Upper && uplo != blas.Lower {
+		panic(badUplo)
+	}
+	if len(work) < 3*n {
+		panic(badWork)
+	}
+	if len(iwork) < n {
+		panic(badWork)
+	}
+	clapack.Dpocon(uplo, n, a, lda, anorm, rcond)
```
kortschak: Missing return.

kortschak:
LAPACKE seems to be pretty unhealthy.

btracey:
As a library or as something that works with Travis?

kortschak:
https://github.com/xianyi/OpenBLAS/issues/639

btracey:
Ah. Not good. This kind of thing makes me think we're the only ones who have tested a bunch of this interface ...

I'd like to just say "disable the cgo test", but I'm more worried about that here since we actually plan on calling this function directly. The cgo tests also have the LQ factorizaton test commented out because of a bug. That's a significant chuck of Solve that won't work with a c-based lapack. I guess the answer is to make sure that the mat64 tests are robust, so the issues are visible on that level? I don't want to abandon clapack altogether because it's a useful check, but at the same time it's frustrating how many bugs we're finding (in such a widely used interface at that!)

kortschak:
Indeed. I'm going to spend a little time on seeing if I can figure out why it's broken.

I think we should disable the cgo test in the interim.

btracey:
Test disabled.

kortschak:
Fix found. You will not believe it.

btracey:
Oh no! So sad...

kortschak:
LGTM

kortschak:
The cgo failures on my machine:

```
--- FAIL: TestDpocon (0.01s)
	dpocon.go:126: Cond mismatch. Want 0.056005662446310195, got 0.021282043474993715.
	dpocon.go:126: Cond mismatch. Want 0.09149436558852704, got 0.09181637098950025.
```

Do you want to send a PR to OpenBLAS for the fix you have?

kortschak:
BTW What is the bug that's blocking Dormlq?

btracey:
I'm not sure what you mean by "The fix I have". The Go version just works, and I haven't been able to test this with cgo because of the openblas failures.

LQ problem is https://github.com/xianyi/OpenBLAS/issues/615

kortschak:
Oh, sorry, conflating different threads (Dtrcon).

PR #50, opened 2015-09-11
Add Dlanxx routines to lapack64 interface.

btracey:


btracey:
Needed for computing condition numbers also.

lapack64/lapack64.go
```diff
@@ -178,6 +178,34 @@ func Getrs(trans blas.Transpose, a blas64.General, b blas64.General, ipiv []int)
...
+//  lapack.MaxRowSum: the maximum row sum of the absolute values of the entries.
+//  lapack.Frobenius: the square root of the sum of the squares of the entries.
+// If norm == lapack.MaxColumnSum, work must be of length n, and this function will panic otherwise.
+// There are no restrictions on work for the other matrix norms.
+func Lange(norm lapack.MatrixNorm, a blas64.General, work []float64) float64 {
+	return lapack64.Dlange(norm, a.Rows, a.Cols, a.Data, a.Stride, work)
+}
+
+// Dlansy computes the specified norm of an n×n symmetric matrix. If
+// norm == lapack.MaxColumnSum or norm == lapackMaxRowSum work must have length
```
kortschak: Above you say there's a panic, but here you don't. What is our system?

lapack64/lapack64.go
```diff
@@ -178,6 +178,34 @@ func Getrs(trans blas.Transpose, a blas64.General, b blas64.General, ipiv []int)
...
+// at least n, otherwise work is unused.
+func Lansy(norm lapack.MatrixNorm, a blas64.Symmetric, work []float64) float64 {
+	return lapack64.Dlansy(norm, a.Uplo, a.N, a.Data, a.Stride, work)
+}
+
+// Lantr computes the specified norm of an m×n trapezoidal matrix A. If
+// norm == lapack.MaxColumnSum work must have length at least n, otherwise work
```
kortschak: Same here.

kortschak:
LGTM

lapack64/lapack64.go
btracey: I think it should be self evident that the enum-like types need to be one of the valid values. If the function is restricted to a subset of them, then it should be documented. Sound good?

btracey: Oh, missed the point. I'm adding the panic to the comments. We do it everywhere so far.

PR #57, opened 2015-10-02
Add Dlacpy and tests

btracey:


btracey:
@kortschak , this has three options for uplo. blas.Upper, blas.Lower, and blas.All. I'm not sure what the correct fix to the generator is (nor how special case this function is).

kortschak:
#58 should fix that.

btracey:
Thanks.

kortschak:
This gives me an understanding of why the assembler does know all the instruction sets.

btracey:
Do you mean doesn't?

kortschak:
Indeed I do.

kortschak:
LGTM after rebase onto master and cgo tests pass.

PR #59, opened 2015-10-06
Add dlasr

btracey:


btracey:
Better documentation and more checks coming.

I'm not sure how to test this other than to rewrite the code again. However, this code will be called by a bunch of routines so it should be effectively tested then.

btracey:
Good news: Just added tests for Dlasr
Bad news: It exposed a mismatch between the condition number computation for Dgecon and Dpocon that's significant (0.021 to 0.024 while all of the others match to 1e-14)
Worse news: Cgo and native give the same wrong answer.

Suggestions? Log a bug with OpenBLAS, and comment out the tests I've added for that case. What about the dpocon script itself. It's generating random cases at present.

kortschak:
>Worse news: Cgo and native give the same wrong answer.

This is not good, but it's not terrible. At least the issue is consistent and code that work elsewhere works just as well here.

>Suggestions? Log a bug with OpenBLAS, and comment out the tests I've added for that case.

I think so.

>What about the dpocon script itself? It's generating random cases at present.

What do you mean?

btracey:
The added case arises when generating the random test cases (at least with the random number stream on my machine). In order to pass the tests this case shouldn't be generated. I could, for example, make the loop shorter, but that dousn't feel like the right solution.

kortschak:
Maybe log cases >1e-14 and fail case >1e-2?

btracey:
Filed https://github.com/xianyi/OpenBLAS/issues/664

btracey:
Filed http://icl.cs.utk.edu/lapack-forum/viewtopic.php?f=13&t=4817

btracey:
Implemented suggestions.

native/dlasr.go
```diff
@@ -0,0 +1,232 @@
...
+// license that can be found in the LICENSE file.
+
+package native
+
+import (
+	"github.com/gonum/blas"
+	"github.com/gonum/lapack"
+)
+
+// Dlasr applies a sequence of plane rotations to the m×n matrix A. P(k) is a 2×2
```
kortschak: The docs in LAPACK are very informative. Could this be expanded?

testlapack/dgecon.go
```diff
@@ -44,6 +44,20 @@ func DgeconTest(t *testing.T, impl Dgeconer) {
 			condOne: 0.024740155174938,
 			condInf: 0.012034465570035,
 		},
+		/*
+			// Dgecon does not match Dpocon for this case. https://github.com/xianyi/OpenBLAS/issues/664.
+			{
```
kortschak: The comment for this could be more informative. Same for the one in dpocon, though that one makes more sense; at first glance it is difficult to understand why dpocon is relevant to the dgecon test. It would be good to have a bit more explanation about what is expected and what is wrong.

Am I right that dgecon is wrong here since the inf-norm doesn't match the 1-norm and the matrix is symmetric?

kortschak: Additionally, is there a reason these don't get the same near miss treatment as the randomised tests?

kortschak:
LGTM after test comment improved. I'll leave the doc comment up to you.

native/dlasr.go
btracey: Done.

testlapack/dgecon.go
btracey: Good points. Updated.

testlapack/dgecon.go
```diff
@@ -64,11 +77,25 @@ func DgeconTest(t *testing.T, impl Dgeconer) {
...
 		}
-		if math.Abs(condInf-test.condInf) > 1e-13 {
-			t.Errorf("Inf norm mismatch. Want %v, got %v.", test.condInf, condInf)
+		if !floats.EqualWithinAbsOrRel(condInf, test.condInf, 1e0, 1e0) {
+			t.Errorf("One norm mismatch. Want %v, got %v.", test.condInf, condInf)
+		} else if !floats.EqualWithinAbsOrRel(condInf, test.condInf, 1e-14, 1e-14) {
+			log.Printf("Dgecon one norm mismatch. Want %v, got %v.", test.condInf, condInf)
 		}
+		/*
+			if math.Abs(condOne-test.condOne) > 1e-13 {
```
kortschak: Why is this here?

PR #60, opened 2015-10-13
Add Dlasrt

btracey:


native/dlasrt.go
```diff
@@ -0,0 +1,27 @@
...
+// the elements are sorted in increasing order. If sort == lapack.SortDecreasing,
+// the elements are sorted in decreasing order.
+func (impl Implementation) Dlasrt(s lapack.Sort, n int, d []float64) {
+	checkVector(n, d, 1)
+	d = d[:n]
+	switch s {
+	default:
+		panic("lapack: bad sort")
+	case lapack.SortIncreasing:
+		sort.Sort(sort.Reverse(sort.Float64Slice(d)))
```
kortschak: Can we do this by a normal sort and then a simple slice reverse function written here? That avoids the double interface resolution that what's here costs.

btracey: Sure. I thought this was the idiomatic code based on the example.

kortschak: See what the performance difference is?

btracey: Can we leave it as is now and re-assess when we have the functions that call it in context?  This way we have a better guess on typical sizes.

kortschak: No worries.

kortschak:
LGTM

PR #61, opened 2015-10-13
Adddlas2

btracey:


testlapack/dlas2.go
```diff
@@ -0,0 +1,33 @@
...
+
+type Dlas2er interface {
+	Dlas2(f, g, h float64) (min, max float64)
+}
+
+func Dlas2Test(t *testing.T, impl Dlas2er) {
+	for i, test := range []struct {
+		f, g, h, ssmin, ssmax float64
+	}{
+		{10, 30, 12, 3.567778859365365, 33.634371616111189},
```
kortschak: Values from...?

kortschak:
LGTM

testlapack/dlas2.go
btracey: Octave

kortschak: Would you write that in.

PR #64, opened 2015-10-15
Add Dorg2r and test

btracey:


btracey:
@kortschak do you see anything wrong with my translation? The tests don't pass. I don't understand if this is because my code is wrong, or if I don't understand what the code is doing. I'm pretty sure the "constructQ" function is correct, because it's used to verify QR. It's a weird function because the real Q is an mxm matrix, but it's somehow being condensed into an mxn matrix, so maybe it's not just the first n columns like the documentation seems to suggest.

kortschak:
I can't see anything obviously wrong, but it feels to me like an off-by-one from changing the indexing from 1 to 0. Note that the first column of the failing case is correct.

btracey:
I agree, but I don't see it at the moment. Do you agree with my interpretation of the function (that it computes the first N columns of Q?). If you do, I can keep looking / explore the diff.

kortschak:
In the docs, should the H indices be from 0 or 1?

btracey:
The Hs start in column 0, so it should be 0-indexed, yes. 

the constructQ function in test_lapack shows how the h's work.

kortschak:
I agree with your interpretation of the docs.

kortschak:
I've had a look at exposing dorg2r in OpenBLAS and it's not trivial.

btracey:
That's unfortunate.

I implemented Dorgl2. It has the same problems (first column matches, other does not). There is obviously me as a common implementer, but still.

btracey:
We can just compare to the output of Dorgqr and Dorglq (the blocked version of the algorithm). I'll try looking at that when I get a chance.

btracey:
cgo Dorglq and Dorgqr pass our tests in testlapack. Back to the bug hunting.

btracey:
Ah ha! Found it. I'll update documentation when I get a chance.

native/dorg2r.go
```diff
@@ -0,0 +1,61 @@
...
+		}
+		if i < n-1 {
+			a[i*lda+i] = 1
+			impl.Dlarf(blas.Left, m-i, n-i-1, a[i*lda+i:], lda, tau[i], a[i*lda+i+1:], lda, work)
+		}
+		if i < m-1 {
+			bi.Dscal(m-i-1, -tau[i], a[(i+1)*lda+i:], lda)
+		}
+		a[i*lda+i] = 1 - tau[i]
+		for l := 0; l <= i-1; l++ {
```
kortschak: `l < i`

native/dorgl2.go
```diff
@@ -0,0 +1,53 @@
...
+	for i := k - 1; i >= 0; i-- {
+		if i < n-1 {
+			if i < m-1 {
+				a[i*lda+i] = 1
+				impl.Dlarf(blas.Right, m-i-1, n-i, a[i*lda+i:], 1, tau[i], a[(i+1)*lda+i:], lda, work)
+			}
+			bi.Dscal(n-i-1, -tau[i], a[i*lda+i+1:], 1)
+		}
+		a[i*lda+i] = 1 - tau[i]
+		for l := 0; l <= i-1; l++ {
```
kortschak: `l < i`

cgo/lapack_test.go
```diff
@@ -84,6 +84,18 @@ func (d blockedTranslate) Dorml2(side blas.Side, trans blas.Transpose, m, n, k i
 	impl.Dormlq(side, trans, m, n, k, a, lda, tau, c, ldc, work, len(work))
 }
 
+func (d blockedTranslate) Dorg2r(m, n, k int, a []float64, lda int, tau, work []float64) {
+	impl.Dorgqr(m, n, k, a, lda, tau, work)
+}
+
+func TestDorglq(t *testing.T) {
+	testlapack.Dorgl2Test(t, blockedTranslate{impl})
```
kortschak: This needs a different shim or an additional method to the current one:

```
cgo/lapack_test.go:92: cannot use blockedTranslate literal (type blockedTranslate) as type testlapack.Dorgl2er in argument to testlapack.Dorgl2Test:
	blockedTranslate does not implement testlapack.Dorgl2er (missing Dorgl2 method)
```

kortschak:
Code LGTM with minor comments and cgo test fix.

cgo/lapack_test.go
kortschak: Presumably this is the right thing:

```
func (d blockedTranslate) Dorgl2(m, n, k int, a []float64, lda int, tau, work []float64) {
	impl.Dorglq(m, n, k, a, lda, tau, work)
}
```

Dlantr is panicking again.

btracey: On my machine, Dlantr doesn't panic. It does intermittently fail, but because of Dlange giving the wrong result.

vladimir-ch: In `lapack/native` on `adddorg2r` I get no panics and no failures (except for Dgecon). Am I doing anything differently?

kortschak: The panics are in clapack calls into OpenBLAS due to LAPACKE.

btracey: But only sometimes on some machines. Works fine for me.

btracey: Done.

native/dorg2r.go
btracey: Done.

native/dorgl2.go
btracey: Done.

btracey:
The next step is to add Dorgqr and Dorglq. I'll add comments to the cgo routines at that point.

PR #67, opened 2015-10-27
Add dorgqr and tests, update dorg2r tests, and fix cgo dorgqr

btracey:


btracey:
@kortschak Would you mind taking a look? There must be another indexing error, but I can't find it. I hacked the values of nb and nx so that the blocked code can be reached with smaller matrices.

native/dorgqr.go
```diff
@@ -0,0 +1,97 @@
...
+// amount of blocking is limited by the usable length. If lwork == -1, instead of
+// computing Dorg2r the optimal work length is stored into work[0].
+//
+// Dorgqr will panic if the conditions on input values are not met.
+func (impl Implementation) Dorgqr(m, n, k int, a []float64, lda int, tau, work []float64, lwork int) {
+	nb := impl.Ilaenv(1, "DORGQR", " ", m, n, k, -1)
+	nb = 2
+	// work is treated as an n×nb matrix
+	if lwork == -1 {
+		work[0] = float64(n * nb)
```
kortschak: float64(max(1, n) * nb)

native/dorgqr.go
```diff
@@ -0,0 +1,97 @@
...
+	for i := ki; i >= 0; i -= nb {
+		ib := min(nb, k-i)
+		if i+ib < n {
+			impl.Dlarft(lapack.Forward, lapack.ColumnWise, m-i, ib, a[i*lda+i:], lda, tau[i:], work, ldwork)
+			impl.Dlarfb(blas.Left, blas.NoTrans, lapack.Forward, lapack.ColumnWise, m-i, n-i-ib, ib, a[i*lda+i:], lda, work, ldwork, a[i*lda+i+ib:], lda, work[ib:], ldwork)
+		}
+		impl.Dorg2r(m-i, ib, ib, a[i*lda+i:], lda, tau[i:], work)
+		// Set rows 0:i-1 of current block to zero
+		for j := i; j < i+ib; j++ {
+			for l := 0; l < i; l++ {
```
kortschak: Should this be `l < i-1`? It doesn't fix the problem.

native/dorgqr.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+	if kk == 0 {
+		return
+	}
+	// Perform the operation on column-blocks
+	for i := ki; i >= 0; i -= nb {
+		ib := min(nb, k-i)
+		if i+ib < n {
+			impl.Dlarft(lapack.Forward, lapack.ColumnWise, m-i, ib, a[i*lda+i:], lda, tau[i:], work, ldwork)
+			impl.Dlarfb(blas.Left, blas.NoTrans, lapack.Forward, lapack.ColumnWise, m-i, n-i-ib, ib, a[i*lda+i:], lda, work, ldwork, a[i*lda+i+ib:], lda, work[ib:], ldwork)
```
kortschak: The fortran indexes into work here. I tried that and it doesn't fix things.

btracey: Fortran indexes into the second time work is used, just like it is here.

kortschak:
I can't see anything other than the two things above. Neither of them fix the problem.

native/dorgqr.go
btracey: No, I don't think so. For one, this was the bug in dorm2r that took a while to find. Two, i is one indexed in the fortran code, and zero indexed in the go code. Normally, one is subtracted from both of the index bounds, but that has already happened in the shift of i. Thus, I think the bounds should be as they are here.

native/dorgqr.go
kortschak: Too early for me obviously.

native/dorgqr.go
kortschak: Yep, OK.

btracey:
Bleh. Sad times.

native/dorgqr.go
kortschak: Maybe split that line.

btracey: Aha! The problem is that work is a matrix, so work(ib) is really work(ib, 1). Since we're row indexed....

Fixed!

native/dorgqr.go
btracey: Done.

btracey:
Tests now pass, and cgo comment updated. 

PTAL.

btracey:
Actually, the algorithms are virtually identical, so I'm going to include the lq version in this as well.

btracey:
Okay. Pushed. PTAL.

kortschak:
LGTM

btracey:
Failures do not seem related to this PR and tests pass locally.

PR #68, opened 2015-11-07
Adddlasq routines and tests

btracey:


btracey:
This is not ready for submission, but pretty close to get the right idea.

Rather than the loops at the bottom just checking for infinite loops (which I had for a while), I'd like to turn that into a test of the behavior.

The dlasq2 tests having a tolerance of 1e-6 I think is because of single precision issues I couldn't figure out. Fortran wasn't giving the same answer depending if I generated the numbers randomly vs. actually giving it the number.

There are a lot of ugly parts of the code. The gotos are highly nontrivial. At least for now I'd like to leave all the weird loop index stuff. If we discover more bugs down the line it's way easier to debug this way than if the code is idiomatic go. 

I also have all the debugging scripts. I assume we don't want them in the tree, but they'll be very useful if we find bugs in the future.

The coverage of paths on this is decent. It seems like the unused paths come in for underflow.

btracey:
My reading of the matrices is an is the construction of triDi. However, this constructed matrix is not symmetric (let alone positive definite). Am I interpreting something wrong? Does "The tridiagonal is L*U or, if you prefer, the symmetric tridiagonal to which it is similar." mean one needs to transform L*U?

kortschak:
I'm a little confused by this PR. The additions, I'll take a look at tomorrow, but I don't understand what's happening with Dorglq and Dorgqr in cgo.

I think that we may want the debugging code available - I was actually thinking of adding a cgo/internal/fortran to add utilities to help with direct use of fortran calls that are made available via the CL Seb proposed. This would fit in the same space.

BTW I edited your comment above so the \* does not get interpretted as *italic*. That was very confusing.

btracey:
I am also confused by those changes. Will fix.

I can add the debugging code

btracey:
Part of the problem was my reading of the documentation for U. However, that doesn't solve the problem. Opened Stack Overflow question where maybe someone will see it. http://stackoverflow.com/questions/33614240/lapack-dlasq2-matrix-input-format
Do you happen to understand @xianyi ?

My debugging code is all strictly Fortran with some print statements used to translate between fortran and go inputs. Should I include that code as a subfolder of lapack/native, lapack/testlapack, or somewhere else?

btracey:
@kortschak I don't understand. If I do 
````
git diff origin/master:native/dorglq.go native/dorglq.go
````
There is no difference shown. Is this a bug with the github interface, or do I not know how to use git properly?

**Comment by kortschak** [comment]
Neither. There is no difference in native/dorglq.go between those two branches. Though, the command can be more easily said as `git diff origin/master adddlasq native/dorgql.go` assuming you want diff against your local adddlasq branch.

**Comment by btracey** [comment]
Oh, I see, it's in the c interface. Sorry.

I'll check that, and I'll clean up the Fortran code to remove a lot of the unnecessary print statements and comments. Otherwise I think the Go code is good enough to go. Working up to the full SVD will allow us to have better tests of the full integrated system.

**Comment by btracey** [comment]
Okay, cleaned up the Fortran code.

PTAL

I know there's still a bunch of ugly code there, but I'd prefer to leave it as is until we have integrated SVD tests if that's okay.

File: internal/dlasqtest/commands.txt [hunk 1]
```diff
@@ -0,0 +1,9 @@
+Advice: nDiv in dlasq2 can be an effective counter for isolating a bug. Using os.Exit(1) and "STOP" when nDiv hits a certain number.
```
kortschak: Make this something that can be executed, or at least sourced - prefix this line with `#`.

kortschak:
It seems to me that `internal/dlasqtest` is more meaningfully named `testdata/dlasqtest` or something else.

kortschak:
Are the whitespace diffs between the fortran code in dlasqtest and OpenBLAS/lapack-netlib/SRC there intentionally or residually? Just curious.

internal/dlasqtest/dlasq6.f
```diff
@@ -0,0 +1,258 @@
...
+      INTRINSIC          MIN
+*     ..
+*     .. Executable Statements ..
+*
+      IF( ( N0-I0-1 ).LE.0 )
+     $   RETURN
+*
+
+      print *, "In dlasq6"
+      STOP
```
kortschak: Should this still be here?

internal/dlasqtest/testdlasq1.f90
```diff
@@ -0,0 +1,636 @@
...
+4.5015392507594826D-01, &
+3.3897738839543617D-01, &
+4.7249205225111501D-01, &
+9.8599436000817042D-01/)
+n =  100
+info = 0
+
+
+open(unit = 4, STATUS='REPLACE', file = "gen4tests.txt")
+open(unit = 3,STATUS='REPLACE', file = "gen3tests.txt")
```
kortschak: Spaces after ",".

internal/dlasqtest/testdlasq1.f90
```diff
@@ -0,0 +1,636 @@
...
+do i = 1,n
+    print *, d(i)
+end do
+
+close(5)
+close(3)
+close(4)
+
+end
+
```
kortschak: Kill trailing empty lines. I'll ignore the paired empty lines above ;)

native/dlasq1.go
```diff
@@ -0,0 +1,81 @@
...
+package native
+
+import (
+	"math"
+
+	"github.com/gonum/blas/blas64"
+	"github.com/gonum/lapack"
+)
+
+// DLASQ1 computes the singular values of an n×n bidiagonal matrix with diagonal
```
kortschak: Dlasq1

native/dlasq1.go
```diff
@@ -0,0 +1,81 @@
...
+// d and off-diagonal e. The input work must have length 4*n and Dlasq1 will panic
```
kortschak: Explicitly state the length requirements of d and e and check in the preamble.

native/dlasq2.go
```diff
@@ -0,0 +1,326 @@
...
+
+package native
+
+import (
+	"math"
+
+	"github.com/gonum/lapack"
+)
+
+// Dlasq2 computes all the eigenvalues of the symmetric positive
```
kortschak: This sentence has a semi-duplicated fragment.

native/dlasq2.go
```diff
@@ -0,0 +1,326 @@
...
+// relative accuracy are computed to high relative accuracy, in the
+// absence of denormalization, underflow and overflow.
+//
+// To see the relation of Z to the tridiagonal matrix, let L be a
+// unit lower bidiagonal matrix with subdiagonals Z(2,4,6,,..) and
+// let U be an upper bidiagonal matrix with 1's above and diagonal
+// Z(1,3,5,,..). The tridiagonal is L*U or, if you prefer, the
+// symmetric tridiagonal to which it is similar.
+//
+// info returns a status error.
```
kortschak: Specify the returnable values of info and what they mean.

native/dlasq2.go
```diff
@@ -0,0 +1,326 @@
...
+func (impl Implementation) Dlasq2(n int, z []float64) (info int) {
+
+	if len(z) < 4*n {
+		panic("lapack: bad z")
+	}
+	const (
```
kortschak: Single line.

native/dlasq2.go
```diff
@@ -0,0 +1,326 @@
...
+//
+// To see the relation of Z to the tridiagonal matrix, let L be a
+// unit lower bidiagonal matrix with subdiagonals Z(2,4,6,,..) and
+// let U be an upper bidiagonal matrix with 1's above and diagonal
+// Z(1,3,5,,..). The tridiagonal is L*U or, if you prefer, the
+// symmetric tridiagonal to which it is similar.
+//
+// info returns a status error.
+func (impl Implementation) Dlasq2(n int, z []float64) (info int) {
+
```
kortschak: Delete line.

native/dlasq2.go
```diff
@@ -0,0 +1,326 @@
...
+	if len(z) < 4*n {
+		panic("lapack: bad z")
+	}
+	const (
+		cbias = 1.5
+	)
+
+	info = 0
```
kortschak: Unnecessary assignment.

native/dlasq3.go
```diff
@@ -0,0 +1,155 @@
+// Copyright ©2015 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package native
+
+import "math"
+
+// Dlasq3 checks for deflation, computes a shift (TAU) and calls dqds.
```
kortschak: tau

native/dlasq3.go
```diff
@@ -0,0 +1,155 @@
...
+// In case of failure it changes shifts, and tries again until output
+// is positive.
+func (impl Implementation) Dlasq3(i0, n0 int, z []float64, pp int, dmin, sigma, desig, qmax float64, nFail, iter, nDiv int, ttype int, dmin1, dmin2, dn, dn1, dn2, g, tau float64) (
+	i0Out, n0Out, ppOut int, dminOut, sigmaOut, desigOut, qmaxOut float64, nFailOut, iterOut, nDivOut, ttypeOut int, dmin1Out, dmin2Out, dnOut, dn1Out, dn2Out, gOut, tauOut float64) {
+	// Note: documentation says dmin is an output only, but if that is so it is
+	// used before it is set.
+	const (
```
kortschak: Single line.

native/dlasq3.go
```diff
@@ -0,0 +1,155 @@
...
+package native
+
+import "math"
+
+// Dlasq3 checks for deflation, computes a shift (TAU) and calls dqds.
+// In case of failure it changes shifts, and tries again until output
+// is positive.
+func (impl Implementation) Dlasq3(i0, n0 int, z []float64, pp int, dmin, sigma, desig, qmax float64, nFail, iter, nDiv int, ttype int, dmin1, dmin2, dn, dn1, dn2, g, tau float64) (
+	i0Out, n0Out, ppOut int, dminOut, sigmaOut, desigOut, qmaxOut float64, nFailOut, iterOut, nDivOut, ttypeOut int, dmin1Out, dmin2Out, dnOut, dn1Out, dn2Out, gOut, tauOut float64) {
+	// Note: documentation says dmin is an output only, but if that is so it is
```
kortschak: The documentation ...

native/dlasq4.go
```diff
@@ -0,0 +1,228 @@
...
+// Dlasq4 computes an approximation to the smallest eigenvalue using values of d
+// from the previous transform.
+// i0, n0, and n0in are zero-indexed.
+func (impl Implementation) Dlasq4(i0, n0 int, z []float64, pp int, n0in int, dmin, dmin1, dmin2, dn, dn1, dn2, tau float64, ttype int, g float64) (tauOut float64, ttypeOut int, gOut float64) {
+	const (
+		cnst1 = 0.563
+		cnst2 = 1.01
+		cnst3 = 1.05
+
+		cnstthird = 0.333 // TODO(btracey): Fix?
```
kortschak: Yes. Later.

testlapack/dlasq1.go
```diff
@@ -0,0 +1,109 @@
...
+	"fmt"
+	"math"
+	"math/rand"
+	"testing"
+
+	"github.com/gonum/blas"
+	"github.com/gonum/blas/blas64"
+)
+
+func printFortranArray(z []float64, name string) {
```
kortschak: These fortran output mimics could go in there own file with more documentation about why they exist.

testlapack/dlasq1.go
```diff
@@ -0,0 +1,109 @@
...
+				if i != n-1 {
+					m[(i+1)*ldm+i] = e[i]
+				}
+			}
+
+			ldmm := n
+			mm := make([]float64, n*ldmm)
+			bi.Dgemm(blas.Trans, blas.NoTrans, n, n, n, 1, m, ldm, m, ldm, 0, mm, ldmm)
+
+			//printDlasq1FortranInput(d, e, work, n)
```
kortschak: Comment why this exists if it should stay.

kortschak:
I can't comment on testlapack/dlasq2-5 because of github limits presumably. So...

dlasq2.go L26 - delete commented out interface

dlasq2.go L638 - do these need to stay?

dlasq3.go L2643 - use t.Error or show values

dlasq3.go L2646-2697 - use idiomatic test failure phrasing

dlasq4.go L24-54 - move these out to a fortran helper file. If you make a fortran64 type with a String method the code for string representation of float64 will read better.

All - add copyright header.

kortschak:
What is native/c.out?

native/dlasq1.go
```diff
@@ -0,0 +1,81 @@
...
+// otherwise. On exit, d contains the singular values in decreasing order, and
+// e is overwritten.
+func (impl Implementation) Dlasq1(n int, d, e, work []float64) (info int) {
```
kortschak: At some point I think we should replace info with an error.

native/dlasq1.go
```diff
@@ -0,0 +1,81 @@
...
+		sigmx = math.Max(sigmx, d[i])
+	}
+	eps := dlamchP
+	safmin := dlamchS
+	scale := math.Sqrt(eps / safmin)
+	bi := blas64.Implementation()
+	bi.Dcopy(n, d, 1, work, 2)
+	bi.Dcopy(n-1, e, 1, work[1:], 2)
+
+	_ = scale
```
kortschak: Delete.

native/dlasq1.go
```diff
@@ -0,0 +1,81 @@
...
+	impl.Dlascl(lapack.General, 0, 0, sigmx, scale, 2*n-1, 1, work, 1)
+	for i := 0; i < 2*n-1; i++ {
+		v := work[i]
```
kortschak: work[i] *= work[i]

native/dlasq2.go
```diff
@@ -0,0 +1,326 @@
...
+// unit lower bidiagonal matrix with subdiagonals Z(2,4,6,,..) and
+// let U be an upper bidiagonal matrix with 1's above and diagonal
+// Z(1,3,5,,..). The tridiagonal is L*U or, if you prefer, the
+// symmetric tridiagonal to which it is similar.
+//
+// info returns a status error.
+func (impl Implementation) Dlasq2(n int, z []float64) (info int) {
+
+	if len(z) < 4*n {
+		panic("lapack: bad z")
```
kortschak: I know that this is phrased similarly to "bad work", but it's very similar to the message below. Can one or both be changed to make a clearer distinction. Maybe change those below to "negative z value" (use a constant).

native/dlasq2.go
```diff
@@ -0,0 +1,326 @@
...
+// absence of denormalization, underflow and overflow.
+//
+// To see the relation of Z to the tridiagonal matrix, let L be a
+// unit lower bidiagonal matrix with subdiagonals Z(2,4,6,,..) and
+// let U be an upper bidiagonal matrix with 1's above and diagonal
+// Z(1,3,5,,..). The tridiagonal is L*U or, if you prefer, the
+// symmetric tridiagonal to which it is similar.
+//
+// info returns a status error.
+func (impl Implementation) Dlasq2(n int, z []float64) (info int) {
```
kortschak: Later we should change the return type to an error.

kortschak:
I'm going to come back to dlasq2- later. github PR commenting on this PR is as responsive as a snail on valium and it's a bit draining.

native/dlasq2.go
```diff
@@ -0,0 +1,326 @@
...
+		e += z[k+1]
+		qmax = math.Max(qmax, z[k])
+		emin = math.Min(emin, z[k+1])
+		zmax = math.Max(math.Max(qmax, zmax), z[k+1])
+	}
+	if z[2*(n-1)] < 0 {
+		panic("lapack: bad z value")
+	}
+	d += z[2*(n-1)]
+	zmax = math.Max(qmax, zmax)
```
kortschak: Should there not be a `qmax = math.Max(qmax, z[2*n-1])` before this line?

native/dlasq2.go
```diff
@@ -0,0 +1,326 @@
...
+		impl.Dlasrt(lapack.SortDecreasing, n, z)
+		z[2*(n-1)] = d
+		return info
+	}
+	trace := d + e
+	if trace == 0 {
+		z[2*(n-1)] = 0
+		return info
+	}
+	for k := 2 * n; k >= 2; k -= 2 {
```
kortschak: Blank line here.

native/dlasq2.go
```diff
@@ -0,0 +1,326 @@
...
+		z[2*k-1] = 0
+		z[2*k-2] = z[k-1]
+		z[2*k-3] = 0
+		z[2*k-4] = z[k-2]
+	}
+	i0 := 0
+	n0 := n - 1
+
+	if cbias*z[4*i0] < z[4*n0] {
```
kortschak: Can you explain what's going on here with the change in indexing you've made cf the fortran?

btracey:
> It seems to me that internal/dlasqtest is more meaningfully named testdata/dlasqtest or something else.

I moved it into internal/testdata/dlasqtest. It's not go, but internal is a good sign of "here be dragons", which is definitely the case with the code here.

btracey:
Whitespace diffs are all residual. There was a lot of debugging.

btracey:
c.out was a coverprofile I forgot to delete. It's gone.

btracey:
Nice idea on the Fortran64 type

internal/dlasqtest/commands.txt
btracey: Done.

internal/dlasqtest/dlasq6.f
btracey: I'm not sure. None of the random test cases I generated ever reached dlasq6. My error rate was low in translation, but not non-zero (and sometimes in tricky ways). Part of me wants a trigger in dlasq6.go so when a case is found we can debug it.

internal/dlasqtest/testdlasq1.f90
btracey: I miss gofmt already.

internal/dlasqtest/testdlasq1.f90
btracey: Done.

native/dlasq1.go
btracey: Done.

native/dlasq1.go
btracey: Done.

native/dlasq1.go
btracey: added TODO

native/dlasq1.go
btracey: Done.

native/dlasq1.go
btracey: Clever.

native/dlasq2.go
btracey: Done.

native/dlasq2.go
btracey: Added TODO

native/dlasq2.go
btracey: Done.

native/dlasq2.go
btracey: Done.

native/dlasq2.go
btracey: Done.

native/dlasq2.go
btracey: Yes, thanks.

native/dlasq2.go
btracey: Done.

native/dlasq2.go
btracey: i0 is 1-indexed in Fortran and 0-indexed in go
z[4*i0-3] --> z[4*(i0+1)-3-1] --> z[4*i0]
Is that what you mean?

native/dlasq3.go
btracey: Done.

native/dlasq3.go
btracey: Done.

native/dlasq3.go
btracey: Done.

testlapack/dlasq1.go
btracey: Done.

testlapack/dlasq1.go
btracey: Gone.

native/dlasq2.go
kortschak: Thanks. For some of the less obvious identities like this, it might be worth adding commentary.

native/dlasq2.go
```diff
@@ -0,0 +1,330 @@
...
+// info returns a status error. If info is 0, the algorithm completed
+// successfully. If info is 1, a split was marked by a positive value in e.
+// If info is 2, current block of Z not diagonalized after 100*n
+// iterations (in inner while loop).  On exit Z holds a qd array with the same
+// eigenvalues as the given Z. If info is 3,  termination criterion of outer
+// while loop not met (program created more than N unreduced blocks).
+func (impl Implementation) Dlasq2(n int, z []float64) (info int) {
+	// TODO(btracey): make info an error.
+	if len(z) < 4*n {
+		panic("lapack: insufficient z length")
```
kortschak: use badZ

native/dlasq2.go
```diff
@@ -0,0 +1,330 @@
...
+	tol2 := tol * tol
+	if n < 0 {
+		panic(nLT0)
+	}
+	if n == 0 {
+		return info
+	}
+	if n == 1 {
+		if z[0] < 0 {
+			panic("lapack: negative z value")
```
kortschak: Use a constant here. There are others that are now not in sync below.

native/dlasq2.go
kortschak: "Dlasq2 computes all the eigenvalues of the symmetric positive definite tridiagonal matrix associated with the qd array Z to high relative accuracy~~ are computed to high relative accuracy~~, in the absence of denormalization, underflow and overflow."

The original FORTRAN comment is not English.

native/dlasq2.go
```diff
@@ -0,0 +1,330 @@
...
+// relative accuracy are computed to high relative accuracy, in the
+// absence of denormalization, underflow and overflow.
+//
+// To see the relation of Z to the tridiagonal matrix, let L be a
+// unit lower bidiagonal matrix with subdiagonals Z(2,4,6,,..) and
+// let U be an upper bidiagonal matrix with 1's above and diagonal
+// Z(1,3,5,,..). The tridiagonal is L*U or, if you prefer, the
+// symmetric tridiagonal to which it is similar.
+//
+// info returns a status error. If info is 0, the algorithm completed
```
kortschak: I think a table of return states reads more eaily. The FORTRAN comment is actually a good model here.

kortschak:
I think that the internal commentary in the functions should probably be retained from the FORTRAN. In the cases of the simpler functions you've ported there generally remains a good 1-to-1 correspondence between the Go code and the FORTRAN, but here, when this *is* massaged into more idiomatic Go code there is likely to be a significant change in code layout. It would be good to be able to tie the two back together.

Actually in some of the previously ported functions this might be worth while too (where there has been a restructuring of code). Sorry, I probably should have said something at the time.

native/dlasq2.go
```diff
@@ -0,0 +1,330 @@
...
+			ipn4 := ipn4Out - 1
+			z[i4-3], z[ipn4-i4-4] = z[ipn4-i4-4], z[i4-3]
+			z[i4-1], z[ipn4-i4-6] = z[ipn4-i4-6], z[i4-1]
+		}
+	}
+
+	pp := 0
+	for k := 0; k < 2; k++ {
+		d = z[4*n0+pp]
+		for i4loop := 4*n0 + pp; i4loop >= 4*(i0+1)+pp; i4loop -= 4 {
```
kortschak: Why is this done with two vars?

native/dlasq2.go
```diff
@@ -0,0 +1,330 @@
...
+			if z[i4-1] <= tol2*d {
+				z[i4-1] = math.Copysign(0, -1)
+				d = z[i4-3]
+			} else {
+				d = z[i4-3] * (d / (d + z[i4-1]))
+			}
+		}
+		emin = z[4*(i0+1)+pp]
+		d = z[4*i0+pp]
+		for i4loop := 4*(i0+1) + pp; i4loop <= 4*n0+pp; i4loop += 4 {
```
kortschak: Same question here.

native/dlasq2.go
```diff
@@ -0,0 +1,330 @@
...
+			i4 := i4loop - 1
+			z[i4-2*pp-2] = d + z[i4-1]
+			if z[i4-1] <= tol2*d {
+				z[i4-1] = math.Copysign(0, -1)
```
kortschak: Wut! This is horrible. I know that this is what is done in the FORTRAN, but please add a comment.

native/dlasq2.go
```diff
@@ -0,0 +1,330 @@
...
+				tmp := z[i4+1] / z[i4-2*pp-2]
+				z[i4-2*pp] = z[i4-1] * tmp
+				d *= tmp
+			} else {
+				z[i4-2*pp] = z[i4+1] * (z[i4-1] / z[i4-2*pp-2])
+				d = z[i4+1] * (d / z[i4-2*pp-2])
+			}
+			emin = math.Min(emin, z[i4-2*pp])
+		}
+		z[4*(n0+1)-pp-3] = d
```
kortschak: Can't you drop the +1 and the -3 (changed to +1) here and the next line? Or are you keeping this to maintain similarity with the FORTRAN?

native/dlasq2.go
```diff
@@ -0,0 +1,330 @@
...
+		qmax = z[4*(i0+1)-pp-3]
+		for i4loop := 4*(i0+1) - pp + 2; i4loop <= 4*(n0+1)+pp-2; i4loop += 4 {
```
kortschak: Why two vars?

native/dlasq2.go
btracey: Yea... I don't know. Unfortunately, in this PR I'm a scribe and not a translator. I don't know what effect is has, if any, so I kept as is. I can add a "this is what was done in Fortran", but I'm afraid I can't give a real explanation.

native/dlasq2.go
btracey: Just trying to minimize the transcription errors. I was getting confused with the interactions of zero vs one indexing, > vs >= and positive vs. negative increments. This seemed much less bug prone.

native/dlasq2.go
```diff
@@ -0,0 +1,330 @@
...
+	var nFail int
+	nDiv := 2 * (n0 - i0)
+	var i4 int
+	for iwhila := 1; iwhila <= n+1; iwhila++ {
+		if n0 < 0 {
+			goto OneSeventy
+		}
+		var desig float64
+		var sigma float64
+		if n0 != n-1 {
```
kortschak: Because of things like this, vars like n0 should get brief commentary on what they mean.

native/dlasq2.go
btracey: It was to maintain similarity, and it seemed like this ended up a good decision during debugging. I think the only thing I simplified was z[n*(i+1) - n].

native/dlasq2.go
```diff
@@ -0,0 +1,330 @@
...
+		}
+		var emax float64
+		if n0 > i0 {
+			emin = math.Abs(z[4*(n0+1)-6])
+		} else {
+			emin = 0
+		}
+		qmin := z[4*(n0+1)-4]
+		qmax = qmin
+		var out bool
```
kortschak: Can you find a more informative name for this. Also initialise it to false explicitly. Out of curiosity why not use the goto?

native/dlasq2.go
```diff
@@ -0,0 +1,330 @@
...
+		for i4loop := 4 * (n0 + 1); i4loop >= 8; i4loop -= 4 {
+			i4 = i4loop - 1
```
kortschak: Why two vars?

native/dlasq2.go
```diff
@@ -0,0 +1,330 @@
...
+			i4 = 3
+		}
+		i0 = (i4+1)/4 - 1
+		pp = 0
+		if n0-i0 > 1 {
+			dee := z[4*i0]
+			deemin := dee
+			kmin := i0
+			for i4loop := 4*(i0+1) + 1; i4loop <= 4*(n0+1)-3; i4loop += 4 {
+				i4 := i4loop - 1
```
kortschak: Ditto.

native/dlasq2.go
```diff
@@ -0,0 +1,330 @@
...
+				dee = z[i4] * (dee / (dee + z[i4-2]))
+				if dee <= deemin {
+					deemin = dee
+					kmin = (i4+4)/4 - 1
+				}
+			}
+			if (kmin-i0)*2 < n0-kmin && deemin <= 0.5*z[4*n0] {
+				ipn4Out := 4 * (i0 + n0 + 2)
+				pp = 2
+				for i4loop := 4 * (i0 + 1); i4loop <= 2*(i0+n0+1); i4loop += 4 {
```
kortschak: Ditto.

native/dlasq2.go
```diff
@@ -0,0 +1,330 @@
...
+					i4 := i4loop - 1
+					ipn4 := ipn4Out - 1
```
kortschak: Same but spread over more lines.

native/dlasq2.go
```diff
@@ -0,0 +1,330 @@
...
+					z[i4-3], z[ipn4-i4-4] = z[ipn4-i4-4], z[i4-3]
```
kortschak: This block illustrates why 0-based indexing is superior. The offsets in FORTRAN mean different things depending on the context.

native/dlasq2.go
btracey: The actual code description is "Local Variables: I0:N0 defines a current unreduced segment of Z". The unreduced part is useful, but otherwise it's incomplete. It doesn't really help here either -- why is sigma taking values from z[4*n]? The description of z only talks about the z[2*n] elements. 

I'll add in the comments. I'm just saying I'm with you.

native/dlasq2.go
btracey: I removed this goto because it was clear to me how to eliminate it. Most of the gotos work into spaghetti code that was hard to disentangle.

native/dlasq2.go
```diff
@@ -0,0 +1,330 @@
...
+
+			pp = 1 - pp
+			if pp == 0 && n0-i0 >= 3 {
+				if z[4*(n0+1)-1] <= tol2*qmax || z[4*(n0+1)-2] <= tol2*sigma {
+					splt := i0 - 1
+					qmax = z[4*i0]
+					emin = z[4*(i0+1)-2]
+					oldemn := z[4*(i0+1)-1]
+					for i4loop := 4 * (i0 + 1); i4loop <= 4*(n0-2); i4loop += 4 {
+						i4 := i4loop - 1
```
kortschak: Why two vars?

native/dlasq2.go
```diff
@@ -0,0 +1,330 @@
...
+					z[4*(n0+1)-2] = emin
+					z[4*(n0+1)-1] = oldemn
+					i0 = splt + 1
+				}
+			}
+		}
+		info = 2
+		i1 = i0
+		n1 = n0
+	OneFourtyFive:
```
kortschak: New line above.

native/dlasq2.go
```diff
@@ -0,0 +1,330 @@
...
+					z[i4-1], z[ipn4-i4-6] = z[ipn4-i4-6], z[i4-1]
+					z[i4], z[ipn4-i4-5] = z[ipn4-i4-5], z[i4]
+				}
+			}
+		}
+		dmin := -math.Max(0, qmin-2*math.Sqrt(qmin)*math.Sqrt(emax))
+		nbig := 100 * (n0 - i0 + 1)
+		for iwhilb := 0; iwhilb < nbig; iwhilb++ {
+			if i0 > n0 {
+				goto OneFifty
```
kortschak: This can be written as continue, no?

native/dlasq2.go
kortschak: Yes, it's a nightmare. I am wondering how to deal with that in the long run. Do we want to break these things out into func? I don't think that will be possible because of the degree of state being kept in the sea of vars this has.

native/dlasq2.go
kortschak: Fair enough. Maybe later.

native/dlasq2.go
kortschak: Fair enough. We can clean these up later.

native/dlasq2.go
kortschak: It seems to me that this goto is actually one that I would keep, but we can deal with that later.

native/dlasq2.go
kortschak: I suspect it is related to something like this: http://play.golang.org/p/ylDBNguxEx

However, I can't see anywhere that that has an impact. Out of curiosity, does it fail if removed?

btracey: The tests pass if removed. It probably has something to do with that, but 0 == -0 and 0 !> -0. The practical difference is the behavior on things like 1/negZero vs. 1/0, but I don't know why that would matter here -- singular values shouldn't be infinity.

kortschak: I tried with setting the element to 1 (and +Inf) and it passes too - so I don't know how valuable a pass with just changing the sign of zero is.

The only place where it would make a difference is when you take the max and get the index (which you can't unless you explicitly test for the sign bit, which doesn't happen here - at least in dlasq2, maybe in the subroutines it does).

kortschak: I think the most likely explanation is this 'Informally, one may use the notation "−0" for a negative value that was rounded to zero.' from [wp](https://en.wikipedia.org/wiki/Signed_zero#Scientific_uses); it is purely informational. If you look at the surrounding contexts where this happens, you can see the value was below some threshold.

native/dlasq3.go
```diff
@@ -0,0 +1,153 @@
...
+package native
+
+import "math"
+
+// Dlasq3 checks for deflation, computes a shift (tau) and calls dqds.
+// In case of failure it changes shifts, and tries again until output
+// is positive.
+func (impl Implementation) Dlasq3(i0, n0 int, z []float64, pp int, dmin, sigma, desig, qmax float64, nFail, iter, nDiv int, ttype int, dmin1, dmin2, dn, dn1, dn2, g, tau float64) (
+	i0Out, n0Out, ppOut int, dminOut, sigmaOut, desigOut, qmaxOut float64, nFailOut, iterOut, nDivOut, ttypeOut int, dmin1Out, dmin2Out, dnOut, dn1Out, dn2Out, gOut, tauOut float64) {
+	// Note: The documentation says dmin is an output only, but if that is so it is
```
kortschak: I think this is a copy paste error. There are routines here where it's true.

native/dlasq3.go
```diff
@@ -0,0 +1,153 @@
...
+	// used before it is set.
+	const cbias = 1.5
+
+	n0in := n0
+	eps := dlamchP
+	tol := eps * 100
+	tol2 := tol * tol
+	var nn int
+	var t float64
+Ten:
```
kortschak: `for {`

native/dlasq3.go
```diff
@@ -0,0 +1,153 @@
...
+	if n0 == i0+1 {
+		goto Fourty
+	}
+	if z[nn-5] > tol2*(sigma+z[nn-3]) && z[nn-2*pp-4] > tol2*z[nn-7] {
+		goto Thirty
+	}
+Twenty:
+	z[4*(n0+1)-4] = z[4*(n0+1)+pp-4] + sigma
+	n0--
+	goto Ten
```
kortschak: `continue`

native/dlasq3.go
```diff
@@ -0,0 +1,153 @@
...
+			s = z[nn-3] * (z[nn-5] / (t + math.Sqrt(t)*math.Sqrt(t+s)))
+		}
+		t = z[nn-7] + (s + z[nn-5])
+		z[nn-3] *= z[nn-7] / t
+		z[nn-7] = t
+	}
+	z[4*(n0+1)-8] = z[nn-7] + sigma
+	z[4*(n0+1)-4] = z[nn-3] + sigma
+	n0 -= 2
+	goto Ten
```
kortschak: `}`

native/dlasq3.go
```diff
@@ -0,0 +1,153 @@
...
+	if z[nn-5] > tol2*(sigma+z[nn-3]) && z[nn-2*pp-4] > tol2*z[nn-7] {
+		goto Thirty
+	}
+Twenty:
+	z[4*(n0+1)-4] = z[4*(n0+1)+pp-4] + sigma
+	n0--
+	goto Ten
+Thirty:
+	if z[nn-9] > tol2*sigma && z[nn-2*pp-8] > tol2*z[nn-11] {
+		goto Fifty
```
kortschak: `break`

native/dlasq3.go
```diff
@@ -0,0 +1,153 @@
...
+		}
+		t = z[nn-7] + (s + z[nn-5])
+		z[nn-3] *= z[nn-7] / t
+		z[nn-7] = t
+	}
+	z[4*(n0+1)-8] = z[nn-7] + sigma
+	z[4*(n0+1)-4] = z[nn-3] + sigma
+	n0 -= 2
+	goto Ten
+Fifty:
```
kortschak: Delete.

native/dlasq4.go
```diff
@@ -0,0 +1,228 @@
...
+	)
+
+	if dmin <= 0 {
+		tau = -dmin
+		ttype = -1
+		return tau, ttype, g
+	}
+	nn := 4*(n0+1) + pp - 1 //-1 for zero indexing
```
kortschak: Space before -1 in comment.

native/dlasq4.go
```diff
@@ -0,0 +1,228 @@
...
+	s := math.NaN()         // Make sure one of the paths below is taken
```
kortschak: You are not making sure one is taken, you are marking when one is not.

"Poison s so that failure to take a path below is obvious."

native/dlasq4.go
```diff
@@ -0,0 +1,228 @@
...
+	if n0in == n0 {
+		// No eigenvalues deflated.
+		if dmin == dn || dmin == dn1 {
+			b1 := math.Sqrt(z[nn-3]) * math.Sqrt(z[nn-5])
+			b2 := math.Sqrt(z[nn-7]) * math.Sqrt(z[nn-9])
+			a2 := z[nn-7] + z[nn-5]
+			if dmin == dn && dmin1 == dn1 {
+				gap2 := dmin2 - a2 - dmin2*0.25
```
kortschak: `/4`?

native/dlasq5.go
```diff
@@ -0,0 +1,122 @@
...
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package native
+
+import "math"
+
+// Dlasq5 computes one dqds transform in ping-pong form.
+// i0 and n0 are zero-indexed.
+func (impl Implementation) Dlasq5(i0, n0 int, z []float64, pp int, tau, sigma, dmin, dmin1, dmin2, dn, dnm1, dnm2 float64) (i0Out, n0Out, ppOut int, tauOut, sigmaOut, dminOut, dmin1Out, dmin2Out, dnOut, dnm1Out, dnm2Out float64) {
```
kortschak: Here dmin is actually an out only, so drop it from the input parameter list.

native/dlasq5.go
```diff
@@ -0,0 +1,122 @@
...
+		return
+	}
+	eps := dlamchP
+	dthresh := eps * (sigma + tau)
+	if tau < dthresh*0.5 {
+		tau = 0
+	}
+	var j4 int
+	var emin float64
+	if tau != 0 {
```
kortschak: Wow. The fortran here just decides to not indent. Making it even harder to read.

native/dlasq6.go
```diff
@@ -0,0 +1,107 @@
...
+
+package native
+
+import "math"
+
+// Dlasq6 computes one dqd transform in ping-pong form with protection against
+// overflow and underflow. z has length at least 4*n and holds the qd array.
+// i0 is first index (should be zero-indexed rather than one indexed)
+// n0 is last index (should be zero-indexed)
+func (impl Implementation) Dlasq6(i0, n0 int, z []float64, pp int, dmin, dmin1, dmin2, dn, dnm1, dnm2 float64) (dminOut, dmin1Out, dmin2Out, dnOut, dnm1Out, dnm2Out float64) {
```
kortschak: dmin, dmin1, dmin2, dn, dnm1 and dnm2 are out only.

native/dlasq6.go
```diff
@@ -0,0 +1,107 @@
...
+		return
+	}
+	safmin := dlamchS
+	j4 := 4*(i0+1) + pp - 4 // -4 rather than -3 for zero indexing
+	emin := z[j4+4]
+	d := z[j4]
+	dmin = d
+	if pp == 0 {
+		for j4loop := 4 * (i0 + 1); j4loop <= 4*((n0+1)-3); j4loop += 4 {
+			j4 := j4loop - 1 // Translate back to zero-indexed
```
kortschak: Maybe if you want to keep this approach, write a helper `func zeroBased(i int) int { return i-1 }` and us it instead of the comment.

kortschak:
OK finished first pass.

@vladimir-ch Would you please take a look as well. There is a lot here and even a low review error rate (which I don't guarantee) is troubling.

vladimir-ch:
There is a binary file `internal/testdata/dlasqtest/dlasq1test`

native/dlasq2.go
btracey: Done.

native/dlasq2.go
btracey: Done.

native/dlasq2.go
btracey: Done.

native/dlasq2.go
btracey: Done. I understand what you mean now, sorry.

native/dlasq2.go
btracey: Added here.

native/dlasq2.go
btracey: Makes sense.

native/dlasq2.go
btracey: Done.

native/dlasq2.go
btracey: Renamed.

native/dlasq2.go
btracey: No, this goes back up to iwhila while continue would just go to iwhilb

native/dlasq3.go
btracey: I'm not sure what you mean. It's used after Fifty and never set.

native/dlasq3.go
btracey: Thanks.

native/dlasq3.go
btracey: Apparently unused goto lables are also a compile error. I'm not sad I haven't encountered this before.

native/dlasq4.go
btracey: Done.

native/dlasq4.go
btracey: Thanks.

native/dlasq4.go
btracey: Done.

native/dlasq2.go
kortschak: continue with a label.

native/dlasq5.go
btracey: It's not that simple. You'll notice that the first return happens, and that in the LAPACK function none of the values are initialized. Well... since Fortran passes by reference their value is preserved through the call. It seems like they are actually outputs here (the tests don't fail if I add it), but I ran into these bugs in Dlasq4. I added a TODO.

native/dlasq5.go
btracey: Indeed.

native/dlasq6.go
btracey: Same as above.

native/dlasq6.go
btracey: When we have more full integration tests we should attempt a cleanup of the code or switch it. This probably shouldn't be permanent.

native/dlasq2.go
btracey: Right. Done.

native/dlasq3.go
kortschak: I mean the fortran statement that it's output only.

native/dlasq5.go
```diff
@@ -0,0 +1,125 @@
...
+// license that can be found in the LICENSE file.
+
+package native
+
+import "math"
+
+// Dlasq5 computes one dqds transform in ping-pong form.
+// i0 and n0 are zero-indexed.
+func (impl Implementation) Dlasq5(i0, n0 int, z []float64, pp int, tau, sigma, dmin, dmin1, dmin2, dn, dnm1, dnm2 float64) (i0Out, n0Out, ppOut int, tauOut, sigmaOut, dminOut, dmin1Out, dmin2Out, dnOut, dnm1Out, dnm2Out float64) {
+	// TODO(btracey): It seems like the listed outputs of this function are actually
```
kortschak: "listed inputs"?

native/dlasq5.go
kortschak: That's horrible.

native/dlasq5.go
```diff
@@ -0,0 +1,125 @@
...
+	// outputs (dmin, dn, etc.). Make them true outputs.
+	// The lapack function has inputs for ieee and eps, but Go requires ieee so
+	// these are unnecessary.
+	if n0-i0-1 <= 0 {
+		return
```
kortschak: Does your comment above mean that this should return "i0, n0, pp, tau, sigma, dmin, dmin1, dmin2, dn, dnm1, dnm2"?

native/dlasq6.go
```diff
@@ -0,0 +1,109 @@
...
+package native
+
+import "math"
+
+// Dlasq6 computes one dqd transform in ping-pong form with protection against
+// overflow and underflow. z has length at least 4*n and holds the qd array.
+// i0 is first index (should be zero-indexed rather than one indexed)
+// n0 is last index (should be zero-indexed)
+func (impl Implementation) Dlasq6(i0, n0 int, z []float64, pp int, dmin, dmin1, dmin2, dn, dnm1, dnm2 float64) (dminOut, dmin1Out, dmin2Out, dnOut, dnm1Out, dnm2Out float64) {
+	// TODO(btracey): It seems like the listed outputs of this function are actually
```
kortschak: Same as above.

native/dlasq3.go
btracey: I still see it as such http://www.netlib.org/lapack/explore-html/db/d5b/dlasq3_8f_source.html but I also submitted a bug that was supposedly fixed http://icl.cs.utk.edu/lapack-forum/viewtopic.php?f=13&t=4830 so probably depends on where one looks.

native/dlasq6.go
```diff
@@ -0,0 +1,109 @@
...
+	// outputs (dmin, dn, etc.). Make them true outputs.
+	if len(z) < 4*(n0+1) {
+		panic(badZ)
+	}
+	if n0-i0-1 <= 0 {
+		return
```
kortschak: Same query about returned values as above.

native/dlasq5.go
btracey: Yes it is. Not a fun bug to find. I fixed the first one, found a second bug along those lines, and then just decided not to believe the documentation and make the function behave like the Fortran one. I probably overreacted, but we can fix later.

native/dlasq5.go
btracey: No, I meant outputs, but I improved the documentation.

native/dlasq5.go
btracey: Yes. Thanks.

native/dlasq6.go
btracey: Fixed. Thanks.

kortschak:
I think that with some of the issues you've found here due to pass through, we should not allow unnamed returns in native. There are a few cases elsewhere that do.

native/dlasq2.go
```diff
@@ -0,0 +1,366 @@
...
+			if k < n0 {
+				z[2*(k+1)-1] = z[4*(k+1)-1]
+			} else {
+				z[2*(k+1)] = 0
+			}
+		}
+		return info
+	}
+	info = 3
+	return
```
kortschak: return info

internal/testdata/dlasqtest/testdlasq3.f90
```diff
@@ -0,0 +1,29 @@
...
+logical :: ieee
+
+n = 84
+
+ieee = .True.
+
+print *, "Starting Dlasq3"
+open(unit = 4, file = "gen4tests.txt")
+open(unit = 5, file = "gen5tests.txt")
+call Dlasq3(i0, n0, z, pp, dmin, sigma, desig, qmax, nFail, iter, nDiv, ieee, ttype, dmin1, dmin2, dn, dn1, dn2, g, tau)
```
vladimir-ch: I will be doing something wrong but I get a segfault at `dlasq3.f:246`.

native/dlasq1.go
```diff
@@ -0,0 +1,95 @@
...
+	"math"
+
+	"github.com/gonum/blas/blas64"
+	"github.com/gonum/lapack"
+)
+
+// Dlasq1 computes the singular values of an n×n bidiagonal matrix with diagonal
+// d and off-diagonal e. On exit, d contains the singular values in decreasing
+// order, and e is overwritten. d must have length at least n, and e must have
+// length at least n-1, and the input work must have length 4*n. Dlasq1 will panic
```
vladimir-ch: and ~~the input~~ work must have length _at least_ 4*n.

native/dlasq1.go
```diff
@@ -0,0 +1,95 @@
...
+import (
+	"math"
+
+	"github.com/gonum/blas/blas64"
+	"github.com/gonum/lapack"
+)
+
+// Dlasq1 computes the singular values of an n×n bidiagonal matrix with diagonal
+// d and off-diagonal e. On exit, d contains the singular values in decreasing
+// order, and e is overwritten. d must have length at least n, and e must have
```
vladimir-ch: at least n, ~~and~~ e must have ?

native/dlasq1.go
```diff
@@ -0,0 +1,95 @@
...
+	work[2*n-1] = 0
+
+	info = impl.Dlasq2(n, work)
+	if info == 0 {
+		for i := 0; i < n; i++ {
+			d[i] = math.Sqrt(work[i])
+		}
+		impl.Dlascl(lapack.General, 0, 0, scale, sigmx, n, 1, d, 1)
+	} else if info == 2 {
+		//  Maximum number of iterations exceeded.  Move data from WORK
```
vladimir-ch: Remove extra spaces and add a dot at the end of the sentence.

native/dlasq2.go
```diff
@@ -0,0 +1,366 @@
...
+//  0: The algorithm completed successfully.
+//  1: A split was marked by a positive value in e.
+//  2: Current block of Z not diagonalized after 100*n iterations (in inner
+//     while loop). On exit Z holds a qd array with the same eigenvalues as
+//     the given Z.
+//  3: Termination criterion of outer while loop not met (program created more
+//     than N unreduced blocks).
+func (impl Implementation) Dlasq2(n int, z []float64) (info int) {
+	// TODO(btracey): make info an error.
+	if len(z) < 4*n {
```
vladimir-ch: Add this requirement to the docs.

native/dlasq2.go
```diff
@@ -0,0 +1,366 @@
...
+		z[5] = z[1] + z[0]
+		return info
+	}
+	// Check for negative data and compute sums of q's and e's.
+	z[2*n-1] = 0
+	emin := z[1]
+	var d, e, qmax, zmax float64
+	var i1, n1 int
+	for k := 0; k < 2*(n-1); k += 2 {
+		if z[k] < 0 {
```
vladimir-ch: Make it `if z[k] < 0 || z[k+1] < 0` ? Also, it is not clear (at least not to me) where the requirement that elements of z must be non-negative comes from. At least it should be mentioned in the docs.

native/dlasq6.go
```diff
@@ -0,0 +1,112 @@
...
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package native
+
+import "math"
+
+// Dlasq6 computes one dqd transform in ping-pong form with protection against
+// overflow and underflow. z has length at least 4*n and holds the qd array.
+// i0 is first index (should be zero-indexed rather than one indexed)
```
vladimir-ch: `// i0 is first zero-based index` instead of the parentheses and should ?

native/dlasq6.go
```diff
@@ -0,0 +1,112 @@
...
+// n0 is last index (should be zero-indexed)
```
vladimir-ch: `// n0 is last zero-based index` ?

native/dlasq6.go
```diff
@@ -0,0 +1,112 @@
+// Copyright ©2015 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package native
+
+import "math"
+
+// Dlasq6 computes one dqd transform in ping-pong form with protection against
+// overflow and underflow. z has length at least 4*n and holds the qd array.
```
vladimir-ch: There is no `n` here but yea ...

internal/testdata/dlasqtest/commands.txt
```diff
@@ -0,0 +1,9 @@
+# Advice: nDiv in dlasq2 can be an effective counter for isolating a bug by using os.Exit(1) and "STOP" when nDiv hits a certain number.
+
+gfortran -o dlasq1test dlamch.f lsame.f xerbla.f ieeeck.f iparmq.f ilaenv.f dlasrt.f dlaisnan.f disnan.f dlasq6.f dlasq5.f dlasq4.f dlasq3.f dlasq2.f dcopy.f dlas2.f dlascl.f dlasq1.f testdlasq1.f90
```
vladimir-ch: Change this into a `Makefile` with a `clean` target so that the repository can be easily reverted into a clean state, or is it not worth the hassle?

vladimir-ch:
Finally had some time for the first time in two weeks to take a superficial look at this. I'm not sure if my comments are helpful and if I will be able to look into this deeper.

native/dlasq6.go
kortschak: `n0 is the zero-based last index`

native/dlasq6.go
kortschak: `i0 is the zero-based first index`

internal/testdata/dlasqtest/commands.txt
kortschak: I had similar thoughts, though also not sure of the value/cost ratio.

vladimir-ch: If possible, I would prefer having a `Makefile` instead, and if I may be really annoying, then also a `README.md` with just a short explanation what this is, what it does, what are the textfiles it generates.

This `Makefile` works for me, untested on OSX:
```make
FC = gfortran
FFLAGS = -O2

targets = testdlasq1 testdlasq2 testdlasq3 testdlasq4
objects = dlamch.o lsame.o xerbla.o ieeeck.o iparmq.o ilaenv.o dlasrt.o \
          dlaisnan.o disnan.o dlasq6.o dlasq5.o dlasq4.o dlasq3.o dlasq2.o

default : $(targets)
.PHONY : default

testdlasq1 : testdlasq1.o $(objects) dcopy.o dlas2.o dlascl.o dlasq1.o

testdlasq2 : testdlasq2.o $(objects)

testdlasq3 : testdlasq3.o $(objects)

testdlasq4 : testdlasq4.o $(objects)

% : %.o
    $(FC) $(FFLAGS) $^ -o $@

%.o : %.f90
    $(FC) $(FFLAGS) -c -o $@ $<

clean :
    rm -f *.o $(targets) *.txt
.PHONY : clean
```

kortschak: Thanks for posting this. I agree given that it exists now.

btracey: Done. Thanks.

internal/testdata/dlasqtest/testdlasq3.f90
btracey: You're right.... how did that happen?

btracey: Oh, it's because there isn't any data in testdlasq3 at the moment.

native/dlasq1.go
btracey: Done.

native/dlasq1.go
btracey: Done.

native/dlasq1.go
btracey: Done.

native/dlasq2.go
btracey: Done.

native/dlasq2.go
btracey: Not clear to me either, but it's in the lapack code...

Done.

native/dlasq6.go
btracey: Done.

native/dlasq6.go
btracey: Done.

native/dlasq6.go
btracey: Done.

btracey:
Thanks for the commentary. 

Onward to routines we can write real tests for!

PR #69, opened 2015-11-09
internal/fortran: new package - conversion utilities

kortschak:
This is intended to be used in conjunction with the gc cgo fortran patch to aid LAPACK porting, but could in the longer term be made public elsewhere after being made more friendly.

@btracey Please take a look.

### DIFF
--- a/internal/fortran/fortran.go
+++ b/internal/fortran/fortran.go
@@ -1,0 +1,476 @@
+// Copyright ©2015 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package fortran provides utilities to help interface to fortran routines.
+//
+// The utility functions provided in fortran are not intended to be used generally
+// and provide very limited friendliness. They are intended to be used in development
+// of Go ports of FORTRAN routines.
+package fortran
+
+import (
+	"github.com/gonum/blas"
+	"github.com/gonum/blas/blas64"
+)
+
+// TransCharacter returns the FORTRAN character for the given transpose operation.
+func TransCharacter(trans blas.Transpose) byte {
+	switch trans {
+	case blas.NoTrans:
+		return 'N'
+	case blas.Trans:
+		return 'T'
+	case blas.ConjTrans:
+		return 'C'
+	default:
+		panic("fortran: bad BLAS trans")
+	}
+}
+
+// UploCharacter returns the FORTRAN character for the given uplo state.
+func UploCharacter(uplo blas.Uplo) byte {
+	switch uplo {
+	case blas.Upper:
+		return 'U'
+	case blas.Lower:
+		return 'L'
+	case blas.All:
+		return 'A'
+	default:
+		panic("fortran: bad BLAS uplo")
+	}
+}
+
+// DiagCharacter returns the FORTRAN character for the given diagonal state.
+func DiagCharacter(diag blas.Diag) byte {
+	switch diag {
+	case blas.Unit:
+		return 'U'
+	case blas.NonUnit:
+		return 'N'
+	default:
+		panic("fortran: bad BLAS diag")
+	}
+}
+
+// SideCharacter returns the FORTRAN character for the given side state.
+func SideCharacter(side blas.Side) byte {
+	switch side {
+	case blas.Left:
+		return 'L'
+	case blas.Right:
+		return 'R'
+	default:
+		panic("fortran: bad BLAS side")
+	}
+}
+
+// General is a column major general matrix.
+type General blas64.General
+
+// NewColMajorGeneralFrom returns a column major general matrix
+// with the same dimensions and data elements as the row major a.
+func NewColMajorGeneralFrom(a blas64.General) General {
+	t := General{
+		Rows:   a.Rows,
+		Cols:   a.Cols,
+		Stride: a.Rows,
+		Data:   make([]float64, a.Rows*a.Cols),
+	}
+	t.From(a)
+	return t
+}
+
+// From fills the receiver with elements from a. The receiver
+// must have the same dimensions as a and have adequate backing
+// data storage.
+func (t General) From(a blas64.General) {
+	if t.Rows != a.Rows || t.Cols != a.Cols {
+		panic("fortran: mismatched dimension")
+	}
+	if len(t.Data) < (t.Cols-1)*t.Stride+t.Rows {
+		panic("fortran: short data slice")
+	}
+	for i := 0; i < a.Rows; i++ {
+		for j := 0; j < a.Cols; j++ {
+			t.Data[i+j*t.Stride] = a.Data[i*a.Stride+j]
+		}
+	}
+}
+
+// NewRowMajorGeneralFrom returns a row major general matrix
+// with the same dimensions and data elements as a column major a.
+func NewRowMajorGeneralFrom(a General) blas64.General {
+	t := blas64.General{
+		Rows:   a.Rows,
+		Cols:   a.Cols,
+		Stride: a.Cols,
+		Data:   make([]float64, a.Rows*a.Cols),
+	}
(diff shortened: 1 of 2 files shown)


kortschak:
There is a difficulty with the package directory layout - internal/fortran needs to be at the root so it can be used by tests in the packages being developed, but the approach used in the .travis.yml is to pushd to native to avoid trying to build/test cgo when not using OpenBLAS - this means that fortran is only tested for the cgo half of the test matrix. I don't want fortran to be public at this stage, so it needs to be internal. The blunt way to deal with this is explicitly test ./internal/..., but that's not elegant.

btracey:
Sorry, I don't understand what you mean. "the approach used in the .travis.yml is to pushd to native to avoid trying to build/test cgo when not using OpenBLAS"

jonlawlor:
The relevant line in the .travis.yml file:
```
if [[ "$BLAS_LIB" == "gonum" ]]; then pushd native; fi
```
which means that when testing the native code paths it is in the gonum subdirectory, so that the following lines:
```
 - go get -d -t -v ./...
 - go build -v ./...
 - go test -v ./...
```
Only apply to code in that directory.

An alternative might be using testing's Skip().

kortschak:
Skip won't work, because the issue is that clapack and its dependencies can't be built without a C library to link to.

btracey:
@kortschak is your thought that the cases in testlapack would call fortran directly? Similar to the approach that I have at the moment, but instead of listing the fortran results they would be computed on the fly?

It's better for documentation purposes to see where they're coming from, but otherwise it's not that different. One can write a few fmt.Printf statements to generate the test structures, and the answers can be confirmed by calling "go test" in internal/fortran. There seems to have been a lot of utility in direct tests of function behavior instead of numerical output, and so we should still strive for that where possible. Seeing a long list of test structures like is there is a pretty obvious signal that the tests are wonky.

internal/fortran/fortran.go
```diff
@@ -0,0 +1,422 @@
...
+// and provide very limited friendliness. They are intended to be used in development
+// of Go ports of FORTRAN routines.
+package fortran
+
+import (
+	"github.com/gonum/blas"
+	"github.com/gonum/blas/blas64"
+)
+
+// TransCharacter returns the FORTRAN character for the given transpose operation.
```
btracey: Is there a reasons these couldn't be simply fortran.Trans rather than the full TransCharacter?

internal/fortran/fortran.go
```diff
@@ -0,0 +1,422 @@
...
+		panic("fortran: bad BLAS side")
+	}
+}
+
+// General is a column major general matrix.
+type General blas64.General
+
+// FromRowMajorGeneral returns a column major general matrix
+// with the same dimensions and data elements as the row major a.
+func FromRowMajorGeneral(a blas64.General) General {
```
btracey: Could we keep the stride of the input matrix (and fill in with 0s or garbage or whatever)? There are bugs that occur that are easily found with a mismatched stride. Sometimes indexing, but more frequently the trickiness that happens with temporary work arrays.

internal/fortran/fortran.go
```diff
@@ -0,0 +1,422 @@
...
+		panic("fortran: short data slice")
+	}
+	for i := 0; i < src.Rows; i++ {
+		for j := 0; j < src.Cols; j++ {
+			dst.Data[i*dst.Stride+j] = src.Data[i*src.Stride+j]
+		}
+	}
+}
+
+// CopyGeneralColMajor performs a copy of src to dst which are
```
btracey: Maybe make this Clone just to avoid getting our wires crossed with mat64?

internal/fortran/fortran.go
```diff
@@ -0,0 +1,422 @@
...
+		Uplo:   a.Uplo,
+	}
+	switch a.Uplo {
+	case blas.Lower:
+		for i := 0; i < a.N; i++ {
+			for j := 0; j <= i; j++ {
+				t.Data[i+j*t.Stride] = a.Data[i*a.Stride+j]
+			}
+		}
+	case blas.Upper:
```
btracey: Nitpicking, but in basically all of the blas/lapack code the Upper case is first.

btracey:
Aside from the minor suggestions, code LGTM. I don't fully understand the intent yet of the location/testing issue to be able to properly comment. Sorry for being dense.

internal/fortran/fortran.go
kortschak: I want the code to be completely explicit about what is going on. The code here (more below in comment to answer your broader questions) is intended to be scaffolding for porting FORTRAN routines, not something for general use, thought that may change.

internal/fortran/fortran.go
kortschak: I had some ideas about changing the API that will have this as a natural consequence. I'll push them in a second.

internal/fortran/fortran.go
kortschak: Again, API changes will address this.

internal/fortran/fortran.go
kortschak: Will do.

kortschak:
>@kortschak is your thought that the cases in testlapack would call fortran directly?

Sort of. The intention is that during porting you could do that if you wanted to. Generally, at the moment you could call the LAPACKE routines with `LAPACK_COL_MAJOR` and essentially bypass the C code to simplify the bug domain, but with Seb's patch you could call the FORTRAN routine directly - adding debug prints etc to the FORTRAN and avoiding to monotony of rebuilding OpenBLAS each time.

>Similar to the approach that I have at the moment, but instead of listing the fortran results they would be computed on the fly?

Yes. If means you can leverage the general randomised testing you already have.

>It's better for documentation purposes to see where they're coming from, but otherwise it's not that different. One can write a few fmt.Printf statements to generate the test structures, and the answers can be confirmed by calling "go test" in internal/fortran. There seems to have been a lot of utility in direct tests of function behavior instead of numerical output, and so we should still strive for that where possible. Seeing a long list of test structures like is there is a pretty obvious signal that the tests are wonky.

Agreed.

>I don't fully understand the intent yet of the location/testing issue to be able to properly comment.

@jonlawlor's comment answers this, but more explicitly when `$BLAS_LIB` is "gonum", the directory is changed to ./native and all the build and test invocations are called there. This is so that cgo/... packages are not built or tested (because that would fail). The consequence of this is that since ./internal is outside ./native not packages in that tree are built or tested.

btracey:
> @jonlawlor's comment answers this, but more explicitly when $BLAS_LIB is "gonum", the directory is changed to ./native and all the build and test invocations are called there. This is so that cgo/... packages are not built or tested (because that would fail). The consequence of this is that since ./internal is outside ./native not packages in that tree are built or tested.

Sorry, I understood that, but I did not understand why that is a problem. If the plan was to call Fortran during the native tests, then that makes sense.

kortschak:
The problem is that internal/native is only tested when testing cgo. That would not normally be a problem except that that test path almost always fails anyway.

kortschak:
API changes. PTAL

kortschak:
I think I've come to the opinion that this should be in blas somewhere. I'm torn between a separate package (cannot be internal because of the use case here) or in blas64 with From methods added to all covered types (instead of the From and To methods here) - packed types would also need to be implemented I guess in that case.

kortschak:
@btracey Following this up: Do you agree this should move to blas in some form?

btracey:
Yes, in some form, I agree. I'm not sure exactly what form that is at the moment.

kortschak:
Since Fortran cgo support is now available at tip this has more significant utility. I'm leaning towards adding to blas, but with the conversion methods on all types - this breaks our original design, but I think it is worth it. The thing that is troubling me is the naming of the two sets of types.

vladimir-ch:

Do we still want (a subset of) this in the new repo, @kortschak ?

PR #71, opened 2015-11-17
native: clean up return use

### LINKED ISSUE #70: native: clean up return use in a number of routines
There are a number of routines in native that have naked returns or returns that are unnecessary because of other flow control:

Dgetrf, Dlacpy, Dlaswp, Dlatrs, Dorml2, Dormlq, Dormqr and Dtritri.

kortschak:
Fixes #70.

@btracey Please take a look.

### DIFF
--- a/native/dgetrf.go
+++ b/native/dgetrf.go
@@ -29,7 +29,7 @@ func (impl Implementation) Dgetrf(m, n int, a []float64, lda int, ipiv []int) (o
 		panic(badIpiv)
 	}
 	if m == 0 || n == 0 {
-		return
+		return false
 	}
 	bi := blas64.Implementation()
 	nb := impl.Ilaenv(1, "DGETRF", " ", m, n, -1, -1)
--- a/native/dtrtri.go
+++ b/native/dtrtri.go
@@ -20,7 +20,7 @@ func (impl Implementation) Dtrtri(uplo blas.Uplo, diag blas.Diag, n int, a []flo
 		panic(badDiag)
 	}
 	if n == 0 {
-		return
+		return false
 	}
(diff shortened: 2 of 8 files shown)


kortschak:
*ping*

native/dgetrf.go
```diff
@@ -29,7 +29,7 @@ func (impl Implementation) Dgetrf(m, n int, a []float64, lda int, ipiv []int) (o
 		panic(badIpiv)
 	}
 	if m == 0 || n == 0 {
-		return
+		return false
```
btracey: Should this return false? Lapack returns the equivalent of true. I hate these "empty set" issues.

native/dtrtri.go
```diff
@@ -20,7 +20,7 @@ func (impl Implementation) Dtrtri(uplo blas.Uplo, diag blas.Diag, n int, a []flo
 		panic(badDiag)
 	}
 	if n == 0 {
-		return
+		return false
```
btracey: Same question here.

btracey:
LGTM minus the "return false" issues. Probably doesn't really matter.

native/dgetrf.go
kortschak: I'm just doing what it was doing before. If you think it should return true, I'm fine with that, but it's not what it was doing prior to this change.

native/dtrtri.go
kortschak: Same answer here.

native/dgetrf.go
btracey: Good point. I hadn't thought about that. Hasn't caused any issues, so LGTM. If it breaks something we can reassess.

PR #74, opened 2015-12-08
Add Dormbr and test

btracey:


kortschak:
Test for cgo? clapack has Dormbr.

btracey:
We're back at that level? Exiting!

I just ran the cgo tests to make sure they pass before implementing the Dormbr tests. I hit the Dlantr panic. Not reproducable, but copying here for future reference

````
brendan:~/Documents/mygo/src/github.com/gonum/lapack/cgo$ go test
fatal error: unexpected signal during runtime execution
[signal 0xa code=0x2 addr=0xc820500000 pc=0x5447870]

runtime stack:
runtime.throw(0x425fda0, 0x2a)
	/Users/brendan/gover/go/src/runtime/panic.go:527 +0x90
runtime.sigpanic()
	/Users/brendan/gover/go/src/runtime/sigpanic_unix.go:12 +0x5a

goroutine 7 [syscall, locked to thread]:
runtime.cgocall(0x417ccd0, 0xc82003fa78, 0xc800000000)
	/Users/brendan/gover/go/src/runtime/cgocall.go:120 +0x11b fp=0xc82003fa30 sp=0xc82003fa00
github.com/gonum/lapack/cgo/clapack._Cfunc_LAPACKE_dlantr(0x4e554900000065, 0xa00000005, 0xc8204ffdc0, 0xb, 0x0)
	github.com/gonum/lapack/cgo/clapack/_obj/_cgo_gotypes.go:6622 +0x3a fp=0xc82003fa78 sp=0xc82003fa30
github.com/gonum/lapack/cgo/clapack.Dlantr(0x49, 0x55, 0x4e, 0x5, 0xa, 0xc8204ffdc0, 0x37, 0x37, 0xb, 0x80)
	/Users/brendan/Documents/mygo/src/github.com/gonum/lapack/cgo/clapack/clapack.go:3120 +0xa1 fp=0xc82003fab8 sp=0xc82003fa78
github.com/gonum/lapack/cgo.Implementation.Dlantr(0x585d449, 0x79, 0x83, 0x5, 0xa, 0xc8204ffdc0, 0x37, 0x37, 0xb, 0xc8204f8e80, ...)
	/Users/brendan/Documents/mygo/src/github.com/gonum/lapack/cgo/lapack.go:152 +0x21d fp=0xc82003fb20 sp=0xc82003fab8
github.com/gonum/lapack/cgo.(*Implementation).Dlantr(0x4327df0, 0x49, 0x79, 0x83, 0x5, 0xa, 0xc8204ffdc0, 0x37, 0x37, 0xb, ...)
	<autogenerated>:4 +0x12c fp=0xc82003fb90 sp=0xc82003fb20
github.com/gonum/lapack/testlapack.DlantrTest(0xc820094120, 0x585d6e0, 0x4327df0)
	/Users/brendan/Documents/mygo/src/github.com/gonum/lapack/testlapack/dlantr.go:73 +0x607 fp=0xc82003ff28 sp=0xc82003fb90
github.com/gonum/lapack/cgo.TestDlantr(0xc820094120)
	/Users/brendan/Documents/mygo/src/github.com/gonum/lapack/cgo/lapack_test.go:25 +0x76 fp=0xc82003ff68 sp=0xc82003ff28
testing.tRunner(0xc820094120, 0x4326850)
	/Users/brendan/gover/go/src/testing/testing.go:456 +0x98 fp=0xc82003ffa0 sp=0xc82003ff68
runtime.goexit()
	/Users/brendan/gover/go/src/runtime/asm_amd64.s:1721 +0x1 fp=0xc82003ffa8 sp=0xc82003ffa0
created by testing.RunTests
	/Users/brendan/gover/go/src/testing/testing.go:561 +0x86d

goroutine 1 [chan receive]:
testing.RunTests(0x4278158, 0x4326820, 0x16, 0x16, 0x1)
	/Users/brendan/gover/go/src/testing/testing.go:562 +0x8ad
testing.(*M).Run(0xc82003df08, 0x4006978)
	/Users/brendan/gover/go/src/testing/testing.go:494 +0x70
main.main()
	github.com/gonum/lapack/cgo/_test/_testmain.go:96 +0x116

goroutine 17 [syscall, locked to thread]:
runtime.goexit()
	/Users/brendan/gover/go/src/runtime/asm_amd64.s:1721 +0x1
exit status 2
FAIL	github.com/gonum/lapack/cgo	0.044s
````

File: native/dormbr.go [hunk 1]
```diff
@@ -0,0 +1,121 @@
...
+	if vect == lapack.ApplyQ {
+		checkMatrix(nq, min(nq, k), a, lda)
+	} else {
+		checkMatrix(min(nq, k), nq, a, lda)
+	}
+
+	applyQ := vect == lapack.ApplyQ
+	left := side == blas.Left
+	var nb int
+	opts := string(side) + string(trans)
```
kortschak: I don't understand this for two reasons. First, blas.Side and blas.Transpose are not letters (we translate them in the wrappers around clapack in cgo for example), and second, opts is ignored by Ilaenv in native (in the FORTRAN it is only considered when ispec > 11 which the Go code panics for).

btracey: You're right that this is definitely wrong. What would you like me to do? I'm under the impression that Ilaenv is hypothetically tunable to the architecture, like ATLAS does, and I imagine this is the reason the values are passed explicitly instead of ignored even though the netlib implementation doesn't use them. Would you like me to construct and pass the string explicitly, or pass in empty?

kortschak: We face the user with blas types rather than FORTRAN chars, so I'm less fussed by the translation thing (there is no internal native package like clapack - though the rest of Ilaenv does use letters), but it seems odd to me that ILAENV is called with ISPEC=1 and the OPTS are passed through with work to construct them.

The approach that you are taking is probably the only reasonable way to do it (alternative is as []int) since the types we use for these options are distinct. Maybe just add a comment that the current implementation does not use opts, but a future change may make that happen. Whether they get converted to 'L'/'R' and 'T'/'N' is up to you - and changeable later since nothing is done with them.

testlapack/dormbr.go
```diff
@@ -0,0 +1,137 @@
...
+					ldc := test.ldc
+					if ldc == 0 {
+						ldc = n
+					}
+					nq := n
+					if side == blas.Left {
+						nq = m
+					}
+
+					// Compute a decomposition
```
kortschak: Period.

testlapack/dormbr.go
```diff
@@ -0,0 +1,137 @@
...
+					d := make([]float64, nTau)
+					e := make([]float64, nTau)
+					lwork := -1
+					work := make([]float64, 1)
+					impl.Dgebrd(ma, na, a, lda, d, e, tauQ, tauP, work, lwork)
+					work = make([]float64, int(work[0]))
+					lwork = len(work)
+					impl.Dgebrd(ma, na, a, lda, d, e, tauQ, tauP, work, lwork)
+
+					// Apply and compare update
```
kortschak: Period.

testlapack/dormbr.go
```diff
@@ -0,0 +1,137 @@
...
+					}
+
+					if side == blas.Left {
+						bi.Dgemm(trans, blas.NoTrans, m, n, m, 1, mulMat.Data, mulMat.Stride, cOrig.Data, cOrig.Stride, 0, cAns.Data, cAns.Stride)
+					} else {
+						bi.Dgemm(blas.NoTrans, trans, m, n, n, 1, cOrig.Data, cOrig.Stride, mulMat.Data, mulMat.Stride, 0, cAns.Data, cAns.Stride)
+					}
+
+					if !floats.EqualApprox(cAns.Data, c, 1e-10) {
+						t.Errorf("C mismatch")
```
kortschak: More informative error?

btracey:
The cgo tests fail on Dormbr. This is weird, because before seeing this I would have put a high probability on the go code being correct. We know P and Q are computed appropriately. This is tested by Dgebd2, which both native and cgo pass. The only step after this is to multiply C by the respective matrix. We know Dgemm is correct (otherwise everything would break). It also seems unlikely I would have a bug in the native implementation (a translation from Fortran), and the test (independently coded and using different matrix data) that exactly cancel out to create a passing native test.

I pushed the cgo tests because I'm not sure I have time to debug at the moment and maybe you'll see something obvious.

testlapack/dormbr.go
btracey: Done.

testlapack/dormbr.go
btracey: Done.

testlapack/general.go
```diff
@@ -321,140 +321,8 @@ func constructQK(kind string, m, n, k int, a []float64, lda int, tau []float64)
 func checkBidiagonal(t *testing.T, m, n, nb int, a []float64, lda int, d, e, tauP, tauQ, aCopy []float64) {
 	// Check the answer.
 	// Construct V.and U
```
kortschak: Construct V and U.

testlapack/general.go
```diff
@@ -321,140 +321,8 @@ func constructQK(kind string, m, n, k int, a []float64, lda int, tau []float64)
...
-			Data: u.Data[i*u.Stride:],
-		}
-		blas64.Ger(-tauP[i], ui, ui, gMat)
-
-		// Q = Q * G[1]
-		blas64.Gemm(blas.NoTrans, blas.NoTrans, 1, qCopy, hMat, 0, qMat)
-		// P = P * G[i]
-		blas64.Gemm(blas.NoTrans, blas.NoTrans, 1, pCopy, gMat, 0, pMat)
-	}
+	qMat := constructQPBidiagonal(lapack.ApplyQ, m, n, nb, a, lda, tauQ)
```
kortschak: Nice.

testlapack/dormbr.go
btracey: Done.

btracey:
Well, it's just a small subset that is failing (applyP for specific sizes), so it's probably a bug in the implementation and test.

testlapack/dormbr.go
kortschak: Sorry, add the test details - vect, side, trans, shape params etc.

kortschak:
In a few days I can have a look using a direct FORTRAN call if you don't get to it first.

btracey:
I wonder. The section that is failing has a call to Dormlq, and we have a comment in the cgo test for Dormlq
````
// Test disabled because of bug in c interface. Leaving stub for easy reproducer.
````
I'll try again with the latest OpenBLAS version tonight

File: cgo/lapack.go [hunk 7]
```diff
@@ -170,6 +171,74 @@ func (impl Implementation) Dpotrf(ul blas.Uplo, n int, a []float64, lda int) (ok
 	return clapack.Dpotrf(ul, n, a, lda)
 }
 
+// Dgebrd reduces a general m×n matrix A to upper or lower bidiagonal form B by
+// an orthogonal transformation:
+//  Q^T * A * P = B.
+// The diagonal elements of B are stored in d and the off diagonal elements stored
```
vladimir-ch: off-diagonal

_are_ stored? Or delete?

btracey:
Symptom: Cgo only fails Dorml2 the test when Side == Right, and only under certain conditions. 

I think I see the problem. From the documentation of Dorml2

````
// If side == blas.Left, a is a matrix of side k×m, and if side == blas.Right
// a is of size k×n.
````
From dormql_work.c

````
r = LAPACKE_lsame( side, 'l' ) ? m : n;
// later
LAPACKE_dge_trans( matrix_order, r, k, a, lda, a_t, lda_t );
````

This seems wrong. The size should depend on the value of size. Agreed? This matches that it only fails when side is right (since it's the correct size when left). Not fully clear why it passes under some of the Right conditions, but that may be due to an oddity of the way our test is constructed.

**Comment by btracey** [comment]
I see, I'm probably back to rediscovering https://github.com/xianyi/OpenBLAS/issues/615

I guess I'll go back to seeing if I can fix linking errors.

**Comment by kortschak** [comment]
I think you are right.

**Comment by btracey** [comment]
Nope, I get linking errors on tip as well.

**Comment by kortschak** [comment]
Working for me.

**Comment by btracey** [comment]
It compiles for me if I delete the following functions 

````
cgeqpf
cggsvd
cggsvp
dgeqpf
dggsvd
dggsvp
sgeqpf
sggsvd
sggsvp
zgeqpf
zggsvd
zggsvp
````

**Comment by btracey** [comment]
The Dormlq test is still failing for me, even with the updated OpenBLAS. The transpode code has changed, but it now says (line 86 of dormlq_work.c)

````
LAPACKE_dge_trans( matrix_layout, k, m, a, lda, a_t, lda_t );
`````
I think this is still wrong. The size of A depends on the side, as quoted above.

**Comment by kortschak** [comment]
Yes, I agree to both of those. BTW what OpenBLAS sha are you on?

**Comment by btracey** [comment]
brendan:~/software/OpenBLAS$ git rev-parse HEAD
25116788ef73cd8c5a0f7395818e83def724cfbf

**Comment by kortschak** [comment]
I am at a loss. That's where I am and it builds fine here.

**Comment by btracey** [comment]
Maybe it's a clang/ gcc thing?

**Comment by btracey** [comment]
Size hypothesis correct. If I change the code to

````
  if (LAPACKE_lsame( side, 'l' )){
            a_t = (double*)LAPACKE_malloc( sizeof(double) * lda_t * MAX(1,m) );
        }else{
            a_t = (double*)LAPACKE_malloc( sizeof(double) * lda_t * MAX(1,n) );
        }
        if( a_t == NULL ) {
            info = LAPACK_TRANSPOSE_MEMORY_ERROR;
            goto exit_level_0;
        }
        c_t = (double*)LAPACKE_malloc( sizeof(double) * ldc_t * MAX(1,n) );
        if( c_t == NULL ) {
            info = LAPACK_TRANSPOSE_MEMORY_ERROR;
            goto exit_level_1;
        }
        /* Transpose input matrices */
        if (LAPACKE_lsame( side, 'l' )){
            LAPACKE_dge_trans( matrix_layout, k, m, a, lda, a_t, lda_t );
        }else{
            LAPACKE_dge_trans( matrix_layout, k, n, a, lda, a_t, lda_t );
        }
````
The Dorml2 tests pass

**Comment by kortschak** [comment]
Nice work.

**Comment by btracey** [comment]
http://icl.cs.utk.edu/lapack-forum/viewtopic.php?f=13&t=4862

**Comment by kortschak** [comment]
Send a PR to OpenBLAS?

**Comment by btracey** [comment]
It looks to me like a similar problem with the failing cgo tests for the new function, Dormbr. It has 

````
nq = LAPACKE_lsame( side, 'l' ) ? m : n;
r = LAPACKE_lsame( vect, 'q' ) ? nq : MIN(nq,k);
...
LAPACKE_dge_trans( matrix_layout, r, MIN(nq,k), a, lda, a_t, lda_t );
````
The documentation says that a can be any of mxk, nxk, kxm, or kxn so it doesn't seem like there is any way this code can be correct.

**Comment by btracey** [comment]
For dormlq 
https://github.com/xianyi/OpenBLAS/pull/711

**Comment by btracey** [comment]
And yes, confirmed for Dormbr. Following code gives passing go tests.

````
        if( LAPACKE_lsame( vect, 'q' ) ){
          a_t = (double*)LAPACKE_malloc( sizeof(double) * lda_t * k );
        }else{
          a_t = (double*)LAPACKE_malloc( sizeof(double) * lda_t * nq );
        }
      
        if( a_t == NULL ) {
            info = LAPACK_TRANSPOSE_MEMORY_ERROR;
            goto exit_level_0;
        }
        c_t = (double*)LAPACKE_malloc( sizeof(double) * ldc_t * MAX(1,n) );
        if( c_t == NULL ) {
            info = LAPACK_TRANSPOSE_MEMORY_ERROR;
            goto exit_level_1;
        }
        /* Transpose input matrices */
        if( LAPACKE_lsame( vect, 'q' ) ){
          LAPACKE_dge_trans( matrix_layout, nq, k, a, lda, a_t, lda_t );
        }else{
          LAPACKE_dge_trans( matrix_layout, k, nq, a, lda, a_t, lda_t );
        }
````

**Comment by btracey** [comment]
For dormbr

https://github.com/xianyi/OpenBLAS/pull/713

**Comment by btracey** [comment]
And lapack bug

http://icl.cs.utk.edu/lapack-forum/viewtopic.php?f=13&t=4863

File: testlapack/dormbr.go (hunk 4 above)
Comment (btracey) [inline testlapack/dormbr.go]: Done.

File: testlapack/general.go (hunk 5 above)
Comment (btracey) [inline testlapack/general.go]: Done.

File: cgo/lapack.go (hunk 7 above)

File: native/dormbr.go (hunk 1 above)
Comment (btracey) [inline native/dormbr.go]: Done.

**Comment by btracey** [comment]
Responded to PR comments. 

I disabled the currently failing Lapack tests with good error messages. We can reenable them when/if OpenBLAS updates. Note that this is not a cascading failure. The problem is with the c wrapper, and so functions that call Dormbr / Dormlq  (i.e. SVD) through fortran get the correct answer, it's just calling these functions directly (as the test does) where the wrong answer is achieved.

**Comment by btracey** [comment]
PTAL

File: native/dormbr.go [hunk 8]
```diff
@@ -0,0 +1,134 @@
...
+
+	applyQ := vect == lapack.ApplyQ
+	left := side == blas.Left
+	var nb int
+
+	// The current implementation does not use opts, but a future change may
+	// use these options so construct them.
+	var opts string
+	if side == blas.Left {
+		opts += "L"
```
kortschak: `opts = "L"`, equivalent for "R". The compiler should optimise them to the same operation, but might not.

native/dormbr.go
```diff
@@ -0,0 +1,134 @@
...
+			nb = impl.Ilaenv(1, "DORMLQ", opts, m, n-1, n-1, -1)
+		}
+	}
+	lworkopt := max(1, nw) * nb
+	if lwork == -1 {
+		work[0] = float64(lworkopt)
+	}
+	if applyQ {
+		if nq >= k {
+			// Q was determined by a call to Dgebrd with nq >= k.
```
kortschak: I know these comments reflect the original in the FORTRAN, but they are almost redundant. I'd say they would be improved by using a single comment above the `if nq >= k` condition saying why this is important rather than reflecting the conditional.

https://www.lysator.liu.se/c/pikestyle.html

native/dormbr.go
```diff
@@ -0,0 +1,134 @@
...
+			impl.Dormqr(side, trans, mi, ni, nq-1, a[1*lda:], lda, tau, c[i1*ldc+i2:], ldc, work, lwork)
+		}
+		return
+	}
+	transt := blas.Trans
+	if trans == blas.Trans {
+		transt = blas.NoTrans
+	}
+	if nq > k {
+		// P was determined by a call to Dgebrd with nq > k.
```
kortschak: Same here.

native/dormbr.go
btracey: Done.

native/dormbr.go
btracey: With these comment translations, I'm not always sure what's useful. Lapack is pretty consistent, so a lot of stuff is "obvious" once you've implemented 40 routines. Thanks for the link though, I'll take a look.

btracey: Done.

native/dormbr.go
btracey: Done.

kortschak:
LGTM

PR #75, opened 2015-12-09
cgo/clapack: update codegen for LAPACKE 3.6

kortschak:
@btracey Please take a look.

### DIFF
--- a/cgo/clapack/clapack.go
+++ b/cgo/clapack/clapack.go
@@ -110,6 +110,32 @@ func Zbdsqr(ul blas.Uplo, n int, ncvt int, nru int, ncc int, d []float64, e []fl
 	return isZero(C.LAPACKE_zbdsqr((C.int)(rowMajor), (C.char)(ul), (C.lapack_int)(n), (C.lapack_int)(ncvt), (C.lapack_int)(nru), (C.lapack_int)(ncc), (*C.double)(&d[0]), (*C.double)(&e[0]), (*C.lapack_complex_double)(&vt[0]), (C.lapack_int)(ldvt), (*C.lapack_complex_double)(&u[0]), (C.lapack_int)(ldu), (*C.lapack_complex_double)(&c[0]), (C.lapack_int)(ldc)))
 }
 
+// See http://www.netlib.org/cgi-bin/netlibfiles.txt?format=txt&filename=/lapack/lapack_routine/sbdsvdx.f.
+func Sbdsvdx(ul blas.Uplo, jobz lapack.Job, rng byte, n int, d []float32, e []float32, vl int, vu int, il int, iu int, ns int, s []float32, z []float32, ldz int, superb []int32) bool {
+	switch ul {
+	case blas.Upper:
+		ul = 'U'
+	case blas.Lower:
+		ul = 'L'
+	default:
+		panic("lapack: illegal triangle")
+	}
+	return isZero(C.LAPACKE_sbdsvdx((C.int)(rowMajor), (C.char)(ul), (C.char)(jobz), (C.char)(rng), (C.lapack_int)(n), (*C.float)(&d[0]), (*C.float)(&e[0]), (C.lapack_int)(vl), (C.lapack_int)(vu), (C.lapack_int)(il), (C.lapack_int)(iu), (C.lapack_int)(ns), (*C.float)(&s[0]), (*C.float)(&z[0]), (C.lapack_int)(ldz), (*C.lapack_int)(&superb[0])))
+}
+
+// See http://www.netlib.org/cgi-bin/netlibfiles.txt?format=txt&filename=/lapack/lapack_routine/dbdsvdx.f.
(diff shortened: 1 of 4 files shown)


kortschak:
This can be left until we want to upgrade to 3.6 since it now seems that we can link against 3.6 with ours (on my machine) and an upgrade will break installations that do not have 3.6.

kortschak:
BTW Do not trust the github diff. It is clearly wrong in places.

btracey:
Does this remove the functions that are no longer in Lapack? It looked like it did more from the real diff.

kortschak:
It actually doesn't remove anything. It changes a parameter name and adds some new functions and changes some signatures.

btracey:
Ah, that makes sense. Could it also remove the extra functions (so I can use the new OpenBLAS)?

kortschak:
Can you try building with this before I do that. I don't actually see any thing that says those were deleted intentionally.

btracey:
As is currently set up, I get the same linking error (about undefined symbols). Do I need to change anything?

kortschak:
You shouldn't need to. @vladimir-ch can you replicate this failure?

vladimir-ch:
It took me a while to figure out that you are using the develop branch. Yes, I get the same linking errors as @btracey

vladimir-ch:
I think the cause is `OpenBLAS/lapack-netlib/SRC/Makefile:451`, the deprecated routines are not built.

vladimir-ch:
No, that doesn't seem to be it.

vladimir-ch:
`OpenBLAS/lapack-netlib/LAPACKE/src/Makefile:2201`

vladimir-ch:
`make BUILD_DEPRECATED=1` works for me

kortschak:
OK, that makes sense. I can now replicate the failure - artifacts for those functions were left behind and linked in, so hosing the repo and re-checking out gives the failure.

So this PR should also add a list of deprecated functions to exclude.

kortschak:
PTAL

btracey:
Could you explain this a bit? It seems to delete a lot of functions "LAPACKE_zuncsd2by1" for example. Are these being determined from a master list somewhere?

Is the copyright year change intentional?

kortschak:
I got the list of deprecated functions from [here](https://github.com/xianyi/OpenBLAS/blob/develop/lapack-netlib/LAPACKE/src/Makefile#L2122) and the copyright date is a function of the fact that the header files are just copied from [here](https://github.com/xianyi/OpenBLAS/tree/develop/lapack-netlib/LAPACKE/include).

btracey:
Code compiles correctly for me, so LGTM

vladimir-ch:
Just for my better understanding of the big picture (a bit too late): this PR requires OpenBLAS to be compiled from the develop branch. It is not difficult at all but still it is a slightly harsh requirement that should be at least mentioned somewhere. Also, what is the motivation or justification of this? Otherwise we are rather conservative.

kortschak:
For whatever reason (I don't understand this myself), OpenBLAS's [default branch is develop](https://github.com/xianyi/OpenBLAS/branches).

The motivation is that it fixes the build - when we pull for travis we get develop and so our builds were breaking since the deprecated functions were not being built.

vladimir-ch:
> For whatever reason (I don't understand this myself), OpenBLAS's default branch is develop.

Interesting. I'm wondering how many packages actually use the develop branch. I didn't do any rigorous comparison but building our clapack against OpenBLAS master seemed to produce the same errors as against my Fedora 22 system OpenBLAS.

btracey:
Other groups are at least struggling with the issue on what to do with these deprecations and the breaks they cause. https://github.com/scipy/scipy/issues/5266

The relevant factor in our case is that it seems we are the first group to actually test the RowMajor version of Lapacke. There are a bunch of memory errors and other size problems. See, for example, the two bugs I discovered in https://github.com/gonum/lapack/pull/74 alone. Using older versions of OpenBLAS give you incorrect results. Current versions do too, but that's only a stronger argument to stay with develop for the time being.

vladimir-ch:
> Using older versions of OpenBLAS give you incorrect results. Current versions do too ...

That's not nice but I guess the alternatives are limited. Out of curiosity, does the Accelerate framework on OSX have similar issues? Or has someone tested MKL?

btracey:
The problems are with Lapack functions. I don't know of alternate Lapack implementations except for some specific functions, so all Lapack versions would have these problems (since they all call the same C API)

vladimir-ch:
You mean that (y)our implementation of Lapack in native Go is not something common? All others end up calling the same Fortran code?

btracey:
As far as I know, yes. OpenBLAS has a few functions, ATLAS has a few
functions. Mkl may have everything, they have the resources, knowledge and
market that they may care enough.

It must be the case that we're the first to really test the row major
implementation of lapack given the nature of the bugs we've seen

vladimir-ch:
Wow. Thanks.

PR #78, opened 2015-12-12
Add Dbdsqr and test

btracey:


cgo/lapack.go
```diff
@@ -171,6 +171,74 @@ func (impl Implementation) Dpotrf(ul blas.Uplo, n int, a []float64, lda int) (ok
...
+		panic(badD)
+	}
+	if len(e) < n-1 {
+		panic(badE)
+	}
+	if len(work) < 4*n {
+		panic(badWork)
+	}
+	// An address must be passed to cgo. If lengths are zero, allocate a slice.
+	if len(vt) == 0 {
```
kortschak: I could conceivably add a helper to clapack that passes nil for zero-length slices which would obviate this. What do you think?

native/dbdsqr.go
```diff
@@ -0,0 +1,493 @@
...
+//
+// U is a matrix of size nru×n whose elements are stored in u. The elements
+// of u are modified to contain U * Q on exit. U is not used if nru == 0.
+//
+// C is a matrix of size n×ncc whose elements are stored in c. The elements
+// of c are modified to contain Q^T * C on exit. C is not used if ncc == 0.
+//
+// work contains temporary storage and must have length at least 4*n. Dbdsqr
+// will panic if there is insufficient working memory.
+func (impl Implementation) Dbdsqr(uplo blas.Uplo, n, ncvt, nru, ncc int, d, e, vt []float64, ldvt int, u []float64, ldu int, c []float64, ldc int, work []float64) (ok bool) {
```
kortschak: Add comment indicating what ok means (also above).

cgo/lapack.go
btracey: I don't think that's necessary. This is the first function we've had a problem with. It's not a memory corruption, just an out if bounds. If we find a lot that need it we can reconsider, but not needed yet I don't think

native/dbdsqr.go
```diff
@@ -0,0 +1,493 @@
...
+	const (
+		hndrth    = 1.0 / 100
+		negEighth = -1.0 / 8
+		maxIter   = 6
+	)
+	if n == 0 {
+		return true
+	}
+	rotate := ncvt > 0 || nru > 0 || ncc > 0
+	if n == 1 {
```
kortschak: Can this not just be an `if n != 1 {` and delete to L85 and the label?

native/dbdsqr.go
```diff
@@ -0,0 +1,493 @@
...
+	bi := blas64.Implementation()
+	const (
+		hndrth    = 1.0 / 100
+		negEighth = -1.0 / 8
+		maxIter   = 6
+	)
+	if n == 0 {
+		return true
+	}
+	rotate := ncvt > 0 || nru > 0 || ncc > 0
```
kortschak: This is only used once, in the block below. Define in there, just above the condition.

native/dbdsqr.go
```diff
@@ -0,0 +1,493 @@
...
+		tol := tolmul * eps
+		var thresh float64
+		var smax float64
+		for i := 0; i < n; i++ {
+			smax = math.Max(smax, math.Abs(d[i]))
+		}
+		for i := 0; i < n-1; i++ {
+			smax = math.Max(smax, math.Abs(e[i]))
+		}
+		var smin1 float64
```
kortschak: This is `sminl` in the fortran.

native/dbdsqr.go
```diff
@@ -0,0 +1,493 @@
...
+		maxIt := maxIter * n * n
+		iter := 0
+		oldll := -1
+		oldm := -1
+		// m points to the last element of unconverged part of matrix.
+		m := n
+
+	Outer:
+		for {
+			if m <= 1 {
```
kortschak: `for m > 1 {`

native/dbdsqr.go
```diff
@@ -0,0 +1,493 @@
...
+				info = 0
+				for i := 0; i < n-1; i++ {
+					if e[i] != 0 {
+						info++
+					}
+				}
+				return info == 0
+			}
+			// Find diagonal block of matrix to work on.
+			if tol < 0 && d[m-1] <= thresh {
```
kortschak: Is `d[m-1]` always positive? The fortran doesn't think so.

native/dbdsqr.go
```diff
@@ -0,0 +1,493 @@
...
+					d[ll] = 0
+				}
+				if abse <= thresh {
+					broke = true
+					break
+				}
+				smin = math.Min(smin, abss)
+				smax = math.Max(math.Max(smax, abss), abse)
+			}
+			if broke {
```
kortschak: It seems to me that this would be simpler if the code in this block were place in the condition that sets `broke`. The else clause is then omitted. The continue becomes `continue Outer`.

native/dbdsqr.go
```diff
@@ -0,0 +1,493 @@
...
+					continue
+				}
+			} else {
+				ll = -1
+			}
+			ll++
+			// e[ll] through e[m-2] are nonzero, e[ll-1] is zero
+			if ll == m-2 {
+				// Handle 2×2 block separately.
+				sigmn, sigmx, sinr, cosr, sin1, cos1 := impl.Dlasv2(d[m-2], e[m-2], d[m-1])
```
kortschak:
```
var sinr, cosr, sin1, cos1 float64
d[m-1], d[m-2], sinr, cosr, sin1, cos1 = impl.Dlasv2(d[m-2], e[m-2], d[m-1])
e[m-2] = 0
```
?

kortschak: These are `sinl` and `cosl` in the fortran which makes more sense given the `r` suffix on the other returns.

native/dbdsqr.go
```diff
@@ -0,0 +1,493 @@
...
+				if nru > 0 {
+					bi.Drot(nru, u[m-2:], ldu, u[m-1:], ldu, cos1, sin1)
+				}
+				if ncc > 0 {
+					bi.Drot(ncc, c[(m-2)*ldc:], 1, c[(m-1)*ldc:], 1, cos1, sin1)
+				}
+				m -= 2
+				continue
+			}
+			// If working on a new submatrix, chose shift direction from larger end
```
kortschak: choose

native/dbdsqr.go
```diff
@@ -0,0 +1,493 @@
...
+			// diagonal element to smaller.
```
kortschak: s/to/toward/

native/dbdsqr.go
```diff
@@ -0,0 +1,493 @@
...
+					idir = 1
+				} else {
+					idir = 2
+				}
+			}
+			// Apply convergence tests.
+			if idir == 1 {
+				// Run convergence test in forward direction.
+				// First apply standard test to bottom of matrix.
+				if math.Abs(e[m-2]) <= math.Abs(tol)*math.Abs(d[m-1]) || (tol < 0 && math.Abs(e[m-2]) <= thresh) {
```
kortschak: We probably want to go over this later to reduce the amount of duplicated work.

native/dbdsqr.go
```diff
@@ -0,0 +1,493 @@
...
+				// First apply standard test to top of matrix.
+				if math.Abs(e[ll]) <= math.Abs(tol)*math.Abs(d[ll]) || (tol < 0 && math.Abs(e[ll]) <= thresh) {
+					e[ll] = 0
+					continue
+				}
+				if tol >= 0 {
+					// If relative accuracy desired, apply convergence criterion backward.
+					mu := math.Abs(d[m-1])
+					smin1 = mu
+					for lll := m - 2; lll >= ll; lll-- {
```
kortschak: All this `l`, `ll`, `lll` make me queasy. What do you think about `l`, `l2` and `l3`? Also horrible, but easier to see errors.

native/dbdsqr.go
```diff
@@ -0,0 +1,493 @@
...
+			}
+			oldll = ll
+			oldm = m
+			// Compute shift.  First, test if shifting would ruin relative accuracy,
+			// and if so set the shift to zero.
+			var shift, r float64
+			if tol >= 0 && float64(n)*tol*(smin1/smax) <= math.Max(eps, hndrth*tol) {
+				shift = 0
+			} else {
+				var s11 float64
```
kortschak: `sll` (or `sl2`).

native/dbdsqr.go
```diff
@@ -0,0 +1,493 @@
...
+					shift, r = impl.Dlas2(d[ll], e[ll], d[ll+1])
+				}
+				// Test if shift is negligible
+				if s11 > 0 {
+					if (shift/s11)*(shift/s11) < eps {
+						shift = 0
+					}
+				}
+			}
+			_ = r
```
kortschak: Needed?

native/dbdsqr.go
```diff
@@ -0,0 +1,493 @@
...
+			iter += m - ll + 1
+			// Line 607
```
kortschak: Mark for you?

native/dbdsqr.go
```diff
@@ -0,0 +1,493 @@
...
+			// If shift == 0, do simplified QR iteration.
```
kortschak: `If no shift, do ...`

native/dbdsqr.go
```diff
@@ -0,0 +1,493 @@
...
+						work[i-ll] = cs
+						work[i-ll+nm1] = sn
+						work[i-ll+nm12] = oldcs
+						work[i-ll+nm13] = oldsn
+					}
+					h := d[m-1] * cs
+					d[m-1] = h * oldcs
+					e[m-2] = h * oldsn
+					if ncvt > 0 {
+						impl.Dlasr(blas.Left, lapack.Variable, lapack.Forward, m-ll, ncvt, work[0:], work[n-1:], vt[ll*ldvt:], ldvt)
```
kortschak: `work` without slice?

native/dbdsqr.go
```diff
@@ -0,0 +1,493 @@
...
+			for i := 0; i < n-1; i++ {
+				cs, sn, r = impl.Dlartg(d[i], e[i])
+				d[i] = r
+				e[i] = sn * d[i+1]
+				d[i+1] *= cs
+				work[i] = cs
+				work[nm1+i] = sn
+			}
+			if nru > 0 {
+				impl.Dlasr(blas.Right, lapack.Variable, lapack.Forward, nru, n, work[0:], work[n-1:], u, ldu)
```
kortschak: `work` without slice? (and next case).

native/dbdsqr.go
```diff
@@ -0,0 +1,493 @@
...
+						work[i-ll+nm13-1] = -oldsn
+					}
+					h := d[ll] * cs
+					d[ll] = h * oldcs
+					e[ll] = h * oldsn
+					if ncvt > 0 {
+						impl.Dlasr(blas.Left, lapack.Variable, lapack.Backward, m-ll, ncvt, work[nm12:], work[nm13:], vt[ll*ldvt:], ldvt)
+					}
+					if nru > 0 {
+						impl.Dlasr(blas.Right, lapack.Variable, lapack.Backward, nru, m-ll, work[0:], work[n-1:], u[ll:], ldu)
```
kortschak: Unnecessary slicing.

native/dbdsqr.go
```diff
@@ -0,0 +1,493 @@
...
+					}
+				}
+			} else {
+				// Use nonzero shift.
+				if idir == 1 {
+					// Chase bulge from top to bottom. Save cosines and sines for
+					// later singular vector updates.
+					f := (math.Abs(d[ll]) - shift) * (math.Copysign(1, d[ll]) + shift/d[ll])
+					g := e[ll]
+					var cos1, sin1 float64
```
kortschak: `cosl` and `sinl`

native/dbdsqr.go
```diff
@@ -0,0 +1,493 @@
...
+							e[i+1] = cos1 * e[i+1]
+						}
+						work[i-ll] = cosr
+						work[i-ll+nm1] = sinr
+						work[i-ll+nm12] = cos1
+						work[i-ll+nm13] = sin1
+					}
+					e[m-2] = f
+					if ncvt > 0 {
+						impl.Dlasr(blas.Left, lapack.Variable, lapack.Forward, m-ll, ncvt, work[0:], work[n-1:], vt[ll*ldvt:], ldvt)
```
kortschak: Unnecessary slice.

native/dbdsqr.go
```diff
@@ -0,0 +1,493 @@
...
+					}
+					if math.Abs(e[m-2]) <= thresh {
+						e[m-2] = 0
+					}
+				} else {
+					// Chase bulge from top to bottom. Save cosines and sines for
+					// later singular vector updates.
+					f := (math.Abs(d[m-1]) - shift) * (math.Copysign(1, d[m-1]) + shift/d[m-1])
+					g := e[m-2]
+					for i := m - 1; i >= ll+1; i-- {
```
kortschak: `i > ll`

native/dbdsqr.go
```diff
@@ -0,0 +1,493 @@
...
+						cosr, sinr, r := impl.Dlartg(f, g)
+						if i < m-1 {
+							e[i] = r
+						}
+						f = cosr*d[i] + sinr*e[i-1]
+						e[i-1] = cosr*e[i-1] - sinr*d[i]
+						g = sinr * d[i-1]
+						d[i-1] *= cosr
+						cos1, sin1, r := impl.Dlartg(f, g)
```
kortschak: `cosl` and `sinl`

native/dbdsqr.go
```diff
@@ -0,0 +1,493 @@
...
+					}
+					e[ll] = f
+					if math.Abs(e[ll]) <= thresh {
+						e[ll] = 0
+					}
+					if ncvt > 0 {
+						impl.Dlasr(blas.Left, lapack.Variable, lapack.Backward, m-ll, ncvt, work[nm12:], work[nm13:], vt[ll*ldvt:], ldvt)
+					}
+					if nru > 0 {
+						impl.Dlasr(blas.Right, lapack.Variable, lapack.Backward, nru, m-ll, work[0:], work[n-1:], u[ll:], ldu)
```
kortschak: Unnecessary slicing here and below.

native/dbdsqr.go
```diff
@@ -0,0 +1,493 @@
...
+					}
+					if ncc > 0 {
+						impl.Dlasr(blas.Left, lapack.Variable, lapack.Backward, m-ll, ncc, work[0:], work[n-1:], c[ll*ldc:], ldc)
+					}
+				}
+			}
+		}
+		// All singular values converged, make them positive.
```
kortschak: This comment seems out of place. Move below the closing brace, presumably just above the following for loop - add a blank line before the comment.

native/dbdsqr.go
```diff
@@ -0,0 +1,493 @@
...
+OneSixty:
+	for i := 0; i < n; i++ {
+		if d[i] < 0 {
+			d[i] *= -1
+			if ncvt > 0 {
+				bi.Dscal(ncvt, -1, vt[i*ldvt:], 1)
+			}
+		}
+	}
+	// Sort the singular values in decreasing order.
```
kortschak: Insert blank line above.

native/dbdsqr.go
btracey: Done.

native/dbdsqr.go
btracey: Done

native/dbdsqr.go
btracey: Done.

native/dbdsqr.go
btracey: Done.

native/dbdsqr.go
btracey: Done.

native/dbdsqr.go
btracey: Done. Thanks.

native/dbdsqr.go
btracey: Thanks.

native/dbdsqr.go
btracey: I thought about that, but how do I cleanly deal with ll getting reset to 0 only if break did not happen?

native/dbdsqr.go
btracey: Done.

native/dbdsqr.go
btracey: Done.

native/dbdsqr.go
btracey: Done.

native/dbdsqr.go
btracey: Added note.

native/dbdsqr.go
btracey: Done.

native/dbdsqr.go
btracey: Done.

native/dbdsqr.go
btracey: Gone.

native/dbdsqr.go
btracey: Gone.

native/dbdsqr.go
btracey: Done.

native/dbdsqr.go
btracey: Done.

native/dbdsqr.go
btracey: Done.

native/dbdsqr.go
btracey: Done.

native/dbdsqr.go
btracey: Done.

native/dbdsqr.go
btracey: Done.

native/dbdsqr.go
btracey: Done.

native/dbdsqr.go
btracey: Done.

native/dbdsqr.go
btracey: Done.

native/dbdsqr.go
btracey: Done.

btracey:
PTAL

native/dbdsqr.go
kortschak: Then l2++ goes into the executed break condition and there is no other code. Even if the simplification I'm suggesting doesn't happen that should - at the moment it does nothin to l2 if broke, otherwise it decreases l2, then increases l2. It would be cleaner to not have the else block and put the l2++ in the if block.

But once you've done that you see that the transformation I'm suggesting is obvious.

btracey:
I just pushed a simplification to the logic, but it didn't remove the broke variable entirely. It's not true that the other path decremented broke and then incremented it, the other path resets it to zero. Sorry, I'm not seeing the simplification you mean.

kortschak:
Ah, sorry, the `} else { l2 = -1 }; l2++` got me. It's amazing how unusual constructions can throw me.

It's better, thanks. Now that I understand what you are doing, even this would be OK in the original location - the double action on l2 in the else case was not good:

```
if broke {
	e[l2] = 0
	if l2 == m-2 {
		// Convergence of bottom singular value, return to top.
		m--
		continue
	}
	l2++
} else {
	l2 = 0
}
```

LGTM but wait for @vladimir-ch.

cgo/lapack.go
```diff
@@ -171,6 +171,76 @@ func (impl Implementation) Dpotrf(ul blas.Uplo, n int, a []float64, lda int) (ok
...
+// The SVD of the bidiagonal matrix B is
+//  B = Q * S * P^T
+// where S is a diagonal matrix of singular values, Q is an orthogonal matrix of
+// left singular vectors, and P is an orthogonal matrix of right singular vectors.
+//
+// Q and P are only computed if requested. If left singular vectors are requested,
+// this routine returns U * Q instead of Q, and if right singular vectors are
+// requested P^T * VT is returned instead of P^T.
+//
+// Frequently Dbdsqr is used in conjuction with Dgebrd which reduced a general
```
vladimir-ch: `reduces` instead of `reduced`?

native/dbdsqr.go
```diff
@@ -0,0 +1,486 @@
...
+// The SVD of the bidiagonal matrix B is
+//  B = Q * S * P^T
+// where S is a diagonal matrix of singular values, Q is an orthogonal matrix of
+// left singular vectors, and P is an orthogonal matrix of right singular vectors.
+//
+// Q and P are only computed if requested. If left singular vectors are requested,
+// this routine returns U * Q instead of Q, and if right singular vectors are
+// requested P^T * VT is returned instead of P^T.
+//
+// Frequently Dbdsqr is used in conjuction with Dgebrd which reduced a general
```
vladimir-ch: Also here.

native/dbdsqr.go
```diff
@@ -0,0 +1,486 @@
...
+		maxIter   = 6
+	)
+	if n == 0 {
+		return true
+	}
+	if n != 1 {
+		// If the singular vectors do not need to be computed, use qd algorithm.
+		if !(ncvt > 0 || nru > 0 || ncc > 0) {
+			info = impl.Dlasq1(n, d, e, work)
+			// If info is 2 dqds didn't finish, and so try to.
```
vladimir-ch: Could you extend the `try to` a bit?

native/dbdsqr.go
```diff
@@ -0,0 +1,486 @@
...
+	if len(e) < n-1 {
+		panic(badE)
+	}
+	if len(work) < 4*n {
+		panic(badWork)
+	}
+	var info int
+	bi := blas64.Implementation()
+	const (
+		hndrth    = 1.0 / 100
```
vladimir-ch: `hndrth` and `negEightth` are only used at one place further down. Remove them or do you want to keep them for consistency with the reference code?

native/dbdsqr.go
```diff
@@ -0,0 +1,486 @@
...
+		negEighth = -1.0 / 8
+		maxIter   = 6
```
vladimir-ch: Move `maxIter` closer to where the constant is used? Not really important.

native/dbdsqr.go
```diff
@@ -0,0 +1,486 @@
...
+			}
+			if ncc > 0 {
+				impl.Dlasr(blas.Left, lapack.Variable, lapack.Forward, n, ncc, work, work[n-1:], c, ldc)
+			}
+		}
+		// Compute singular values to a relative accuracy of tol. If tol is negative
+		// the values will be computed to an absolute accuracy of math.Abs(tol) * norm(b)
+		tolmul := math.Max(10, math.Min(100, eps*negEighth))
+		tol := tolmul * eps
+		var thresh float64
```
vladimir-ch: Move this line down to 129?

vladimir-ch:
LGTM

cgo/lapack.go
btracey: Done

native/dbdsqr.go
btracey: Done.

native/dbdsqr.go
btracey: ... no? Dlasq1 reduces to bidiagonal form. It's not clear to me why that fails and why the alternate algorithm here may be better.

native/dbdsqr.go
btracey: Done.

native/dbdsqr.go
btracey: Done.

btracey:
Man, these greens don't seem to be a fluke. That's awesome.

native/dbdsqr.go
kortschak: I think @vladimir-ch was asking for "try to" to have an explicit object.

vladimir-ch: Yes, I was, but it is not so important. I always struggle to clearly and concisely express what I mean while you two (unsurprisingly of course) understand each other's language shortcuts (for me often obscure) without any trouble. Argh.

kortschak: I didn't really like that phrasing either, but I couldn't figure out how to fix it. Maybe "If info is 2 dqds didn't finish, and so try to do so now." Not nice, but the sentence is clearly terminated.

There should be a follow-up PR from @btracey to fix this and the comment review comment (urgh) in the previous PR.

PR #79, opened 2015-12-16
Adddgesvd

btracey:


btracey:
Woo!

This provides a partial implementation of Dgesvd. It only supports the case where the full U and V are computed. We had previously thought we may only support computing U and V (no option to not). I'm not entirely sure anymore this is a good idea, as this is a huge amount of memory overhead if m >> n.

A full implementation of the function is ~ 10x more work. We could possibly not support overwriting the initial A. This would save a bunch of work as this is the confusing case. The strides are hard to parse because of row/column major differences. 

@kortschak Do you know what the "superb" field means? Is it possible to have a specific SVDJob type, or should I switch those types?

I fixed a bug in Dorgbr and edited the tests, but I'm not convinced the current implementation is correct. It seems like Q should have to be m x n and not m x k, but the SVD tests fail if I change that logic. I don't understand from the comment why Q should be m x k. If someone could take a second glance it would be appreciated.

kortschak:
> Do you know what the "superb" field means? Is it possible to have a specific SVDJob type, or should I switch those types?

Where?

btracey:
Superb field is the last entry in the clapack code. I have work there, but I'm not sure that's what that means (no other function so far takes in temporary memory). The svdjob type is an individual one I created here, the autogenerated code takes in an unnamed job

btracey:
I'm almost done with the svd implementation ignoring overwrite. I then need to figure out how to write tests. No point in reviewing until that's correct (aside from cgo questions)

kortschak:
It's a place to retain the 1 to min(m-2,n-2)th elements of work. Seems sort of odd.

https://github.com/xianyi/OpenBLAS/blob/develop/lapack-netlib/LAPACKE/src/lapacke_dgesvd.c

btracey:
Okay, found a few bugs that make things make more sense. Getting close to a working implementation.

btracey:
It almost works now. It works if dgebd2 is called instead of the loop in dgebrd. This implies there is something wrong with the Dgebrd loop, but I don't see it.

btracey:
Found it finally

btracey:
Okay! I think this works now.

I don't know how to test the "insufficient work" branches in an automated manner as the recommended work is bigger than just one greater. I did test them by hand and they work.

PTAL!

cgo/lapack.go
```diff
@@ -497,6 +498,83 @@ func (impl Implementation) Dgels(trans blas.Transpose, m, n, nrhs int, a []float
...
+// The singular value decomposition is
+//  A = U * Sigma * V^T
+// where Sigma is an m×n diagonal matrix containing the singular values of A,
+// U is an m×m orthogonal matrix and V is an n×n orthogonal matrix. The first
+// min(m,n) columns of U and V are the left and right singular vectors of A
+// respectively.
+//
+// jobU and jobVT are options for computing the singular vectors. The behavior
+// is as follows
+//  jobU == lapack.SVDAll		All M columns of U are returned in u
```
vladimir-ch: Here and on the next three lines use spaces for the alignment of the second column.

cgo/lapack.go
```diff
@@ -497,6 +498,83 @@ func (impl Implementation) Dgels(trans blas.Transpose, m, n, nrhs int, a []float
...
+//  jobU == lapack.SVDInPlace	The first min(m,n) columns are returned in u
+//  jobU == lapack.SVDOverwrite	The first min(m,n) columns of U are written into a
+//	jobU == lapack.SVDNone		The columns of U are not computed.
```
vladimir-ch: Use two spaces instead of one tab in front of `jobU`

cgo/lapack.go
vladimir-ch: Also: All ~~M~~ m columns ...

lapack.go
```diff
@@ -98,3 +98,13 @@ const (
 	ApplyP DecompUpdate = 'P'
 	ApplyQ DecompUpdate = 'Q'
 )
+
+// SVDJob specifies the singular vector computation type for SVD.
+type SVDJob byte
+
+const (
+	SVDAll       SVDJob = 'A' // Compute all singular vectors
+	SVDInPlace          = 'S' // Compute the first singular vectors and store in provided storage.
```
vladimir-ch: delete `first`?

`... store them ...`?

lapack.go
```diff
@@ -98,3 +98,13 @@ const (
...
+	SVDOverwrite        = 'O' // Compute the singular vectors and store in input matrix
```
vladimir-ch: `... store them ...`?

native/dgesvd.go
```diff
@@ -0,0 +1,1245 @@
...
+//
+// work is a slice for storing temporary memory, and lwork is the usable size of
+// the slice. lwork must be at least max(5*min(m,n), 3*min(m,n)+max(m,n)).
+// If lwork == -1, instead of performing Dgesvd, the optimal work length will be
+// stored into work[0]. Dgesvd will panic if the working memory has insufficient
+// storage.
+//
+// Dgesvd returns whether the decomposition successfully completed.
+func (impl Implementation) Dgesvd(jobU, jobVT lapack.SVDJob, m, n int, a []float64, lda int, s, u []float64, ldu int, vt []float64, ldvt int, work []float64, lwork int) (ok bool) {
+
```
vladimir-ch: Delete empty line

testlapack/dgesvd.go
```diff
@@ -0,0 +1,294 @@
...
+func DgesvdTest(t *testing.T, impl Dgesvder) {
+	// TODO(btracey): Add tests for all of the cases when the SVD implementation
+	// is finished.
+	// TODO(btracey): Add tests for m > mnthr and n > mnthr when other SVD
+	// conditions are implemented. Right now mnthr is 5,000,000 which is too
+	// large to create a square matrix of that size.
+	for _, test := range []struct {
+		m, n, lda, ldu, ldvt int
+	}{
+		/*
```
vladimir-ch: Why are you keeping these? If keep, then at least better aligned?

native/dgesvd.go
```diff
@@ -0,0 +1,1245 @@
...
+// The singular value decomposition is
+//  A = U * Sigma * V^T
+// where Sigma is an m×n diagonal matrix containing the singular values of A,
+// U is an m×m orthogonal matrix and V is an n×n orthogonal matrix. The first
+// min(m,n) columns of U and V are the left and right singular vectors of A
+// respectively.
+//
+// jobU and jobVT are options for computing the singular vectors. The behavior
+// is as follows
+//  jobU == lapack.SVDAll		All M columns of U are returned in u
```
vladimir-ch: Same here

vladimir-ch:
I'll try to take a look at the code itself later.

cgo/lapack.go

cgo/lapack.go
btracey: Oops, thanks.

lapack.go
btracey: No, first is correct. All computes all of them, InPlace computes min(m,n) of them.

btracey: Added them.

lapack.go
btracey: Done.

native/dgesvd.go
btracey: Done.

native/dgesvd.go
btracey: Done.

testlapack/dgesvd.go
btracey: Sorry, they should all be there. Too much debugging.

btracey:
Done. Thanks for the comments.

lapack.go
vladimir-ch: I see. In that case the comment for `SVDOverwrite` should also have `first`.

cgo/lapack.go
```diff
@@ -497,6 +498,83 @@ func (impl Implementation) Dgels(trans blas.Transpose, m, n, nrhs int, a []float
...
+// s is a slice of length at least min(m,n) and on exit contains the singular
+// values in decreasing order.
+//
+// u contains the left singular vectors on exit, stored columnwise. If
+// jobU == lapack.SVDAll, u is of size m×m. If jobU == lapack.SVDInPlace u is
+// of size m×min(m,n). If jobU == lapack.SVDOverwrite or lapack.SVDNone, u is
+// not used.
+//
+// vt contains the left singular vectors on exit, stored rowwise. If
+// jobV == lapack.SVDAll, vt is of size n×m. If jobV == lapack.SVDInPlace vt is
```
vladimir-ch: `jobV` should be `jobVT`

cgo/lapack.go
```diff
@@ -497,6 +498,83 @@ func (impl Implementation) Dgels(trans blas.Transpose, m, n, nrhs int, a []float
...
+// of size min(m,n)×n. If jobU == lapack.SVDOverwrite or lapack.SVDNone, vt is
```
vladimir-ch: `jobU` should be `jobVT`

native/dgesvd.go
```diff
@@ -0,0 +1,1244 @@
...
+// s is a slice of length at least min(m,n) and on exit contains the singular
+// values in decreasing order.
+//
+// u contains the left singular vectors on exit, stored columnwise. If
+// jobU == lapack.SVDAll, u is of size m×m. If jobU == lapack.SVDInPlace u is
+// of size m×min(m,n). If jobU == lapack.SVDOverwrite or lapack.SVDNone, u is
+// not used.
+//
+// vt contains the left singular vectors on exit, stored rowwise. If
+// jobV == lapack.SVDAll, vt is of size n×m. If jobV == lapack.SVDInPlace vt is
```
vladimir-ch: Same here

cgo/lapack.go
```diff
@@ -497,6 +498,83 @@ func (impl Implementation) Dgels(trans blas.Transpose, m, n, nrhs int, a []float
...
+		checkMatrix(min(m, n), n, vt, ldvt)
+	}
+	if jobU == lapack.SVDOverwrite && jobVT == lapack.SVDOverwrite {
+		panic("lapack: both jobU and jobVT are lapack.SVDOverwrite")
+	}
+	if len(s) < min(m, n) {
+		panic(badS)
+	}
+	if jobU == lapack.SVDOverwrite || jobVT == lapack.SVDOverwrite {
+		panic("lapack: SVD not coded to overwrite original matrix")
```
vladimir-ch: Use `noSVDO`?

native/dgesvd.go
```diff
@@ -0,0 +1,1244 @@
...
+						itauq := ie + n
+						itaup := itauq + n
+						iwork = itaup + n
+
+						// Bidiagonalize R in work[iu], copying result to VT.
+						impl.Dgebrd(n, n, work[iu:], ldworku, s, work[ie:], work[itauq:], work[itaup:], work[iwork:], lwork-iwork)
+						impl.Dlacpy(blas.Upper, n, n, work[iu:], ldworku, vt, ldvt)
+
+						// Generate left bidiagonalizing vectors in work[iu]
+
```
vladimir-ch: Delete line

cgo/lapack.go
```diff
@@ -497,6 +498,83 @@ func (impl Implementation) Dgels(trans blas.Transpose, m, n, nrhs int, a []float
...
+// not used.
+//
+// vt contains the left singular vectors on exit, stored rowwise. If
+// jobV == lapack.SVDAll, vt is of size n×m. If jobV == lapack.SVDInPlace vt is
+// of size min(m,n)×n. If jobU == lapack.SVDOverwrite or lapack.SVDNone, vt is
+// not used.
+//
+// The C interface does not support providing temporary storage. To provide compatibility
+// with native, lwork == -1 will not run Dgeqrf but will instead write the minimum
+// work necessary to work[0]. If len(work) < lwork, Dgeqrf will panic.
```
kortschak: s/Dgeqrf/Dgesvd/

cgo/lapack.go
```diff
@@ -497,6 +498,83 @@ func (impl Implementation) Dgels(trans blas.Transpose, m, n, nrhs int, a []float
...
+// respectively.
+//
+// jobU and jobVT are options for computing the singular vectors. The behavior
+// is as follows
+//  jobU == lapack.SVDAll       All M columns of U are returned in u
+//  jobU == lapack.SVDInPlace   The first min(m,n) columns are returned in u
+//  jobU == lapack.SVDOverwrite The first min(m,n) columns of U are written into a
+//  jobU == lapack.SVDNone      The columns of U are not computed.
+// The behavior is the same for jobVT and the rows of V^T. At most one of jobU
+// and jobVT can equal lapack.SVDOverwrite.
```
kortschak: Add that a panic happens otherwise.

cgo/lapack.go
kortschak: Is this because the LAPACKE doesn't support it or we don't copy back?

btracey: Neither. It's because it's about 2x the work to implement it, and I was trying to allow us to shift matrix over to this faster. There's a lot of trickiness with strides and col/row major. I can implement it if we want completeness for it to go in.

kortschak: OK, so this is just to reflect the behaviour in the Go implementation?

There's no need to make the implementation complete for this PR.

cgo/lapack.go
```diff
@@ -497,6 +498,83 @@ func (impl Implementation) Dgels(trans blas.Transpose, m, n, nrhs int, a []float
...
+		}
+		if lwork < minWork {
+			panic(badWork)
+		}
+	}
+	if lwork == -1 {
+		work[0] = float64(minWork)
+		return true
+	}
+	return clapack.Dgesvd(lapack.Job(jobU), lapack.Job(jobVT), m, n, a, lda, s, u, ldu, vt, ldvt, work)
```
kortschak: I don't think this use of work is correct here. The [LAPACKE code](https://github.com/xianyi/OpenBLAS/blob/develop/lapack-netlib/LAPACKE/src/lapacke_dgesvd.c#L74) index-shifts the values in work left one into superb, so work should be passed in as work[1:].

cgo/lapack.go
btracey: I see. Yes. It seems to me like trying to have consistent behavior between implementations is more important.

cgo/lapack.go
btracey: That is very strange.

cgo/lapack.go

cgo/lapack.go

cgo/lapack.go

cgo/lapack.go

cgo/lapack.go
btracey: Done

kortschak: Beyond bizarre.

native/dgesvd.go
btracey: Done.

native/dgesvd.go
btracey: Done.

native/dgesvd.go
```diff
@@ -0,0 +1,1241 @@
...
+	wantva := jobVT == lapack.SVDAll
+	wantvs := jobVT == lapack.SVDInPlace
+	wantvas := wantva || wantvs
+	wantvo := jobVT == lapack.SVDOverwrite
+	wantvn := jobVT == lapack.None
+
+	bi := blas64.Implementation()
+	var mnthr int
+
+	// The netlib implementation checks has this at only length 1. Our implementation
```
kortschak: "checks have"?

native/dgesvd.go
```diff
@@ -0,0 +1,1241 @@
...
+	// checks all input sizes before examining the l == -1 case.
+	dum := make([]float64, m*n)
```
kortschak: Would you explain how this differs in practice so I can be clearer in the review?

kortschak: Ahh, I see, these are just to get work sizes, so tau is not used - but we check it. Can we not make sure that lwork=-1 check happen before tau checks (they do currently for Dgeqrf e.g.).

native/dgesvd.go
```diff
@@ -0,0 +1,1241 @@
...
+		if m >= mnthr {
+			// m >> n
+			if wantun {
+				// Path 1
+				maxwrk = n + lwork_dgeqrf
+				maxwrk = max(maxwrk, 3*n+lwork_dgebrd)
+				if wantvo || wantvas {
+					maxwrk = max(maxwrk, 3*n+lwork_dorgbr_p)
+				}
+				maxwrk = max(maxwrk, bdspac)
```
kortschak: minwrk is ommitted because not used?

native/dgesvd.go
```diff
@@ -0,0 +1,1241 @@
...
+				// Path 9
+				wrkbl = n + lwork_dgeqrf
+				wrkbl = max(wrkbl, n+lwork_dorgqr_m)
+				wrkbl = max(wrkbl, 3*n+lwork_dgebrd)
+				wrkbl = max(wrkbl, 3*n+lwork_dorgbr_q)
+				wrkbl = max(wrkbl, 3*n+lwork_dorgbr_p)
+				wrkbl = max(wrkbl, bdspac)
+				maxwrk = n*n + wrkbl
+			}
+		} else {
```
kortschak: The fortran tests `min(m, n) > 0` here‽

native/dgesvd.go
```diff
@@ -0,0 +1,1241 @@
...
+				impl.Dorgbr(lapack.ApplyQ, m, m, n, a, lda, dum, dum, -1)
+				lwork_dorgbr_q := int(dum[0])
+				maxwrk = max(maxwrk, 3*n+lwork_dorgbr_q)
+			}
+			if !wantvn {
+				maxwrk = max(maxwrk, 3*n+lwork_dorgbr_p)
+			}
+			maxwrk = max(maxwrk, bdspac)
+		}
+	} else {
```
kortschak: The fortran tests `min(m, n) > 0` *here*‽ (now in correct spot - the bewilderment it engendered knocked me off mark).

native/dgesvd.go
```diff
@@ -0,0 +1,1241 @@
...
+		iscl = 1
+		impl.Dlascl(lapack.General, 0, 0, anrm, bignum, m, n, a, lda)
+	}
+
+	var ie int
+	if m >= n {
+		// If A has sufficiently more rows than columns, use the QR decomposition.
+		if m >= mnthr {
+			if wantun {
+				// Path 1
```
kortschak: Some of the fortran comment is useful here: "No left singular vectors to be computed."

native/dgesvd.go
```diff
@@ -0,0 +1,1241 @@
...
+	} else if anrm > bignum {
+		iscl = 1
+		impl.Dlascl(lapack.General, 0, 0, anrm, bignum, m, n, a, lda)
+	}
+
+	var ie int
+	if m >= n {
+		// If A has sufficiently more rows than columns, use the QR decomposition.
+		if m >= mnthr {
+			if wantun {
```
kortschak: `// n >> m` above.

native/dgesvd.go
```diff
@@ -0,0 +1,1241 @@
...
+				// Path 2
+				panic(noSVDO)
+			} else if wantuo && wantvas {
+				// Path 3
+				panic(noSVDO)
+			} else if wantus {
+				if wantvn {
+					// Path 4
+					if lwork >= n*n+max(4*n, bdspac) {
+						ir := 0
```
kortschak: `ir = 1` in the fortran.

kortschak: Ahh, index. nvm.

native/dgesvd.go
```diff
@@ -0,0 +1,1241 @@
...
+
+						// Generate Q in A.
+						impl.Dorgqr(m, n, n, a, lda, work[itau:], work[iwork:], lwork-iwork)
+						ie := itau
+						itauq := ie + n
+						itaup := itauq + n
+						iwork = itaup + n
+
+						// Bidiagonalize R in work[ir:].
+						impl.Dgebrd(n, n, work[ir:], ldworkr, s, work[ie:], work[itauq:], work[itaup:], work[iwork:], lwork-iwork)
```
kortschak: Split over lines - other long calls below too.

native/dgesvd.go
```diff
@@ -0,0 +1,1241 @@
...
+						// singular vectors of A in U.
+						ok = impl.Dbdsqr(blas.Upper, n, 0, m, 0, s, work[ie:], dum, 1, u, ldu, dum, 1, work[iwork:])
+					}
+				} else if wantvo {
+					// Path 5
+					panic(noSVDO)
+				} else if wantvas {
+					// Path 6
+					if lwork >= n*n+max(4*n, bdspac) {
+						// Sufficient workspace for a fast algorithm.
```
kortschak: Use this pair of comments more consistently.

native/dgesvd.go
```diff
@@ -0,0 +1,1241 @@
...
+						// Compute A = Q * R, copying result to U.
+						impl.Dgeqrf(m, n, a, lda, work[itau:], work[iwork:], lwork-iwork)
+						impl.Dlacpy(blas.Lower, m, n, a, lda, u, ldu)
+
+						// Generate Q in U
+						impl.Dorgqr(m, n, n, u, ldu, work[itau:], work[iwork:], lwork-iwork)
+
+						// Copy R to VT, zeroing out below it.
+						impl.Dlacpy(blas.Upper, n, n, a, lda, vt, ldvt)
+						if n > 1 {
```
kortschak: Why do this test here?

native/dgesvd.go
```diff
@@ -0,0 +1,1241 @@
...
+
+						// Compute A = Q * R, copying result to U
+						impl.Dgeqrf(m, n, a, lda, work[itau:], work[iwork:], lwork-iwork)
+						impl.Dlacpy(blas.Lower, m, n, a, lda, u, ldu)
+
+						// Generate Q in U
+						impl.Dorgqr(m, m, n, u, ldu, work[itau:], work[iwork:], lwork-iwork)
+						// Copy R from A to VT, zeroing out below it.
+						impl.Dlacpy(blas.Upper, n, n, a, lda, vt, ldvt)
+						if n > 1 {
```
kortschak: Why do this test here?

native/dgesvd.go
```diff
@@ -0,0 +1,1241 @@
...
+						// Compute A = Q*R, copying result to U.
+						impl.Dgeqrf(m, n, a, lda, work[itau:], work[iwork:], lwork-iwork)
+						impl.Dlacpy(blas.Lower, m, n, a, lda, u, ldu)
+
+						// Generate Q in U.
+						impl.Dorgqr(m, m, n, u, ldu, work[itau:], work[iwork:], lwork-iwork)
+
+						// Copy R from A to VT, zeroing out below it.
+						impl.Dlacpy(blas.Upper, n, n, a, lda, vt, ldvt)
+						if n > 1 {
```
kortschak: Why do this test here?

native/dgesvd.go
```diff
@@ -0,0 +1,1241 @@
...
+
+						// Perform bidiagonal QR iteration, computing left singular
+						// vectors of A in U and computing right singular vectors
+						// of A in VT.
+						impl.Dbdsqr(blas.Upper, n, n, m, 0, s, work[ie:], vt, ldvt, u, ldu, dum, 1, work[iwork:])
+					}
+				}
+			}
+		} else {
+			// Path 10.
```
kortschak: `// M at least N, but not much larger.` or similar.

native/dgesvd.go
```diff
@@ -0,0 +1,1241 @@
...
+				ncvt = n
+			}
+			if wantvn {
+				ncvt = 0
+			}
+			if !wantuo && !wantvo {
+				// Perform bidiagonal QR iteration, if desired, computing left
+				// singular vectors in U and right singular vectors in VT.
+				ok = impl.Dbdsqr(blas.Upper, n, ncvt, nru, 0, s, work[ie:], vt, ldvt, u, ldu, dum, 1, work[iwork:])
+			} else {
```
kortschak: Either have both other branches here or leave a comment that this will be two branches when the complete implementation exists.

native/dgesvd.go
```diff
@@ -0,0 +1,1241 @@
...
+				panic(noSVDO)
+			}
+		}
+	} else {
+		// A has more columns than rows. If A has sufficiently more columns than
+		// rows, first reduce using the LQ decomposition.
+		if n >= mnthr {
+			if wantvn {
```
kortschak: `// n >> m`

native/dgesvd.go
```diff
@@ -0,0 +1,1241 @@
...
+						ie := itau
+						itauq := ie + n
+						itaup := itauq + n
+						iwork = itaup + n
+
+						// Bidiagonalize R in work[iu], copying result to VT.
+						impl.Dgebrd(n, n, work[iu:], ldworku, s, work[ie:], work[itauq:], work[itaup:], work[iwork:], lwork-iwork)
+						impl.Dlacpy(blas.Upper, n, n, work[iu:], ldworku, vt, ldvt)
+
+						// Generate left bidiagonalizing vectors in work[iu]
```
kortschak: `work[iu:]` presumably (and equivalent changes below). Though `iu` is always 0 here, no (same for other equivalent cases below)?

native/dgesvd.go
```diff
@@ -0,0 +1,1241 @@
...
+
+						// Perform bidiagonal QR iteration, computing left singular
+						// vectors of A in U and computing right singular vectors
+						// of A in VT.
+						ok = impl.Dbdsqr(blas.Upper, m, n, m, 0, s, work[ie:], vt, ldvt, u, ldu, dum, 1, work[iwork:])
+					}
+				}
+			}
+		} else {
+			// Path 10t
```
kortschak: `// N greater than M, but not much larger.` or equivalent.

native/dgesvd.go
```diff
@@ -0,0 +1,1241 @@
...
+			if wantvas || wantvo {
+				ncvt = n
+			}
+			if !wantvo && !wantvo {
+				// Perform bidiagonal QR iteration, if desired, computing left
+				// singular vectors in U and computing right singular vectors in
+				// VT.
+				ok = impl.Dbdsqr(blas.Lower, m, ncvt, nru, 0, s, work[ie:], vt, ldvt, u, ldu, dum, 1, work[iwork:])
+			} else {
+				panic(noSVDO)
```
kortschak: Same as equivalent location above.

native/dgesvd.go
```diff
@@ -0,0 +1,1241 @@
...
+			}
+		}
+	}
+	minmn := min(m, n)
```
kortschak: Can you not define this earlier and use it in the space calculations as well?

native/dgesvd.go
```diff
@@ -0,0 +1,1241 @@
...
+			}
+		}
+		if ie < 1 {
+			for i := minmn - 2; i >= 0; i-- {
+				work[i+1] = work[i+ie]
+			}
+		}
+	}
+	// Undo scaling if necessary
+	if iscl == 1 {
```
kortschak: So it seems that iscl is really a boolean. Let's make it one.

kortschak:
Done dgesvd.go.

Some thought. Wow, what tedious code. Thank you so much for putting in that effort. It looks like we should be able to generate most of the cases (this is an intellectual only comment - I don't think we should). When it comes time to clean things up throughout, I think many of the chained if else can be converted to switches with improved readability. 

On a more contention point, we hand around work vectors without high bounds, where could actually provide those. The fortran doesn't do this because it can't, but I think we probably should.

kortschak:
Would you provide some direction for reviewing the other smaller changes in other files?

native/dgebrd.go
```diff
@@ -86,7 +86,7 @@ func (impl Implementation) Dgebrd(m, n int, a []float64, lda int, d, e, tauQ, ta
 				if lwork >= (m+n)*nbmin {
 					nb = lwork / (m + n)
 				} else {
-					nb = 1
+					nb = minmn
```
btracey: This change is because our loop subtracts nb rather than nx. When this if statement is met, it restricts the block size.

native/dlacpy.go
```diff
@@ -24,11 +24,10 @@ func (impl Implementation) Dlacpy(uplo blas.Uplo, m, n int, a []float64, lda int
 
 	case blas.Lower:
 		for i := 0; i < m; i++ {
-			for j := 0; j < min(i, n); j++ {
+			for j := 0; j < min(i+1, n); j++ {
```
btracey: The previous algorithm was just wrong.

native/dlaset.go
```diff
@@ -20,7 +20,7 @@ func (impl Implementation) Dlaset(uplo blas.Uplo, m, n int, alpha, beta float64,
 		}
 	} else if uplo == blas.Lower {
 		for i := 0; i < m; i++ {
-			for j := 0; j < i; j++ {
+			for j := 0; j < min(i+1, n); j++ {
```
btracey: The previous algorithm was just wrong.

native/dgebrd.go
kortschak: OK

native/dorgbr.go
```diff
@@ -30,9 +30,17 @@ func (impl Implementation) Dorgbr(vect lapack.DecompUpdate, m, n, k int, a []flo
 		}
 	}
 	if wantq {
-		checkMatrix(m, k, a, lda)
+		if m >= k {
```
btracey: I think this is clear, but the effective size of the matrix depends on the magnitude of the values.

native/dorgbr.go
```diff
@@ -72,9 +80,9 @@ func (impl Implementation) Dorgbr(vect lapack.DecompUpdate, m, n, k int, a []flo
 			// Shift the vectors which define the elementary reflectors one
 			// column to the right, and set the first row and column of Q to
 			// those of the unit matrix.
-			for j := m - 2; j >= 0; j-- {
+			for j := m - 1; j >= 1; j-- {
```
btracey: Transcription error.

native/ilaenv.go
```diff
@@ -355,7 +355,7 @@ func (Implementation) Ilaenv(ispec int, s string, opts string, n1, n2, n3, n4 in
 		return 2
 	case 6:
 		// Used by xGELSS and xGESVD
-		return min(n1, n2) * 1e6
+		return int(float64(min(n1, n2)) * 1.6e0)
```
btracey: Transcription error. The test is supposed to be if the matrix is sufficiently rectangular, not if it's sufficiently large.

testlapack/dbdsqr.go
```diff
@@ -35,6 +35,8 @@ func DbdsqrTest(t *testing.T, impl Dbdsqrer) {
 			{10, 10, 10, 10, 30, 40, 50},
 			{10, 12, 11, 13, 30, 40, 50},
 			{20, 12, 13, 11, 30, 40, 50},
+
+			{130, 130, 130, 500, 900, 900, 500},
```
btracey: This came up in debugging. This test adds a case where blocking needs to be used in the subsequent calls. I can remove if desired.

native/dlaset.go
kortschak: Yep.

native/dlacpy.go
kortschak: Yep.

native/dorgbr.go
kortschak: OK

native/dorgbr.go
kortschak: OK

native/ilaenv.go
kortschak: OK

Change the constant to `1.6`.

testlapack/dbdsqr.go
kortschak: OK

native/dgesvd.go
btracey: Done.

native/dgesvd.go
btracey: Made the comment a TODO.

native/dgesvd.go
btracey: Yea. The function comment specifically lists the actual min work, so it's not clear why it's needed.

native/dgesvd.go
btracey: Yes, very odd.

native/dgesvd.go
btracey: Done.

native/dgesvd.go
btracey: But you can see that immediately above in "if wantun"

native/dgesvd.go
btracey: Done.

native/dgesvd.go
btracey: Done.

native/dgesvd.go
btracey: Done.

native/dgesvd.go
btracey: Unclear. Deleted.

native/dgesvd.go
btracey: Gone.

native/dgesvd.go
btracey: Gone.

native/dgesvd.go
btracey: Done.

native/dgesvd.go
btracey: Done.

native/dgesvd.go
btracey: Done.

native/dgesvd.go
btracey: Done.

native/dgesvd.go
btracey: Done.

native/dgesvd.go
btracey: Done.

native/dgesvd.go
btracey: Done.

native/ilaenv.go
btracey: Done.

btracey:
Comments addressed. PTAL.

btracey:
````On a more contention point, we hand around work vectors without high bounds, where could actually provide those. The fortran doesn't do this because it can't, but I think we probably should.````
I agree in principle. This will be trickier than it seems like it should be just due to stride issues. It'll be much easier now that we have tests and a better understanding of the functions.

File: native/dgesvd.go [hunk 44]
```diff
@@ -0,0 +1,1317 @@
...
+	if jobU == lapack.SVDOverwrite && jobVT == lapack.SVDOverwrite {
+		panic("lapack: both jobU and jobVT are lapack.SVDOverwrite")
+	}
+	if len(s) < minmn {
+		panic(badS)
+	}
+	if jobU == lapack.SVDOverwrite || jobVT == lapack.SVDOverwrite {
+		panic(noSVDO)
+	}
+	minWork := max(5*minmn, 3*minmn+max(m, n))
```
kortschak: Sorry, I missed this before. Where does this come from? The use of this in [L351](https://github.com/gonum/lapack/pull/79/files#diff-02b9c83e51870657a1c6a18e52b5d942R351) explains the absence of use of minwrk - they are the same. However, I can't see where a minWork with this defintion exists.

native/dgesvd.go
```diff
@@ -0,0 +1,1317 @@
...
+				// Compute A = Q * R
+				impl.Dgeqrf(m, n, a, lda, work[itau:], work[iwork:], lwork-iwork)
+				// Zero out below R
+				impl.Dlaset(blas.Lower, n-1, n-1, 0, 0, a[lda:], lda)
+				ie = 0
+				itauq := ie + n
+				itaup := itauq + n
+				iwork = itaup + n
+				// Bidiagonalize R in A
+				impl.Dgebrd(n, n, a, lda, s, work[ie:], work[itauq:], work[itaup:], work[iwork:], lwork-iwork)
```
kortschak: Split.

native/dgesvd.go
```diff
@@ -0,0 +1,1317 @@
...
+				ncvt := 0
+				if wantvo || wantvas {
+					// Generate P^T.
+					impl.Dorgbr(lapack.ApplyP, n, n, n, a, lda, work[itaup:], work[iwork:], lwork-iwork)
```
kortschak: Split.

native/dgesvd.go
```diff
@@ -0,0 +1,1317 @@
...
+					ncvt = n
+				}
+				iwork = ie + n
+
+				// Perform bidiagonal QR iteration computing right singular vectors
+				// of A in A if desired.
+				ok = impl.Dbdsqr(blas.Upper, n, ncvt, 0, 0, s, work[ie:], a, lda, dum, 1, dum, 1, work[iwork:])
```
kortschak: Split.

native/dgesvd.go
```diff
@@ -0,0 +1,1317 @@
...
+						ie := itau
+						itauq := ie + n
+						itaup := itauq + n
+						iwork = itaup + n
+
+						// Bidiagonalize R in work[ir:].
+						impl.Dgebrd(n, n, work[ir:], ldworkr, s, work[ie:],
+							work[itauq:], work[itaup:], work[iwork:], lwork-iwork)
+
+						// Generate left vectors bidiagonalizing R in work[ir]
```
kortschak: s/ir/ir:/

native/dgesvd.go
```diff
@@ -0,0 +1,1317 @@
...
+						itauq := ie + n
+						itaup := itauq + n
+						iwork = itaup + n
+
+						// Bidiagonalize R in work[iu:], copying result to VT.
+						impl.Dgebrd(n, n, work[iu:], ldworku, s, work[ie:],
+							work[itauq:], work[itaup:], work[iwork:], lwork-iwork)
+						impl.Dlacpy(blas.Upper, n, n, work[iu:], ldworku, vt, ldvt)
+
+						// Generate left bidiagonalizing vectors in work[iu:]
```
kortschak: Period.

native/dgesvd.go
```diff
@@ -0,0 +1,1317 @@
...
+						impl.Dorgbr(lapack.ApplyQ, n, n, n, work[iu:], ldworku,
+							work[itauq:], work[iwork:], lwork-iwork)
+
+						// Generate right bidiagonalizing vectors in VT
```
kortschak: Period.

native/dgesvd.go
```diff
@@ -0,0 +1,1317 @@
...
+						impl.Dorgbr(lapack.ApplyP, n, n, n, vt, ldvt,
+							work[itaup:], work[iwork:], lwork-iwork)
+						iwork = ie + n
+
+						// Perform bidiagonal QR iteration, computing left singular
+						// vectors of R in work[iu:], and computing right singular
+						// vectors of R in VT
```
kortschak: Period.

native/dgesvd.go
```diff
@@ -0,0 +1,1317 @@
...
+						// Copy R to VT, zeroing out below it.
+						impl.Dlacpy(blas.Upper, n, n, a, lda, vt, ldvt)
+						impl.Dlaset(blas.Lower, n-1, n-1, 0, 0, vt[ldvt:], ldvt)
+
+						ie := itau
+						itauq := ie + n
+						itaup := itauq + n
+						iwork = itaup + n
+
+						// Bidiagonalize R in VT
```
kortschak: Period.

native/dgesvd.go
```diff
@@ -0,0 +1,1317 @@
...
+						impl.Dgebrd(n, n, vt, ldvt, s, work[ie:],
+							work[itauq:], work[itaup:], work[iwork:], lwork-iwork)
+
+						// Multiply Q in U by left bidiagonalizing vectors in VT.
+						impl.Dormbr(lapack.ApplyQ, blas.Right, blas.NoTrans, m, n, n,
+							vt, ldvt, work[itauq:], u, ldu, work[iwork:], lwork-iwork)
+
+						// Generate right bidiagonalizing vectors in VT
```
kortschak: Period.

native/dgesvd.go
```diff
@@ -0,0 +1,1317 @@
...
+						ok = impl.Dbdsqr(blas.Upper, n, 0, m, 0, s, work[ie:],
+							dum, 1, u, ldu, dum, 1, work[iwork:])
+					}
+				} else if wantvo {
+					// Path 8
+					panic(noSVDO)
+				} else if wantvas {
+					// Path 9
+					if lwork >= n*n+max(max(n+m, 4*n), bdspac) {
+						// Sufficient workspace for a fast algorithm
```
kortschak: Period.

native/dgesvd.go
```diff
@@ -0,0 +1,1317 @@
...
+						var ldworku int
+						if lwork >= wrkbl+lda*n {
+							ldworku = lda
+						} else {
+							ldworku = n
+						}
+						itau := iu + ldworku*n
+						iwork := itau + n
+
+						// Compute A = Q * R, copying result to U
```
kortschak: Period.

native/dgesvd.go
```diff
@@ -0,0 +1,1317 @@
...
+						impl.Dgeqrf(m, n, a, lda, work[itau:], work[iwork:], lwork-iwork)
+						impl.Dlacpy(blas.Lower, m, n, a, lda, u, ldu)
+
+						// Generate Q in U
```
kortschak: Period.

native/dgesvd.go
```diff
@@ -0,0 +1,1317 @@
...
+						// Copy R from A to VT, zeroing out below it.
+						impl.Dlacpy(blas.Upper, n, n, a, lda, vt, ldvt)
+						impl.Dlaset(blas.Lower, n-1, n-1, 0, 0, vt[ldvt:], ldvt)
+
+						ie = itau
+						itauq := ie + n
+						itaup := itauq + n
+						iwork = itaup + n
+
+						// Bidiagonalize R in VT
```
kortschak: Period.

native/dgesvd.go
```diff
@@ -0,0 +1,1317 @@
...
+						impl.Dgebrd(n, n, vt, ldvt, s, work[ie:],
+							work[itauq:], work[itaup:], work[iwork:], lwork-iwork)
+
+						// Multiply Q in U by left bidiagonalizing vectors in VT
```
kortschak: Period.

native/dgesvd.go
```diff
@@ -0,0 +1,1317 @@
...
+						impl.Dormbr(lapack.ApplyQ, blas.Right, blas.NoTrans,
+							m, n, n, vt, ldvt, work[itauq:], u, ldu, work[iwork:], lwork-iwork)
+
+						// Generate right bidiagonalizing vectors in VT
```
kortschak: Period.

native/dgesvd.go
```diff
@@ -0,0 +1,1317 @@
...
+						} else {
+							ldworkr = m
+						}
+						itau := ir + ldworkr*m
+						iwork := itau + m
+
+						// Compute A = L*Q.
+						impl.Dgelqf(m, n, a, lda, work[itau:], work[iwork:], lwork-iwork)
+
+						// Copy L to work[ir], zeroing out above it.
```
kortschak: s/ir/ir:/

native/dgesvd.go
```diff
@@ -0,0 +1,1317 @@
...
+						impl.Dlaset(blas.Upper, m-1, m-1, 0, 0, work[ir+1:], ldworkr)
+
+						// Generate Q in A.
+						impl.Dorglq(m, n, m, a, lda, work[itau:], work[iwork:], lwork-iwork)
+						ie := itau
+						itauq := ie + m
+						itaup := itauq + m
+						iwork = itaup + m
+
+						// Bidiagonalize L in work[ir]
```
kortschak: s/ir/ir:/

native/dgesvd.go
```diff
@@ -0,0 +1,1317 @@
...
+						impl.Dgebrd(m, m, work[ir:], ldworkr, s, work[ie:],
+							work[itauq:], work[itaup:], work[iwork:], lwork-iwork)
+
+						// Generate right vectors bidiagonalizing L in work[ir].
```
kortschak: s/ir/ir:/

native/dgesvd.go
```diff
@@ -0,0 +1,1317 @@
...
+						impl.Dorgbr(lapack.ApplyP, m, m, m, work[ir:], ldworkr,
+							work[itaup:], work[iwork:], lwork-iwork)
+						iwork = ie + m
+
+						// Perform bidiagonal QR iteration, computing right singular
+						// vectors of L in work[ir].
```
kortschak: s/ir/ir:/

native/dgesvd.go
```diff
@@ -0,0 +1,1317 @@
...
+						ok = impl.Dbdsqr(blas.Upper, m, m, 0, 0, s, work[ie:],
+							work[ir:], ldworkr, dum, 1, dum, 1, work[iwork:])
+
+						// Multiply right singular vectors of L in work[ir] by
```
kortschak: s/ir/ir:/

native/dgesvd.go
```diff
@@ -0,0 +1,1317 @@
...
+						} else {
+							ldworku = m
+						}
+						itau := iu + ldworku*m
+						iwork := itau + m
+
+						// Compute A = L*Q.
+						impl.Dgelqf(m, n, a, lda, work[itau:], work[iwork:], lwork-iwork)
+
+						// Copy L to work[iu], zeroing out above it.
```
kortschak: s/iu/iu:/

native/dgesvd.go
```diff
@@ -0,0 +1,1317 @@
...
+						impl.Dlaset(blas.Upper, m-1, m-1, 0, 0, work[iu+1:], ldworku)
+
+						// Generate Q in A.
+						impl.Dorglq(m, n, m, a, lda, work[itau:], work[iwork:], lwork-iwork)
+						ie := itau
+						itauq := ie + m
+						itaup := itauq + m
+						iwork = itaup + m
+
+						// Bidiagonalize L in work[iu], copying result to U.
```
kortschak: s/iu/iu:/

native/dgesvd.go
```diff
@@ -0,0 +1,1317 @@
...
+						impl.Dgebrd(m, m, work[iu:], ldworku, s, work[ie:],
+							work[itauq:], work[itaup:], work[iwork:], lwork-iwork)
+						impl.Dlacpy(blas.Lower, m, m, work[iu:], ldworku, u, ldu)
+
+						// Generate right bidiagionalizing vectors in work[iu].
```
kortschak: s/iu/iu:/

native/dgesvd.go
```diff
@@ -0,0 +1,1317 @@
...
+						impl.Dorgbr(lapack.ApplyP, m, m, m, work[iu:], ldworku,
+							work[itaup:], work[iwork:], lwork-iwork)
+
+						// Generate left bidiagonalizing vectors in U.
+						impl.Dorgbr(lapack.ApplyQ, m, m, m, u, ldu, work[itauq:], work[iwork:], lwork-iwork)
+						iwork = ie + m
+
+						// Perform bidiagonal QR iteration, computing left singular
+						// vectors of L in U and computing right singular vectors of
+						// L in work[iu].
```
kortschak: s/iu/iu:/

native/dgesvd.go
```diff
@@ -0,0 +1,1317 @@
...
+						ok = impl.Dbdsqr(blas.Upper, m, m, m, 0, s, work[ie:],
+							work[iu:], ldworku, u, ldu, dum, 1, work[iwork:])
+
+						// Multiply right singular vectors of L in work[iu] by
```
kortschak: s/iu/iu:/

native/dgesvd.go
```diff
@@ -0,0 +1,1317 @@
...
+						// Q in A, storing result in VT
```
kortschak: Period.

native/dgesvd.go
```diff
@@ -0,0 +1,1317 @@
...
+							ldworkr = m
+						}
+						itau := ir + ldworkr*m
+						iwork := itau + m
+
+						// Compute A = L*Q, copying result to VT.
+						impl.Dgelqf(m, n, a, lda, work[itau:], work[iwork:], lwork-iwork)
+						impl.Dlacpy(blas.Upper, m, n, a, lda, vt, ldvt)
+
+						// Copy L to work[ir], zeroing out above it.
```
kortschak: s/ir/ir:/

native/dgesvd.go
```diff
@@ -0,0 +1,1317 @@
...
+
+						// Generate Q in VT.
+						impl.Dorglq(n, n, m, vt, ldvt, work[itau:], work[iwork:], lwork-iwork)
+
+						ie := itau
+						itauq := ie + m
+						itaup := itauq + m
+						iwork = itaup + m
+
+						// Bidiagonalize L in work[ir].
```
kortschak: s/ir/ir:/

native/dgesvd.go
```diff
@@ -0,0 +1,1317 @@
...
+						impl.Dgebrd(m, m, work[ir:], ldworkr, s, work[ie:],
+							work[itauq:], work[itaup:], work[iwork:], lwork-iwork)
+
+						// Generate right bidiagonalizing vectors in work[ir]
```
kortschak: s/ir/ir:/

native/dgesvd.go
```diff
@@ -0,0 +1,1317 @@
...
+						impl.Dorgbr(lapack.ApplyP, m, m, m, work[ir:], ldworkr,
+							work[itaup:], work[iwork:], lwork-iwork)
+						iwork = ie + m
+
+						// Perform bidiagonal QR iteration, computing right
+						// singular vectors of L in work[ir].
```
kortschak: s/ir/ir:/

native/dgesvd.go
```diff
@@ -0,0 +1,1317 @@
...
+						ok = impl.Dbdsqr(blas.Upper, m, m, 0, 0, s, work[ie:],
+							work[ir:], ldworkr, dum, 1, dum, 1, work[iwork:])
+
+						// Multiply right singular vectors of L in work[ir] by
```
kortschak: s/ir/ir:/

native/dgesvd.go
```diff
@@ -0,0 +1,1317 @@
...
+						var ldworku int
+						if lwork >= wrkbl+lda*m {
+							ldworku = lda
+						} else {
+							ldworku = m
+						}
+						itau := iu + ldworku*m
+						iwork := itau + m
+
+						// Generate A = L * Q copying result to VT
```
kortschak: Period.

native/dgesvd.go
```diff
@@ -0,0 +1,1317 @@
...
+						impl.Dgelqf(m, n, a, lda, work[itau:], work[iwork:], lwork-iwork)
+						impl.Dlacpy(blas.Upper, m, n, a, lda, vt, ldvt)
+
+						// Generate Q in VT
```
kortschak: Period.

native/dgesvd.go
```diff
@@ -0,0 +1,1317 @@
...
+
+						// Copy L to work[iu], zeroing out above it.
+						impl.Dlacpy(blas.Lower, m, m, a, lda, work[iu:], ldworku)
+						impl.Dlaset(blas.Upper, m-1, m-1, 0, 0, work[iu+1:], ldworku)
+						ie = itau
+						itauq := ie + m
+						itaup := itauq + m
+						iwork = itaup + m
+
+						// Bidiagonalize L in work[iu], copying result to U
```
kortschak: s/iu/iu:/ and period.

native/dgesvd.go
```diff
@@ -0,0 +1,1317 @@
...
+						impl.Dgebrd(m, m, work[iu:], ldworku, s, work[ie:],
+							work[itauq:], work[itaup:], work[iwork:], lwork-iwork)
+						impl.Dlacpy(blas.Lower, m, m, work[iu:], ldworku, u, ldu)
+
+						// Generate right bidiagonalizing vectors in work[iu]
```
kortschak: s/iu/iu:/

native/dgesvd.go
```diff
@@ -0,0 +1,1317 @@
...
+						impl.Dorgbr(lapack.ApplyP, m, m, m, work[iu:], ldworku,
+							work[itaup:], work[iwork:], lwork-iwork)
+
+						// Generate left bidiagonalizing vectors in U.
+						impl.Dorgbr(lapack.ApplyQ, m, m, m, u, ldu, work[itauq:], work[iwork:], lwork-iwork)
+						iwork = ie + m
+
+						// Perform bidiagonal QR iteration, computing left singular
+						// vectors of L in U and computing right singular vectors
+						// of L in work[iu]
```
kortschak: s/iu/iu:/

native/dgesvd.go
```diff
@@ -0,0 +1,1317 @@
...
+						bi.Dgemm(blas.NoTrans, blas.NoTrans, m, n, m, 1,
+							work[iu:], ldworku, vt, ldvt, 0, a, lda)
+						// Copy right singular vectors of A from A to VT
+						impl.Dlacpy(blas.All, m, n, a, lda, vt, ldvt)
+					} else {
+						// Insufficient workspace for a fast algorithm.
+						itau := 0
+						iwork := itau + m
+
+						// Compute A = L * Q, copying result to VT
```
kortschak: Period.

native/dgesvd.go
```diff
@@ -0,0 +1,1317 @@
...
+						impl.Dgelqf(m, n, a, lda, work[itau:], work[iwork:], lwork-iwork)
+						impl.Dlacpy(blas.Upper, m, n, a, lda, vt, ldvt)
+
+						// Generate Q in VT
```
kortschak: Period.

native/dgesvd.go
```diff
@@ -0,0 +1,1317 @@
...
+						// Copy L to U, zeroing out above it.
+						impl.Dlacpy(blas.Lower, m, m, a, lda, u, ldu)
+						impl.Dlaset(blas.Upper, m-1, m-1, 0, 0, u[1:], ldu)
+
+						ie = itau
+						itauq := ie + m
+						itaup := itauq + m
+						iwork = itaup + m
+
+						// Bidiagonalize L in U
```
kortschak: Period.

native/dgesvd.go
btracey: The comment for the function says 

````
 LWORK >= MAX(1,5*MIN(M,N)) for the paths (see comments inside code):
            - PATH 1  (M much larger than N, JOBU='N') 
             - PATH 1t (N much larger than M, JOBVT='N')
   LWORK >= MAX(1,3*MIN(M,N)+MAX(M,N),5*MIN(M,N)) for the other paths
````
I somehow didn't read this second condition. Seems much easier to implement this directly than to have it coded in each path.

**Comment by kortschak** [comment]
LGTM after comments addressed, but wait for @vladimir-ch.

File: native/dgesvd.go (hunk 44 above)
Comment (kortschak) [inline native/dgesvd.go]: Are you sure the two approaches give the same results. On the surface, it doesn't look like it, but I need to take a closer look.

Comment (btracey) [inline native/dgesvd.go]: By the two approaches, you mean "as listed in the documentation" vs. "computed in the paths"?

Comment (kortschak) [inline native/dgesvd.go]: Yes

Comment (btracey) [inline native/dgesvd.go]: I'm not sure. I can compute min directly in the switches if you'd prefer.

Comment (kortschak) [inline native/dgesvd.go]: We need to wait for Vladimir, I'll have a closer look tomorrow and see if they are actually different.

Comment (btracey) [inline native/dgesvd.go]: For now I changed it to the doc comment. I'm not sure what to do on the cgo side, I'd rather not make a bunch of calls to ilaenv.

File: native/dgesvd.go (hunk 45 above)
Comment (btracey) [inline native/dgesvd.go]: Done.

File: native/dgesvd.go (hunk 46 above)
Comment (btracey) [inline native/dgesvd.go]: Done.

File: native/dgesvd.go (hunk 47 above)
Comment (btracey) [inline native/dgesvd.go]: Done.

**Comment by btracey** [comment]
Addressed PR comments, minus the outstanding issue about what to do with minwork. PTAL @vladimir-ch

File: native/dgesvd.go (hunk 44 above)
Comment (kortschak) [inline native/dgesvd.go]: I added in the path-specific minwrk calculations and compared. They are not different for any of the test cases that the tests run.

File: cgo/lapack.go [hunk 84]
```diff
@@ -497,6 +498,85 @@ func (impl Implementation) Dgels(trans blas.Transpose, m, n, nrhs int, a []float
...
+// The singular value decomposition is
+//  A = U * Sigma * V^T
+// where Sigma is an m×n diagonal matrix containing the singular values of A,
+// U is an m×m orthogonal matrix and V is an n×n orthogonal matrix. The first
+// min(m,n) columns of U and V are the left and right singular vectors of A
+// respectively.
+//
+// jobU and jobVT are options for computing the singular vectors. The behavior
+// is as follows
+//  jobU == lapack.SVDAll       All M columns of U are returned in u
```
vladimir-ch: s/M columns/m columns/

Also below

native/dgesvd.go
```diff
@@ -0,0 +1,1328 @@
...
+
+	bi := blas64.Implementation()
+	var mnthr int
+
+	// TODO(btracey): The netlib implementation checks have this at only length 1.
+	// Our implementation checks all input sizes before examining the l == -1 case.
+	// Fix the failing cases to reduce the needed memory here.
+	dum := make([]float64, m*n)
+
+	// Compute optimal space for subroutines.
```
vladimir-ch: Not exactly a common case but if both n and m are zero, the lapack code skips these maxwrk computations while this code does not. Is it intentional or insignificant?

native/dgesvd.go
```diff
@@ -0,0 +1,1328 @@
...
+						// Copy R from A to VT, zeroing out below it.
+						impl.Dlacpy(blas.Upper, n, n, a, lda, vt, ldvt)
+						impl.Dlaset(blas.Lower, n-1, n-1, 0, 0, vt[ldvt:], ldvt)
+
+						ie = itau
+						itauq := ie + n
+						itaup := itauq + n
+						iwork = itaup + n
+
+						// Bidiagonalize R in VT.
```
vladimir-ch: Perhaps it has been asked before but why is the rest of this if branch different from fortran?

btracey: That's odd. Especially odd that it still passes the tests. Will fix.

btracey: Done.

cgo/lapack.go

native/dgesvd.go
btracey: Added a return true if either size is zero.

native/dgesvd.go
```diff
@@ -0,0 +1,1328 @@
...
+			}
+			iwork = ie + m
+			var nru, ncvt int
+			if wantuas || wantuo {
+				nru = m
+			}
+			if wantvas || wantvo {
+				ncvt = n
+			}
+			if !wantvo && !wantvo {
```
vladimir-ch: `!wantuo`

btracey: Thanks. Done.

native/dgesvd.go
```diff
@@ -0,0 +1,1328 @@
...
+				work[i+1] = work[i+ie]
+			}
+		}
+	}
+	// Undo scaling if necessary.
+	if iscl {
+		if anrm > bignum {
+			impl.Dlascl(lapack.General, 0, 0, bignum, anrm, minmn, 1, s, minmn)
+		}
+		if ok && anrm > bignum {
```
vladimir-ch: Shouldn't be here `!ok` instead of `ok`? Same below. Fortran has `info.NE.0` and 0 indicates success.

testlapack/dgebrd.go
```diff
@@ -35,6 +35,7 @@ func DgebrdTest(t *testing.T, impl Dgebrder) {
 		{300, 400, 500},
 		{300, 400, 500},
 	} {
+
```
vladimir-ch: Delete

testlapack/dgesvd.go
```diff
@@ -0,0 +1,290 @@
...
+		svdCheckPartial(t, impl, lapack.SVDInPlace, errStr, uAllOrig, vtAllOrig, aCopy, m, n, a, lda, s, u, ldu, vt, ldvt, work, false)
+	}
+}
+
+// svdCheckPartial checks that the singular values and vectors are computed when
+// not all of them are computed.
+func svdCheckPartial(t *testing.T, impl Dgesvder, job lapack.SVDJob, errStr string, uAllOrig, vtAllOrig, aCopy []float64, m, n int, a []float64, lda int, s, u []float64, ldu int, vt []float64, ldvt int, work []float64, shortWork bool) {
+	jobU := job
+	jobVT := job
+	// Compare the singular values when computed with {None, None.}
```
vladimir-ch: `{SVDNone, SVDNone}.`

testlapack/dgesvd.go
```diff
@@ -0,0 +1,290 @@
...
+	copy(sCopy, s)
+	copy(a, aCopy)
+	for i := range s {
+		s[i] = rand.Float64()
+	}
+	tmp1 := make([]float64, 1)
+	tmp2 := make([]float64, 1)
+	jobU = lapack.SVDNone
+	jobVT = lapack.SVDNone
+	// If this isn't here the tests fail
```
vladimir-ch: Why?

testlapack/dgesvd.go
```diff
@@ -0,0 +1,290 @@
...
+
+	blas64.Gemm(blas.NoTrans, blas.NoTrans, 1, uMat, sigma, 0, tmp)
+	blas64.Gemm(blas.NoTrans, blas.NoTrans, 1, tmp, vTMat, 0, ans)
+
+	if !floats.EqualApprox(ans.Data, aCopy, 1e-8) {
+		t.Errorf("Decomposition mismatch. Trim = %v, %s", thin, errStr)
+	}
+
+	if !thin {
+		// Check that U and V are orthogonal
```
vladimir-ch: Dot

native/dgesvd.go
btracey: Yes, thank you.

testlapack/dgebrd.go
btracey: Done.

testlapack/dgesvd.go
btracey: Done.

vladimir-ch:
What a joy every time I encountered `if wantuo` or `if wantvo` in the code. LGTM. Thanks!

testlapack/dgesvd.go
btracey: Because of previous bugs that are now gone. Removed.

testlapack/dgesvd.go
btracey: Done.

btracey:
That was my thought too! If you want to figure out the correct mapping of the strides, I could be convinced to cede that job

PR #80, opened 2016-01-04
Adddlae2

btracey:


native/dlae2.go
```diff
@@ -0,0 +1,47 @@
...
+	if adf > ab {
+		rt = adf * math.Sqrt(1+(ab/adf)*(ab/adf))
+	} else if adf < ab {
+		rt = ab * math.Sqrt(1+(adf/ab)*(adf/ab))
+	} else {
+		rt = ab * math.Sqrt2
+	}
+	if sm < 0 {
+		rt1 = 0.5 * (sm - rt)
+		rt2 = (acmx/rt1)*acmn - (b/rt1)*b
```
kortschak: Can you explain the fortran comment: "Order of execution important. To get fully accurate smaller eigenvalue, next line needs to be executed in higher precision."

native/dlae2.go
```diff
@@ -0,0 +1,47 @@
...
+	}
+	var rt float64
+	if adf > ab {
+		rt = adf * math.Sqrt(1+(ab/adf)*(ab/adf))
+	} else if adf < ab {
+		rt = ab * math.Sqrt(1+(adf/ab)*(adf/ab))
+	} else {
+		rt = ab * math.Sqrt2
+	}
+	if sm < 0 {
```
kortschak: I think this set of conditions can be more nicely expressed in Go, but we'll leave it for the progressive clean-up later.

kortschak:
LGTM

native/dlae2.go
btracey: No, but I believe is is that you have to do the divisions before the multiplications. This happens automatically with Go parenthesis rules. I would imagine the next means to do it in 80-bit precision if you can. This happens automatically with the Go spec rules (as we aren't explicitly rounding with a float64())

Issue #88, opened 2016-02-02
Review all functions


btracey:
Much of the code in native sacrifices code quality in the name of correctness. Now there is a lot of redundancy in tests, and code quality can be improved with a reduced risk in error. We should take a second pass over the functions to make them idiomatic go (removing gotos, etc.)

jonlawlor:
I'll take a look at native/dbdsqr.go tonight.  Probably the most exciting thing to do is implement the sort at the end with `sort.Sort`.

kortschak:
@jonlawlor if you do that, please do benchmarking to see the impact.

jonlawlor:
Would you rather benchmarks go into the testlapack package or a new benchlapack package with identical layout to test..?

kortschak:
@btracey owns all of that.

btracey:
BLAS is a bit weird. It was my first "serious" use of the testing mechanisms in Go, and it's a weird case as it is.

It seems to me like the testlapack style works very well for testing. The only thing is that benchmarks can be autogenerated far more easily, so if you see an opportunity instead of manually generating them for each file...

jonlawlor:
I'll see if I can make a benchmark version of the testlapack style then.

PR #90, opened 2016-02-05
Benchmarks for dbdsqr

jonlawlor:
This is my first stab at benchmarks for dbdsqr.  I construct a random
bidiagonal matrix using dgebrd, and then optionally check the
performance when VT is used.  I don't have a sense for what typical
matrix sizes should be, so I picked some of the ones from the tests.
I'm also not sure how useful the additional outputs VT, etc. are, so I
might be heading in the wrong direction here.

PTAL @btracey @kortschak - this is a prerequisite for working on 
dbdsqr for #88

### DIFF
--- a/cgo/lapack_bench_test.go
+++ b/cgo/lapack_bench_test.go
@@ -1,0 +1,35 @@
+// Copyright ©2015 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package cgo
+
+import (
+	"testing"
+
+	"github.com/gonum/lapack/testlapack"
+)
+
+func BenchmarkDbdsqr100x100(b *testing.B) {
+	testlapack.DbdsqrBench(b, impl, 100, 100, 0, false)
+}
+
+func BenchmarkDbdsqr200x200(b *testing.B) {
+	testlapack.DbdsqrBench(b, impl, 200, 200, 0, false)
+}
+
+func BenchmarkDbdsqr300x300(b *testing.B) {
+	testlapack.DbdsqrBench(b, impl, 300, 300, 0, false)
+}
+
+func BenchmarkDbdsqrWithVT100x100(b *testing.B) {
+	testlapack.DbdsqrBench(b, impl, 100, 100, 0, true)
+}
+
+func BenchmarkDbdsqrWithVT200x200(b *testing.B) {
+	testlapack.DbdsqrBench(b, impl, 200, 200, 0, true)
+}
+
+func BenchmarkDbdsqrWithVT300x300(b *testing.B) {
+	testlapack.DbdsqrBench(b, impl, 300, 300, 0, true)
+}
--- a/cgo/lapack_bench_test.go
+++ b/cgo/lapack_bench_test.go
@@ -10,26 +10,32 @@ import (
 	"github.com/gonum/lapack/testlapack"
 )
 
+func BenchmarkDbdsqr10x10(b *testing.B) {
+	testlapack.DbdsqrBench(b, impl, 10, 10, 0, false)
+}
+
+func BenchmarkDbdsqr50x50(b *testing.B) {
+	testlapack.DbdsqrBench(b, impl, 50, 50, 0, false)
+}
 func BenchmarkDbdsqr100x100(b *testing.B) {
 	testlapack.DbdsqrBench(b, impl, 100, 100, 0, false)
 }
-
 func BenchmarkDbdsqr200x200(b *testing.B) {
 	testlapack.DbdsqrBench(b, impl, 200, 200, 0, false)
 }
-
-func BenchmarkDbdsqr300x300(b *testing.B) {
-	testlapack.DbdsqrBench(b, impl, 300, 300, 0, false)
+func BenchmarkDbdsqr1000x1000(b *testing.B) {
+	testlapack.DbdsqrBench(b, impl, 1000, 1000, 0, false)
 }
 
+func BenchmarkDbdsqrWithVT50x50(b *testing.B) {
+	testlapack.DbdsqrBench(b, impl, 50, 50, 0, true)
+}
 func BenchmarkDbdsqrWithVT100x100(b *testing.B) {
 	testlapack.DbdsqrBench(b, impl, 100, 100, 0, true)
 }
-
 func BenchmarkDbdsqrWithVT200x200(b *testing.B) {
 	testlapack.DbdsqrBench(b, impl, 200, 200, 0, true)
 }
-
-func BenchmarkDbdsqrWithVT300x300(b *testing.B) {
-	testlapack.DbdsqrBench(b, impl, 300, 300, 0, true)
+func BenchmarkDbdsqrWithVT1000x1000(b *testing.B) {
+	testlapack.DbdsqrBench(b, impl, 1000, 1000, 0, true)
 }
--- a/cgo/lapack_bench_test.go
+++ b/cgo/lapack_bench_test.go
@@ -17,25 +17,35 @@ func BenchmarkDbdsqr10x10(b *testing.B) {
 func BenchmarkDbdsqr50x50(b *testing.B) {
 	testlapack.DbdsqrBench(b, impl, 50, 50, 0, false)
 }
+
 func BenchmarkDbdsqr100x100(b *testing.B) {
 	testlapack.DbdsqrBench(b, impl, 100, 100, 0, false)
 }
+
 func BenchmarkDbdsqr200x200(b *testing.B) {
 	testlapack.DbdsqrBench(b, impl, 200, 200, 0, false)
 }
+
 func BenchmarkDbdsqr1000x1000(b *testing.B) {
 	testlapack.DbdsqrBench(b, impl, 1000, 1000, 0, false)
+}
+
+func BenchmarkDbdsqrWithVT10x10(b *testing.B) {
+	testlapack.DbdsqrBench(b, impl, 10, 10, 0, true)
 }
 
 func BenchmarkDbdsqrWithVT50x50(b *testing.B) {
 	testlapack.DbdsqrBench(b, impl, 50, 50, 0, true)
 }
+
 func BenchmarkDbdsqrWithVT100x100(b *testing.B) {
 	testlapack.DbdsqrBench(b, impl, 100, 100, 0, true)
 }
+
 func BenchmarkDbdsqrWithVT200x200(b *testing.B) {
 	testlapack.DbdsqrBench(b, impl, 200, 200, 0, true)
 }
+
 func BenchmarkDbdsqrWithVT1000x1000(b *testing.B) {
 	testlapack.DbdsqrBench(b, impl, 1000, 1000, 0, true)
 }
--- a/cgo/lapack_bench_test.go
+++ b/cgo/lapack_bench_test.go
@@ -1,4 +1,4 @@
-// Copyright ©2015 The gonum Authors. All rights reserved.
+// Copyright ©2016 The gonum Authors. All rights reserved.
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
--- a/testlapack/dbdsqrbench.go
+++ b/testlapack/dbdsqrbench.go
@@ -1,0 +1,77 @@
+// Copyright ©2015 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package testlapack
+
+import (
+	"math/rand"
+	"testing"
+
+	"github.com/gonum/blas"
+)
+
+type DbdsqrerDgebrder interface {
+	Dbdsqrer
+	Dgebrder
+}
+
+func DbdsqrBench(b *testing.B, impl DbdsqrerDgebrder, m, n, lda int, useVT bool) {
+	// Construct a random bidiagonal matrix by applying Dgebrd on a random matrix.
+	// The benchmarking section of netlib, found at http://www.netlib.org/lapack/lug/node71.html
+	// indicates that benchmarks are typically done on random matricies that are
+	// transformed to produce desired properties (symmetry, diagonalization, etc.).
+
+	if lda == 0 {
+		lda = n
+	}
+
+	uplo := blas.Upper
+	if m < n {
+		uplo = blas.Lower
+	}
+
+	minmn := min(m, n)
+	a := make([]float64, m*lda)
+	d := make([]float64, minmn)
+	e := make([]float64, minmn-1)
+	tauP := make([]float64, minmn)
+	tauQ := make([]float64, minmn)
+
+	work := make([]float64, 1)
+
+	impl.Dgebrd(m, n, a, lda, d, e, tauP, tauQ, work, -1)
+	work = make([]float64, int(work[0]))
+	lwork := len(work)
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		b.StopTimer()
+
+		// TODO(jonlawlor): When useVT is false, this initialization is a large
+		// portion of the benchmark time.  Because the number of iterations in a
+		// benchmark is inversely proportional to the time of a single iteration,
+		// this causes a large delay, because Dbdsqr is so fast, and the initialization
+		// time is ignored.  Maybe there is a way to amortize the cost of constructing
+		//  a random bidirectional matrix?
+		for i := range a {
+			a[i] = rand.NormFloat64()
+		}
+
+		impl.Dgebrd(m, n, a, lda, d, e, tauQ, tauP, work, lwork)
+		if useVT {
+			// Path might be used by dgesvd.
+			pt := make([]float64, n*n)
+			ldpt := n
+			for i := 0; i < n; i++ {
+				pt[i*ldpt+i] = 1
+			}
+
+			b.StartTimer()
+			impl.Dbdsqr(uplo, minmn, minmn, 0, 0, d, e, pt, minmn, nil, 0, nil, 0, work)
+			continue
+		}
+		b.StartTimer()
+		impl.Dbdsqr(uplo, minmn, 0, 0, 0, d, e, nil, 0, nil, 0, nil, 0, work)
+
+	}
+}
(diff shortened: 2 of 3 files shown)


kortschak:
Initial comments. If the randomisation is important, we need to have an explicit seed for the PRNG for consistency between runs. If it's not, the inputs can be constructed once and reused, addressing the amortisation question. It's worth seeing what @vladimir-ch thinks, since he's recently played with cheap functions with expensive setup.

Also, 2016.

cgo/lapack_bench_test.go
```diff
@@ -0,0 +1,35 @@
...
+
+package cgo
+
+import (
+	"testing"
+
+	"github.com/gonum/lapack/testlapack"
+)
+
+func BenchmarkDbdsqr100x100(b *testing.B) {
```
btracey: Rather than this being explicit, we probably want size constants in BLAS. This way we can 

I'm not convinced 200, 300, 400  is sufficient for benchmarking.I'd rather see something like  10, 50, 100, 200, 1000. One important factor in other lapack codes is the blocking, which is typically of size 32 or 64, so having sizes that branch those breaking points is good.

testlapack/dbdsqrbench.go
```diff
@@ -0,0 +1,77 @@
...
+	tauQ := make([]float64, minmn)
+
+	work := make([]float64, 1)
+
+	impl.Dgebrd(m, n, a, lda, d, e, tauP, tauQ, work, -1)
+	work = make([]float64, int(work[0]))
+	lwork := len(work)
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		b.StopTimer()
```
btracey: Can you actually start and stop the timer in the loop like this? I seem to remember in the past this gave terrible results.

testlapack/dbdsqrbench.go
```diff
@@ -0,0 +1,77 @@
...
+
+		// TODO(jonlawlor): When useVT is false, this initialization is a large
+		// portion of the benchmark time.  Because the number of iterations in a
+		// benchmark is inversely proportional to the time of a single iteration,
+		// this causes a large delay, because Dbdsqr is so fast, and the initialization
+		// time is ignored.  Maybe there is a way to amortize the cost of constructing
+		//  a random bidirectional matrix?
```
btracey: Yes, there is. Construct the random bidiagonal matrix outside the loop and make a copy of the data. Then, inside the loop, recopy it to reinitialize.

jonlawlor: My concern is that different random matrices will provide different benchmark results.  This way each iteration is a different random matrix, instead of looping over the same random matrix.  On the other hand, I have no idea if that matters!

btracey: I'm not an expert on benchmarking, but I think the same matrix is a feature. The goal of these benchmarks is to determine the affect of code changes. Random matrices just increase the variance of that measurement. If there are relevant edge cases those should be added as additional benchmarks, not included into some measure of "expected runtime".

testlapack/dbdsqrbench.go
jonlawlor: I looked through all of the gonum lapack, stat, float, blas, and matrix issues with "benchmark" in them somewhere, and I couldn't find anything discussing this issue.

It would be made moot if I use a fixed random matrix, anyway.

testlapack/dbdsqrbench.go
jonlawlor: I'm not an expert either.  I'll see if using a fixed random matrix has any effect on the benchmark output.  If not, then I'll switch to using it.

jonlawlor:
PTAL @kortschak @btracey

testlapack/dbdsqrbench.go
```diff
@@ -0,0 +1,88 @@
...
+	Dbdsqrer
+	Dgebrd(m, n int, a []float64, lda int, d, e, tauQ, tauP, work []float64, lwork int)
+}
+
+func DbdsqrBench(b *testing.B, impl DbdsqrerDgebrder, m, n, lda int, useVT bool) {
+	// Construct a random bidiagonal matrix by applying Dgebrd on a random matrix.
+	// The benchmarking section of netlib, found at http://www.netlib.org/lapack/lug/node71.html
+	// indicates that benchmarks are typically done on random matricies that are
+	// transformed to produce desired properties (symmetry, diagonalization, etc.).
+	rand.Seed(0)
```
kortschak: Use 1 since it's the default initial seed for rand.Rand.

cgo/lapack_bench_test.go
```diff
@@ -0,0 +1,41 @@
...
+
+package cgo
+
+import (
+	"testing"
+
+	"github.com/gonum/lapack/testlapack"
+)
+
+func BenchmarkDbdsqr10x10(b *testing.B) {
```
kortschak: Are these (and in native) short enough to be on one line?

kortschak:
I'm going to defer to @btracey and @vladimir-ch.

testlapack/dbdsqrbench.go
btracey: I agree that you're right. That's weird given the rest of Go.

testlapack/dbdsqrbench.go
```diff
@@ -0,0 +1,88 @@
...
+		for i := 0; i < n; i++ {
+			pt[i*n+i] = 1
+		}
+		ptCopy = make([]float64, len(pt))
+		copy(pt, ptCopy)
+	}
+
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		b.StopTimer()
```
btracey: Unless I'm doing something wrong, you can't stop and start the timer inside the benchmark. It's only used for initialization. Try running

````
package benchexample

import (
	"testing"
	"time"
)

func BenchmarkAdd(b *testing.B) {
	for i := 0; i < b.N; i++ {
		_ = 1 + 2
	}
}

func BenchmarkAddSleep(b *testing.B) {
	for i := 0; i < b.N; i++ {
		time.Sleep(1)
		_ = 1 + 2
	}
}

func BenchmarkAddSleepReset(b *testing.B) {
	for i := 0; i < b.N; i++ {
		b.StopTimer()
		time.Sleep(1)
		b.StartTimer()
		_ = 1 + 2
	}
}
````

Add is 0.3 n/op, AddSleep is 433 ns/op, and AddSleepReset has been running for 5 minutes without completing.

Comment (jonlawlor) [inline testlapack/dbdsqrbench.go]: What is happening in the last example is that it is trying to do 1/ .3 ns operations, each of which takes 1 second, which is approximately 3 x 10^9 seconds, which is about 100 years (wow, actually got to use the pi seconds in a nanocentury approximation!).

So the test is working, but you won't live to see it finish.

Comment (kortschak) [inline testlapack/dbdsqrbench.go]: You can stop/start the timer in the loop, but benchmarks take a long time. The only issue I'm concerned about is variance on the benchmark.

Comment (btracey) [inline testlapack/dbdsqrbench.go]: Ah, I see. Most of the times I've done similar code it has had this property.

Comment (btracey) [inline testlapack/dbdsqrbench.go]: The "light travels a foot in a nanosecond" approximation is pretty useful.

File: testlapack/dbdsqrbench.go (hunk 4 above)
Comment (jonlawlor) [inline testlapack/dbdsqrbench.go]: I'll change it.  Stat also uses a `rand.Seed(0)` in [weighted_test.go](https://github.com/gonum/stat/blob/c48cebcab18becaa3cfd5471dbbe4932deeacaaa/sample/weighted_test.go#L39)

File: cgo/lapack_bench_test.go (hunk 5 above)
Comment (jonlawlor) [inline cgo/lapack_bench_test.go]: `gofmt` won't let the last function be a one liner for some reason.  The spacing was just a slip on my part.

**Comment by jonlawlor** [comment]
I made the requested changes, PTAL.

**Comment by btracey** [comment]
LGTM

**Comment by kortschak** [comment]
Before you merge, would you run n=10 benchmark sets and post the results of benchstat on that.

**Comment by jonlawlor** [comment]
10x benchstat

```
benchstat bench01.txt bench02.txt bench03.txt bench04.txt bench05.txt bench06.txt bench07.txt bench08.txt bench09.txt bench10.txt
name \ time/op           bench01.txt  bench02.txt  bench03.txt  bench04.txt  bench05.txt  bench06.txt  bench07.txt  bench08.txt  bench09.txt  bench10.txt
Dbdsqr10x10-4            1.23µs ± 0%  1.26µs ± 0%  1.25µs ± 0%  1.29µs ± 0%  1.23µs ± 0%  1.31µs ± 0%  1.37µs ± 0%  1.36µs ± 0%  1.38µs ± 0%  1.41µs ± 0%
Dbdsqr50x50-4            2.46µs ± 0%  2.68µs ± 0%  2.67µs ± 0%  2.58µs ± 0%  2.80µs ± 0%  2.83µs ± 0%  2.90µs ± 0%  2.73µs ± 0%  2.75µs ± 0%  2.79µs ± 0%
Dbdsqr100x100-4          4.12µs ± 0%  3.93µs ± 0%  4.17µs ± 0%  4.07µs ± 0%  4.21µs ± 0%  4.34µs ± 0%  4.36µs ± 0%  4.29µs ± 0%  4.38µs ± 0%  4.39µs ± 0%
Dbdsqr200x200-4          7.35µs ± 0%  7.39µs ± 0%  7.35µs ± 0%  7.28µs ± 0%  7.30µs ± 0%  7.61µs ± 0%  8.16µs ± 0%  7.67µs ± 0%  7.84µs ± 0%  7.67µs ± 0%
Dbdsqr1000x1000-4        33.3µs ± 0%  35.4µs ± 0%  32.4µs ± 0%  33.2µs ± 0%  33.3µs ± 0%  35.2µs ± 0%  34.8µs ± 0%  34.9µs ± 0%  34.7µs ± 0%  35.8µs ± 0%
DbdsqrWithVT10x10-4       871ns ± 0%   901ns ± 0%   845ns ± 0%   905ns ± 0%   945ns ± 0%   956ns ± 0%  1020ns ± 0%   990ns ± 0%  1017ns ± 0%   943ns ± 0%
DbdsqrWithVT50x50-4      7.29µs ± 0%  7.85µs ± 0%  6.95µs ± 0%  6.92µs ± 0%  6.85µs ± 0%  7.36µs ± 0%  7.06µs ± 0%  7.76µs ± 0%  7.65µs ± 0%  7.43µs ± 0%
DbdsqrWithVT100x100-4    20.4µs ± 0%  20.5µs ± 0%  21.5µs ± 0%  21.0µs ± 0%  20.9µs ± 0%  22.0µs ± 0%  21.8µs ± 0%  21.9µs ± 0%  21.9µs ± 0%  21.1µs ± 0%
DbdsqrWithVT200x200-4    71.0µs ± 0%  70.2µs ± 0%  70.4µs ± 0%  72.8µs ± 0%  72.1µs ± 0%  75.3µs ± 0%  75.1µs ± 0%  74.0µs ± 0%  74.7µs ± 0%  76.0µs ± 0%
DbdsqrWithVT1000x1000-4  1.74ms ± 0%  1.74ms ± 0%  1.77ms ± 0%  1.76ms ± 0%  1.77ms ± 0%  1.89ms ± 0%  1.90ms ± 0%  1.92ms ± 0%  1.92ms ± 0%  1.76ms ± 0%
```

Looks like the output varies by as much as 20%.

**Comment by kortschak** [comment]
If you cat all the files together it will do the stats for you.

**Comment by jonlawlor** [comment]
So it does!

```
name                     time/op
Dbdsqr10x10-4            1.31µs ± 8%
Dbdsqr50x50-4            2.72µs ± 9%
Dbdsqr100x100-4          4.23µs ± 7%
Dbdsqr200x200-4          7.56µs ± 8%
Dbdsqr1000x1000-4        34.3µs ± 6%
DbdsqrWithVT10x10-4       939ns ±10%
DbdsqrWithVT50x50-4      7.31µs ± 7%
DbdsqrWithVT100x100-4    21.3µs ± 4%
DbdsqrWithVT200x200-4    73.2µs ± 4%
DbdsqrWithVT1000x1000-4  1.82ms ± 6%
```

File: testlapack/dbdsqrbench.go [hunk 7]
```diff
@@ -0,0 +1,88 @@
...
+	impl.Dgebrd(m, n, a, lda, d, e, tauP, tauQ, work, -1)
+
+	work = make([]float64, int(work[0]))
+	lwork := len(work)
+
+	// Construct the bidiagonal matrix in d and e.
+	impl.Dgebrd(m, n, a, lda, d, e, tauQ, tauP, work, lwork)
+	dCopy := make([]float64, len(d))
+	copy(d, dCopy)
+	eCopy := make([]float64, len(e))
```
vladimir-ch: Is `eCopy` necessary? From the documentation to Dbdsqr it seems it is not overwritten.

testlapack/dbdsqrbench.go
```diff
@@ -0,0 +1,88 @@
...
+			pt[i*n+i] = 1
+		}
+		ptCopy = make([]float64, len(pt))
+		copy(pt, ptCopy)
+	}
+
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		b.StopTimer()
+		copy(dCopy, d)
```
vladimir-ch: Shouldn't the copy be done in the opposite direction?

testlapack/dbdsqrbench.go
```diff
@@ -0,0 +1,88 @@
...
+		copy(eCopy, e)
+
+		if useVT {
+			// Path might be used by dgesvd.
+			copy(ptCopy, pt)
```
vladimir-ch: Same here?

testlapack/dbdsqrbench.go
```diff
@@ -0,0 +1,88 @@
...
+			b.StartTimer()
+			impl.Dbdsqr(uplo, minmn, minmn, 0, 0, d, e, pt, minmn, nil, 0, nil, 0, work)
+			continue
```
vladimir-ch: I would prefer the StopTimer() call at line 74 to move to line 71, and then add it here and at line 87. In this way the code being measured would be explicitly enclosed between `StartTimer` and `StopTimer`.

testlapack/dbdsqrbench.go
```diff
@@ -0,0 +1,88 @@
...
+	var ptCopy []float64
+
+	if useVT {
+		// Path might be used by dgesvd.
+		pt = make([]float64, n*n)
+		for i := 0; i < n; i++ {
+			pt[i*n+i] = 1
+		}
+		ptCopy = make([]float64, len(pt))
+		copy(pt, ptCopy)
```
vladimir-ch: `copy(ptCopy, pt)`

testlapack/dbdsqrbench.go
```diff
@@ -0,0 +1,88 @@
...
+
+	impl.Dgebrd(m, n, a, lda, d, e, tauP, tauQ, work, -1)
+
+	work = make([]float64, int(work[0]))
+	lwork := len(work)
+
+	// Construct the bidiagonal matrix in d and e.
+	impl.Dgebrd(m, n, a, lda, d, e, tauQ, tauP, work, lwork)
+	dCopy := make([]float64, len(d))
+	copy(d, dCopy)
```
vladimir-ch: `copy(dCopy, d)`

testlapack/dbdsqrbench.go
```diff
@@ -0,0 +1,88 @@
...
+
+	"github.com/gonum/blas"
+)
+
+type DbdsqrerDgebrder interface {
+	Dbdsqrer
+	Dgebrd(m, n int, a []float64, lda int, d, e, tauQ, tauP, work []float64, lwork int)
+}
+
+func DbdsqrBench(b *testing.B, impl DbdsqrerDgebrder, m, n, lda int, useVT bool) {
```
vladimir-ch: Is it necessary to take all three `m`, `n`, `lda`? `Dbdsqr` works only on square matrices.

cgo/lapack_bench_test.go
```diff
@@ -0,0 +1,51 @@
...
+
+func BenchmarkDbdsqrWithVT100x100(b *testing.B) {
+	testlapack.DbdsqrBench(b, impl, 100, 100, 0, true)
+}
+
+func BenchmarkDbdsqrWithVT200x200(b *testing.B) {
+	testlapack.DbdsqrBench(b, impl, 200, 200, 0, true)
+}
+
+func BenchmarkDbdsqrWithVT1000x1000(b *testing.B) {
```
vladimir-ch: If the `xCols` part of the function name is removed, `gofmt` will leave it on one line. I think it is nice to have all these functions as one-liners with no empty lines in between. Just my taste.

testlapack/dbdsqrbench.go
```diff
@@ -0,0 +1,88 @@
...
+	// Construct a random bidiagonal matrix by applying Dgebrd on a random matrix.
+	// The benchmarking section of netlib, found at http://www.netlib.org/lapack/lug/node71.html
+	// indicates that benchmarks are typically done on random matricies that are
+	// transformed to produce desired properties (symmetry, diagonalization, etc.).
```
vladimir-ch: Why is it done like this? In order to get a regular matrix? Wouldn't for this benchmark be enough just to generate two random slices?

testlapack/dbdsqrbench.go
```diff
@@ -0,0 +1,88 @@
...
+	e := make([]float64, minmn-1)
+	tauP := make([]float64, minmn)
+	tauQ := make([]float64, minmn)
+
+	work := make([]float64, 1)
+
+	impl.Dgebrd(m, n, a, lda, d, e, tauP, tauQ, work, -1)
+
+	work = make([]float64, int(work[0]))
+	lwork := len(work)
```
vladimir-ch: `lwork` is used only once, so delete it?

testlapack/dbdsqrbench.go
```diff
@@ -0,0 +1,88 @@
...
+package testlapack
+
+import (
+	"math/rand"
+	"testing"
+
+	"github.com/gonum/blas"
+)
+
+type DbdsqrerDgebrder interface {
```
vladimir-ch: If `Dgebrd` (and therefore this interface) is really necessary, then call it `DbdsqrDgebrder`?

btracey: Maybe just DbdsqrBencher

vladimir-ch: Yes, that's better.

vladimir-ch:

PR #100, opened 2016-02-12
Add Dorgtr and test

btracey:


btracey:
I'd be happy to modify the Dsytrd tests if anyone understands why the Dorgtr test hits the bug but the Dsytrd tests do not. It's not even really specifically a bug with the Dsytrd test... copying and pasting the failing case into Dsytrd made the test fail on that specific case.

native/dorgtr.go
```diff
@@ -0,0 +1,92 @@
+// Copyright ©2016 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package native
+
+import "github.com/gonum/blas"
+
+// Dorgtr generates a real orthogonal matrix Q which is defines as the product
```
vladimir-ch: s/defines/defined/

native/dorgtr.go
```diff
@@ -0,0 +1,92 @@
...
+// of n-1 elementary reflectors of order N as returned by Dsytrd.
+//
+// If uplo == blas.Upper
+//  Q = H[n-1] * ... * H[1] * H[0]
+// If uplo == blas.Lower
+//  L = H[0] * ... * H[n-1]
+//
```
vladimir-ch: Could you delete this line and connect the ifs with where to read as a single sentence?

native/dorgtr.go
```diff
@@ -0,0 +1,92 @@
...
+// Where H[i] is constructed from the elementary reflectors as computed by Dsytrd.
+// Please see the documentation for Dsytrd for more information.
```
vladimir-ch: Without `Please`?

native/dorgtr.go
```diff
@@ -0,0 +1,92 @@
+// Copyright ©2016 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package native
+
+import "github.com/gonum/blas"
+
+// Dorgtr generates a real orthogonal matrix Q which is defines as the product
+// of n-1 elementary reflectors of order N as returned by Dsytrd.
```
vladimir-ch: Is `n` and `N` the same number?

native/dorgtr.go
```diff
@@ -0,0 +1,92 @@
...
+	if len(tau) < n-1 {
+		panic(badTau)
+	}
+	upper := uplo == blas.Upper
+	var nb int
+	if upper {
+		nb = impl.Ilaenv(1, "DORGQL", " ", n-1, n-1, n-1, -1)
+	} else {
+		nb = impl.Ilaenv(1, "DORGQR", " ", n-1, n-1, n-1, -1)
+		nb = 4
```
vladimir-ch: Why do you overwrite `nb` with 4?

native/dorgtr.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+	lworkopt := max(1, n-1) * nb
+	work[0] = float64(lworkopt)
+	if lwork == -1 {
+		return
+	}
+	if len(work) < lwork {
```
vladimir-ch: Swap this if with the next one? Let's first check that `lwork` has a good value and then check that work has a good length.

kortschak:
>I'd be happy to modify the Dsytrd tests if anyone understands why the Dorgtr test hits the bug but the Dsytrd tests do not. It's not even really specifically a bug with the Dsytrd test... copying and pasting the failing case into Dsytrd made the test fail on that specific case.

What do dsytrd.f and dorgtr.f do with the failing case?

This dependence on PRNG sequence is troubling - we should make all the tests truly independent (this means each test uses its own rand.Rand since tests may be run in parallel - so merely doing rand.Seed(1) at the beginning is not adequate). Probably having a test flag that specifies the number of test interations would be a worthwhile addition too.

(I thought I had made this comment on this PR over the weekend, but obviously not).

native/dorgtr.go
```diff
@@ -0,0 +1,92 @@
...
+//  Q = H[n-1] * ... * H[1] * H[0]
+// If uplo == blas.Lower
+//  L = H[0] * ... * H[n-1]
+//
+// Where H[i] is constructed from the elementary reflectors as computed by Dsytrd.
+// Please see the documentation for Dsytrd for more information.
+//
+// tau must have length at least n-1, and Dorgtr will panic otherwise.
+//
+// Work is temporary storage, and lwork specifies the usable memory length. At minimum,
```
kortschak: I think we need to come up with a decision about case for tau, work etc.

native/dorgtr.go
```diff
@@ -0,0 +1,92 @@
...
+	if n == 0 {
+		return
+	}
+	if upper {
+		// Q was determined by a call to Dsytrd with uplo == blas.Upper.
+		// Shift the vectors which define the elementary reflectors one column
+		// to the left, and set the last row and column of Q to those of the unit
+		// matrix.
+		for j := 0; j < n-1; j++ {
+			for i := 0; i <= j-1; i++ {
```
kortschak: `i < j`?

native/dorgtr.go
```diff
@@ -0,0 +1,92 @@
...
+		a[(n-1)*lda+n-1] = 1
+
+		// Generate Q[0:n-1, 0:n-1].
+		impl.Dorgql(n-1, n-1, n-1, a, lda, tau, work, lwork)
+	} else {
+		// Q was determined by a call to Dsytrd with uplo == blas.Upper.
+		// Shift the vectors which define the elementary reflectors one column
+		// to the right, and set the first row and column of Q to those of the unit
+		// matrix.
+		for j := n - 1; j >= 1; j-- {
```
kortschak: `j > 0`?

native/dsytrd.go
```diff
@@ -135,7 +135,7 @@ func (impl Implementation) Dsytrd(uplo blas.Uplo, n int, a []float64, lda int, d
 			// Update the unreduced submatrix A[i+ib:n, i+ib:n], using an update
 			// of the form A = A + V*W^T - W*V^T.
 			bi.Dsyr2k(uplo, blas.NoTrans, n-i-nb, nb, -1, a[(i+nb)*lda+i:], lda,
-				work[nb+i:], ldwork, 1, a[(i+nb)*lda+i+nb:], lda)
+				work[nb*ldwork:], ldwork, 1, a[(i+nb)*lda+i+nb:], lda)
```
kortschak: Comment on this fix in the commit message.

kortschak:
LGTM after comments addressed.

btracey:
In this particular case it's not a prng thing, at least not in the sense of the other tests that fail when the random numbers are different. In this case the code was wrong and discovered in the first "real" input (as determined by dsytrd). There's something about the real input which triggers the bug that my direct tests do not

native/dorgtr.go
btracey: Done.

native/dorgtr.go
btracey: Done.

native/dorgtr.go
btracey: Sorry. Done.

native/dorgtr.go
btracey: yep.

native/dorgtr.go
btracey: Hard to find bug. Removed.

native/dorgtr.go
btracey: I don't think that's the right order. It seems much more likely that someone would accidentally input a bad value for lwork (0 or something). Given the len() command in go there is no reason for the input caller to have a lesser value, so if that condition is true something weird is happening and it should be alerted first.

native/dorgtr.go
btracey: Made lowercase now like in the recent batch of functions.

native/dorgtr.go
btracey: Done.

native/dorgtr.go
btracey: Done.

kortschak:
The dsytrd change commit comment was missed.

btracey:
I don't understand how. I edited the commit message when I rebased and also
in the github api

kortschak:
It's in the merge commit, but it should be in the actual commit. During the rebase this would have happened (or not).

kortschak:
Do you know how the fortran equivalents behave with the failing case?

btracey:
Not specifically, but I'd imagine they would similarly fail. I think it may be the same issue that I had when trying to get Dgesvd to work. That is, Go is fine with slicing beyond the end of the matrix, and so if there is an indexing error plus data originally being part of the same array there is no panic. This relates to your valid issue that we really should be using full slicing on the slices and not just pointer movement.

PR #101, opened 2016-02-13
add set -ex and comments to install scripts

jonlawlor:
Added set -ex and some comments to install scripts

Same as https://github.com/gonum/blas/pull/156


jonlawlor:
PTAL

jonlawlor:
ping @vladimir-ch

vladimir-ch:
LGTM

PR #102, opened 2016-02-15
testlapack: make tests independent

kortschak:
@btracey Please take a look.

The other thing to do is make it possible to perform multiple iterations of tests from a cli flag. I had a look at that, but there are a few different cases, some already do this, but others don't.


btracey:
LGTM

kortschak:
Added dorgtr.

PR #104, opened 2016-02-16
{cgo,native}: fix matrix typography

### LINKED ISSUE #97: doc: harmonise use of matrix bracket characters
In some places '(' and ')' are used, while in others '[' and ']' are used. We should be consistent, but a choice needs to be made first.

kortschak:
Use square brackets throughout and use an (almost) uniform matrix cell layout. All matrix cells are right justified and cells are the width of the largest element (negative signs treated according to taste here). Large matrices have 2 space gutter between elements and 2x2 matrices or smaller have 1 space gutter.

Fixes #97.

@btracey Please take a look.


sbinet:
LGTM

btracey:
LGTM

Issue #35, opened 2015-08-14
native,clapack: mark functions as internal when not part of lapack.Float64 or lapack.Complex128


kortschak:
All of clapack can be noted as not really for public consumption (though not in flux - essentially the same as the std syscall package).

Methods in native are present in many cases because they are dependencies for routines that are published in lapack.Float64. We can either unexport them in the short term (probably the correct answer for things like Ilaenv that have been noted as targets for removal) or just document them as internal.

kortschak:
Fixed by #113.

PR #113, opened 2016-03-15
native: mark internal routines

kortschak:
@btracey Please take a look.

### DIFF
--- a/native/dgels.go
+++ b/native/dgels.go
@@ -33,7 +33,7 @@ import (
 // leading submatrix of b contains the solution vectors X. If trans == blas.NoTrans,
 // this submatrix is of size n×nrhs, and of size m×nrhs otherwise.
 //
-// Work is temporary storage, and lwork specifies the usable memory length.
+// work is temporary storage, and lwork specifies the usable memory length.
 // At minimum, lwork >= max(m,n) + max(m,n,nrhs), and this function will panic
 // otherwise. A longer work will enable blocked algorithms to be called.
 // In the special case that lwork == -1, work[0] will be set to the optimal working
--- a/native/dbdsqr.go
+++ b/native/dbdsqr.go
@@ -47,6 +47,8 @@ import (
 // will panic if there is insufficient working memory.
 //
 // Dbdsqr returns whether the decomposition was successful.
+//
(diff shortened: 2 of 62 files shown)


kortschak:
Fixes #35.

btracey:
Is the policy that a function is internal by default? For example, I could see mat64 wanting to use dlatrs. If we decided that's true, we would just remove the internal comment?

kortschak:
Yes, and we would add Dlatrs to the lapack64 interface.

I intially went through and started to make these unexported, but realised (late) that then testing would break.

kortschak:
*ping*

btracey:
clapack is internal by default probably.

If we really want the functions to be uncallable, we could have a wrapper type that lives in native_test.go and exports them. It's not clear we want this. On the one hand, it's nice to allow direct access for advanced users, but on the other hand I don't know that we want to promise API stability for them. Maybe stability only extends to lapack64? We already can't promise to keep clapack stable.

native/dgels.go
```diff
@@ -33,11 +33,13 @@ import (
...
 // this submatrix is of size n×nrhs, and of size m×nrhs otherwise.
 //
-// Work is temporary storage, and lwork specifies the usable memory length.
+// work is temporary storage, and lwork specifies the usable memory length.
 // At minimum, lwork >= max(m,n) + max(m,n,nrhs), and this function will panic
 // otherwise. A longer work will enable blocked algorithms to be called.
 // In the special case that lwork == -1, work[0] will be set to the optimal working
 // length.
+//
+// Dgels is an internal routine. It is exported for testing purposes.
```
btracey: It's part of lapack.Float64, so I don't think this is internal. We don't actually use it anymore in mat64, so we could remove it if we would like.

btracey:
LGTM with Dgels fixed.

native/dgels.go
kortschak: I missed that it was there. We need to make a decision about whether to keep it before the API promise.

kortschak:
>Maybe stability only extends to lapack64?

Possibly.

>We already can't promise to keep clapack stable.

Yes.

PR #114, opened 2016-03-21
lapack: add stability promise

kortschak:
I think we can add this now.

Are there _any_ security perspective that we need to consider? Also see note in https://github.com/gonum/blas/pull/171 about layout conversion.

@gonum/developers Please take a look.


vladimir-ch:

PR #115, opened 2016-03-23
Change install script to use prebuilt xenial .debs

jonlawlor:
This circumvents the typical ubuntu apt-get and installs blas and lapack libraries from ubuntu xenial packages.  Followup to gonum/blas#170.

### DIFF
--- a/.travis/linux/OpenBLAS/install.sh
+++ b/.travis/linux/OpenBLAS/install.sh
@@ -1,33 +1,40 @@
 set -ex
 
+# remove exiting blas packages to prevent conflicts
+sudo apt-get remove libblas3gf libatlas3gf-base
+
 # fetch fortran to build OpenBLAS
 sudo apt-get update -qq
 sudo apt-get install -qq gfortran
 
-# fetch OpenBLAS
-pushd ~
-sudo git clone --depth=1 git://github.com/xianyi/OpenBLAS
+# source for blas and lapack packages
+LAPACK_ROOT="http://mirrors.xmission.com/ubuntu/pool/main/l/lapack"
+OPENBLAS_ROOT="http://mirrors.xmission.com/ubuntu/pool/universe/o/openblas"
 
-# make OpenBLAS
-pushd OpenBLAS
-echo OpenBLAS $(git rev-parse HEAD)
-sudo make FC=gfortran &> /dev/null
-sudo make PREFIX=/usr install
-popd
+packagelist=(
+  "$LAPACK_ROOT/libblas-common_3.6.0-2ubuntu2_amd64.deb"
+  "$LAPACK_ROOT/libblas3_3.6.0-2ubuntu2_amd64.deb"
+  "$LAPACK_ROOT/libblas-dev_3.6.0-2ubuntu2_amd64.deb"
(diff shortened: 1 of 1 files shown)


jonlawlor:
New build succeeds in a bit less than 5 minutes, PTAL @kortschak

.travis/linux/OpenBLAS/install.sh
```diff
@@ -1,33 +1,40 @@
 set -ex
 
+# remove exiting blas packages to prevent conflicts
+sudo apt-get remove libblas3gf libatlas3gf-base
+
 # fetch fortran to build OpenBLAS
 sudo apt-get update -qq
 sudo apt-get install -qq gfortran
```
kortschak: Is this still needed?

kortschak:
This is quite fragile.

.travis/linux/OpenBLAS/install.sh
jonlawlor: Yes, libblas3gf in Ubuntu 12.04 has a dependency on libblas-common, but the dependency tree is different than in Ubuntu Xenial, so it results in a circular dependency.

jonlawlor:
Yes, it is definitely fragile.  If we want faster builds then we need prebuilt packages (I think, although there may be some `make` options in OpenBLAS that can reduce compilation time...).  Travis does not allow manipulation of sources.list, so we have to download and install packages outside of apt-get.  It may be possible to get the packages from somewhere else, and we could make it a bit better by having multiple hosts that can provide the .deb files, and we can pretty it up a bit, but it will never be as stable as apt-get, and we lose the benefit of compiling against the latest OpenBLAS to see if they have any regressions.  It is up to you guys if the tradeoff is worth it.

kortschak:
I don't think it is worth it, but wait for other opinions.

jonlawlor:
Well, this pull was spurred on by a comment you made about how long the builds take, so if you aren't in favor of it then I'll close without merging.  To be clear: I will not be annoyed at all if it is rejected - I didn't know at the start that the way to get pre-built packages for OpenBLAS and LAPACKE working in travis would be to hack around in this manner.

On the plus side, it looks like apt-get would "just work" if we could use Ubuntu Xenial, but that version of Ubuntu is *not even released yet*.  It is likely that if we could add backports to the sources.list file we could get apt-get working in prior versions of Ubuntu, which may be a useful addition to any installation notes.

It may still be possible to reduce compilation time a bit using the current method.  

The *other* possibility is to use a different continuous integration provider; preferably one which has good docker integration.  Then we could build a base environment including OpenBLAS and LAPACKE into the docker container, go get gonum/..., and run tests.

kortschak:
>Well, this pull was spurred on by a comment you made about how long the builds take, so if you aren't in favor of it then I'll close without merging.

Sorry.

>The other possibility is to use a different continuous integration provider; preferably one which has good docker integration. Then we could build a base environment including OpenBLAS and LAPACKE into the docker container, go get gonum/..., and run tests.

Yeah, I had a look at caching in travis and it's only supported for private repos on travis-ci.com.

jonlawlor:
There isn't anything to be sorry about! I thought it was too fragile as well but depending on your appetite it may have been worth the tradeoff. It may still be doable with a different ci provider.

PR #116, opened 2016-04-09
Change Travis CI to use Docker for OpenBLAS

jonlawlor:
This change uses a docker image to test OpenBLAS instead of building
the library during every test.

This involves more network data transfer but it results in tests that
are ~1.5 minutes to run instead of 6 minutes. It should be possible to
use a smaller docker image in principle - currently the docker image is
based on ubuntu, and we could use alpine linux instead. However, I was
unsuccessful at getting OpenBLAS to build there. It would probably be
useful outside of CI testing to have a minimal docker image that can
perform efficient matrix algebra in go.

I’ve also restructured the testing scripts to use an install.sh and
test.sh for each of the blas libraries.

Currently the docker image resides at
hub.docker.com/jonlawlor/gonum-docker-openblas but before merging I
would like to move the image into a gonum repo.


jonlawlor:
Above it should say that tests take 3.5 minutes instead of 10 minutes.  Otherwise I think this is ready, subject to moving the docker image as I mentioned in gonum/blas#174.  PTAL @kortschak @btracey

jonlawlor:
Closed in favor of #156

PR #120, opened 2016-04-14
cgo,native: fix Dormqr work checks

### LINKED ISSUE #117: cgo: Dormqr comment not in sync with reality
The comment says:
```
// ... lwork == -1 will not run Dormqr but will instead write the minimum
// work necessary to work[0]. If len(work) < lwork, Dormqr will panic.
```
The code does:
```
        if lwork == -1 {
                 if left {
                         work[0] = float64(m)
                         return
                 }
                 work[0] = float64(n)
                 return
         }
 
         if left {
                 if lwork < n {
                         panic(badWork)
                 }
         } else {
                 if lwork < m {
                         panic(badWork)
                 }
         }
```
That is,
1. Values put into work[0] and checked values of lwork seem to be swapped.
2. len(work) < lwork is not checked.

### LINKED ISSUE #119: native: check whether len(work) >= lwork in Dormqr

kortschak:
Fixes #117.
Fixes #119.

### DIFF
--- a/cgo/lapack.go
+++ b/cgo/lapack.go
@@ -953,10 +953,10 @@ func (impl Implementation) Dormqr(side blas.Side, trans blas.Transpose, m, n, k 
 
 	if lwork == -1 {
 		if left {
-			work[0] = float64(m)
+			work[0] = float64(n)
 			return
 		}
-		work[0] = float64(n)
+		work[0] = float64(m)
 		return
 	}
 	if len(work) < lwork {
--- a/cgo/lapack.go
+++ b/cgo/lapack.go
@@ -959,7 +959,9 @@ func (impl Implementation) Dormqr(side blas.Side, trans blas.Transpose, m, n, k 
 		work[0] = float64(n)
 		return
 	}
-
+	if len(work) < lwork {
+		panic(badWork)
+	}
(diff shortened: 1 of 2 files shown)


cgo/lapack.go
```diff
@@ -951,13 +951,15 @@ func (impl Implementation) Dormqr(side blas.Side, trans blas.Transpose, m, n, k
...
-			work[0] = float64(m)
+			work[0] = float64(n)
 			return
 		}
-		work[0] = float64(n)
+		work[0] = float64(m)
 		return
 	}
-
+	if len(work) != lwork {
```
vladimir-ch: (At least at the moment) the rest of the code (and the doc comment) use `if len(work) < lwork`.

kortschak: Ugh. See comment about reading in other thread.

Done.

vladimir-ch:
LGTM but perhaps we should wait for @btracey ?

kortschak:
Absolutely. The whole row versus column thing gives me the heebie geebies.

btracey:
LGTM

It's not so bad once you've implemented float64-BLAS and a few lapack functions (except for the times when it still is)

PR #121, opened 2016-04-15
Fix, clean up, and unify documentation of elementary reflectors

vladimir-ch:
* Use zero-based indexing for elementary reflectors (almost) everywhere.
* Denote elementary reflectors as H(i), not H_i, not H[i].
* Remove some lint.

### DIFF
--- a/cgo/lapack.go
+++ b/cgo/lapack.go
@@ -354,7 +354,7 @@ func (impl Implementation) Dgecon(norm lapack.MatrixNorm, n int, a []float64, ld
 //
 // See Dgeqr2 for a description of the elementary reflectors and orthonormal
 // matrix Q. Q is constructed as a product of these elementary reflectors,
-//  Q = H(k-1) ... H(1) * H(0),
+//  Q = H(k-1) * ... * H(1) * H(0),
 // where k = min(m,n).
 //
 // Work is temporary storage of length at least m and this function will panic otherwise.
@@ -415,7 +415,7 @@ func (impl Implementation) Dgelqf(m, n int, a []float64, lda int, tau, work []fl
 // and computing H(i) = I - tau[i] * v * v^T.
 //
 // The orthonormal matrix Q can be constucted from a product of these elementary
-// reflectors, Q = H(0) * H(1) ... H(k-1), where k = min(m,n).
+// reflectors, Q = H(0) * H(1) * ... * H(k-1), where k = min(m,n).
 //
 // Work is temporary storage of length at least n and this function will panic otherwise.
 func (impl Implementation) Dgeqr2(m, n int, a []float64, lda int, tau, work []float64) {
@@ -739,7 +739,7 @@ func (impl Implementation) Dorgbr(vect lapack.DecompUpdate, m, n, k int, a []flo
 
 // Dorglq generates an m×n matrix Q with orthonormal rows defined by the product
 // of elementary reflectors
-//  Q = H(k-1) ... H(1) * H(0)
+//  Q = H(k-1) * ... * H(1) * H(0)
 // as computed by Dgelqf. Dorglq is the blocked version of Dorgl2 that makes
 // greater use of level-3 BLAS routines.
 //
@@ -781,7 +781,7 @@ func (impl Implementation) Dorglq(m, n, k int, a []float64, lda int, tau, work [
 
 // Dorgqr generates an m×n matrix Q with orthonormal columns defined by the
 // product of elementary reflectors
-//  Q = H(0) * H(1) ... H(k-1)
+//  Q = H(0) * H(1) * ... * H(k-1)
 // as computed by Dgeqrf. Dorgqr is the blocked version of Dorg2r that makes
 // greater use of level-3 BLAS routines.
 //
--- a/cgo/lapack.go
+++ b/cgo/lapack.go
@@ -254,13 +254,13 @@ func (impl Implementation) Dbdsqr(uplo blas.Uplo, n, ncvt, nru, ncc int, d, e, v
 //
 // The remaining elements of A store the data needed to construct Q and P.
 // The matrices Q and P are products of elementary reflectors
-//  if m >= n, Q = H(0) * H(1) * ... * H(n-1),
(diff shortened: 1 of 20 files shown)


cgo/lapack.go
```diff
@@ -410,12 +410,12 @@ func (impl Implementation) Dgelqf(m, n int, a []float64, lda int, tau, work []fl
 // The ith elementary reflector can be explicitly constructed by first extracting
 // the
 //  v[j] = 0           j < i
-//  v[j] = i           j == i
-//  v[j] = a[i*lda+j]  j > i
-// and computing h_i = I - tau[i] * v * v^T.
+//  v[j] = 1           j == i
+//  v[j] = a[j*lda+i]  j > i
```
vladimir-ch: I think that v is stored in the columns of a but please check this @btracey

lapack64/lapack64.go
```diff
@@ -99,19 +99,19 @@ func Gels(trans blas.Transpose, a blas64.General, b blas64.General, work []float
...
 // this function will panic otherwise.
 //
 // The ith elementary reflector can be explicitly constructed by first extracting
 // the
 //  v[j] = 0           j < i
-//  v[j] = i           j == i
-//  v[j] = a[i*lda+j]  j > i
-// and computing h_i = I - tau[i] * v * v^T.
+//  v[j] = 1           j == i
+//  v[j] = a[j*lda+i]  j > i
```
vladimir-ch: Also here.

native/dgeqr2.go
```diff
@@ -20,12 +20,12 @@ import "github.com/gonum/blas"
 // The ith elementary reflector can be explicitly constructed by first extracting
 // the
 //  v[j] = 0           j < i
-//  v[j] = i           j == i
-//  v[j] = a[i*lda+j]  j > i
-// and computing h_i = I - tau[i] * v * v^T.
+//  v[j] = 1           j == i
+//  v[j] = a[j*lda+i]  j > i
```
vladimir-ch: And here.

native/dlatrd.go
```diff
@@ -53,11 +53,11 @@ import (
 // H has the form
 //  I - tau * v * v^T
 // If uplo == blas.Upper,
-//  Q = H[n] * H[n-1] * ... * H[n-nb+1]
+//  Q = H(n) * H(n-1) * ... * H(n-nb+1)
```
vladimir-ch: This seems to be still 1-indexed but (at least at the moment) my head refuses to convert it to 0-based confidently, especially together with the description of v.

kortschak:
What is the rationale behind the use of parens versus brackets? Do we have one?

btracey:
I think it should be brackets. This implies the go style and not the fortran one. It's not strictly like the go one since they are columns, but it's a notional documentation anyway.

vladimir-ch:
I have no strong preference, my intention is just to reduce inconsistencies in the documentation.
I would go with the Latex-like notation H_k but that becomes less clear when the index is an expression and using {} makes it look comlicated.
In the end I went with parens because brackets evoke slices and H are neither slices nor vectors. I kept the usage of brackets for things that have their direct equivalent in the code, for example the tau slice. I see parens here not as a Fortran notation, rather as a function notation: H is a function that returns elementary reflector.
But yes, it's a notational documentation with limited means of expression, a bit of a nitpicking, so I'm happy to change it to whatever you think will work best.

kortschak:
If you say they're functions, then I'm happy with that, but I'll leave the decision to you and Brendan.

vladimir-ch:
Looking at `H` as a matrix function of `i` is one possible way to justify using `()` over `[]`. Another justification could be reserving `[]` for indexing matrices, vectors, and slices, which is not the case 
of `H(i)`. Using `H_1` and `H_{n-i+1}` here would probably be most appropriate but I don't know how people not familiar with Latex perceive this notation.

To sum it up, what I'm trying to say: given that we have just ASCII and a limited subset of UTF8 symbols at hand to document a numerical library, having some basic set of conventions applied consistently would help having a clean and confidence-conveying documentation.

btracey:
You're right, the underscore-brace system is probably better.

cgo/lapack.go
btracey: Yes, columns.

lapack.go
```diff
@@ -48,8 +48,8 @@ type Float64 interface {
 type Direct byte
 
 const (
-	Forward  Direct = 'F' // Reflectors are right-multiplied, H_1 * H_2 * ... * H_k
-	Backward Direct = 'B' // Reflectors are left-multiplied, H_k * ... * H_2 * H_1
+	Forward  Direct = 'F' // Reflectors are right-multiplied, H(0) * H(1) * ... * H(k).
```
btracey: I think needs to be k-1

lapack64/lapack64.go
btracey: Yep

native/dlatrd.go
btracey: It does seem to be one indexed. See line 78 of testlapack/dlatrd.

native/dlatrd.go
```diff
@@ -53,11 +53,11 @@ import (
...
 // where v[:i-1] is stored in A[:i-1,i], v[i-1] = 1, and v[i:n] = 0.
 //
 // If uplo == blas.Lower,
-//  Q = H[1] * H[2] * ... H[nb]
+//  Q = H(1) * H(2) * ... H(nb)
```
btracey: This also seems one-indexed

cgo/lapack.go
btracey: yep

btracey:
LGTM, with the small corrections.

native/dlatrd.go
vladimir-ch: The pointer to the test was very useful. Thanks. Done.

lapack.go
vladimir-ch: Done, although here without context the final (or initial) subscript does not really matter.

native/dlatrd.go
vladimir-ch: Done.

PR #126, opened 2016-04-25
native: add Dlaqr1

vladimir-ch:
Needed by `Dlaqr5`. PTAL

### DIFF
--- a/native/dlaqr1.go
+++ b/native/dlaqr1.go
@@ -1,0 +1,57 @@
+// Copyright ©2016 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package native
+
+import "math"
+
+// Dlaqr1 sets v to a scalar multiple of the first column of the product
+//  (H - (sr1 + i*si1)*I)*(H - (sr2 + i*si2)*I)
+// where H is a 2×2 or 3×3 matrix, I is the identity matrix of the same size,
+// and i is the imaginary unit. Scaling is done to avoid overflows and most
+// underflows.
+//
+// n is the order of H and must be either 2 or 3. It must hold that either sr1 =
+// sr2 and si1 = -si2, or si1 = si2 = 0. The length of v must be equal to n. If
+// any of these conditions is not met, Dlaqr1 will panic.
+//
+// Dlaqr1 is an internal routine. It is exported for testing purposes.
+func (impl Implementation) Dlaqr1(n int, h []float64, ldh int, sr1, si1, sr2, si2 float64, v []float64) {
+	if n != 2 && n != 3 {
+		panic(badDims)
+	}
+	checkMatrix(n, n, h, ldh)
+	if len(v) != n {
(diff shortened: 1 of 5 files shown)


native/dlaqr1.go
```diff
@@ -0,0 +1,56 @@
...
+// license that can be found in the LICENSE file.
+
+package native
+
+import "math"
+
+// Dlaqr1 sets v to a scalar multiple of the first column of the product
+//  (H - (sr1 + i*si1)*I)*(H - (sr2 + i*si2)*I)
+// where H is a 2×2 or 3×3 matrix and i is the imaginary unit. Scaling is done
+// to avoid overflows and most underflows.
```
btracey: Could you add that I is the identity matrix? Kind of obvious, but somewhat tricky with all the i/I

native/general.go
```diff
@@ -32,6 +32,7 @@ const (
 	badNorm         = "lapack: bad norm"
 	badPivot        = "lapack: bad pivot"
 	badS            = "lapack: s has insufficient length"
+	badShifts       = "lapack: bad shifts"
```
btracey: This needs to be added to the cgo package as well.

btracey:
LGTM

native/dlaqr1.go
vladimir-ch: Done.

native/general.go
vladimir-ch: Done.

vladimir-ch:
I greatly simplified the test for this. The changes are big enough for another PTAL.

testlapack/dlaqr1.go
```diff
@@ -0,0 +1,91 @@
...
+)
+
+type Dlaqr1er interface {
+	Dlaqr1(n int, h []float64, ldh int, sr1, si1, sr2, si2 float64, v []float64)
+}
+
+func Dlaqr1Test(t *testing.T, impl Dlaqr1er) {
+	rnd := rand.New(rand.NewSource(1))
+
+	for _, n := range []int{2, 3} {
```
kortschak: Curious about the use of a range statement here and for cas. I can see pros and cons, but wondering about the motivation.

vladimir-ch: I just wanted to have a single loop body for all the four possible cases. If you want, I can factor the body out into `dlaqr1Test(t *testing.T, impl Dlaqr1er, n, cas int)` and call it here 4 times with fixed arguments.

kortschak: I was wondering why it was spelled `for _, n := range []int{2, 3} {` instead of `for n := 2; n <= 3; n++ {`.

Now that I write out the alternative. I am happy what you have here was the correct choice.

vladimir-ch: Ah, sorry, I misunderstood what you had meant. I use the range statement where I iterate over a fixed set of options, even if they are consecutive numbers.

kortschak: Agreed - definitely better.

PR #129, opened 2016-05-11
native: add Dlaqr5

vladimir-ch:
Dlaqr5 is a tough start for a lapack-porting beginner. Please review with great care.

For testing I went the way of comparing to generated data in `internal/testdata/dlaqr5test`. I use tip's cgo ability to deal with Fortran files, that's very convenient. I store one test case per text file and zip them together. I thought about generating a go file but it would be naturally bigger and I wanted to avoid that.

I don't like much having duplicate copies of Lapack source files in the repository but I guess that's a future work.

I tried to split the PR into narrowly focused commits so that some may be easily skipped (those that add generated data and fortran files).

Oh, and search me if there are any blatant mistakes.

PTAL


testlapack/dlaqr5.go
```diff
@@ -0,0 +1,151 @@
...
+					n, ktop, kbot,
+					nshfts, sr, si,
+					h, ldh,
+					0, n-1, z, ldh,
+					v, 3,
+					u, 3*nshfts-3,
+					nh, wh, nh,
+					nv, wv, 3*nshfts-3)
+
+				if !equalApprox(n, h, ldh, hwant, 1e-14) {
```
vladimir-ch: The results are not bitwise equal with fortran. For bigger matrices (starting at around n=20) the tolerance must be increased to 1e-13. I don't know whether it indicates a bug or not.

native/dlaqr5.go
```diff
@@ -0,0 +1,609 @@
...
+// v and ldv represent an auxiliary matrix of size (nshfts/2)×3.
+//
+// u and ldv represent an auxiliary matrix of size (3*nshfts-3)×(3*nshfts-3).
+//
+// wh and ldwh represent an auxiliary matrix of size (3*nshfts-3)×nh.
+//
+// wv and ldwv represent an auxiliary matrix of size nv×(3*nshfts-3).
+func (impl Implementation) Dlaqr5(wantt, wantz bool, kacc22 int, n, ktop, kbot, nshfts int, sr, si []float64, h []float64, ldh int, iloz, ihiz int, z []float64, ldz int, v []float64, ldv int, u []float64, ldu int, nh int, wh []float64, ldwh int, nv int, wv []float64, ldwv int) {
+	checkMatrix(n, n, h, ldh)
+	if !(0 <= ktop && ktop < n) {
```
kortschak: `if ktop < 0 || n <= ktop {`

Expresses, "ktop is outside the rage [0,n)" very nicely. Also below.

native/dlaqr5.go
```diff
@@ -0,0 +1,609 @@
...
+	if ktop > 0 && h[ktop*ldh+ktop-1] != 0 {
+		panic("lapack: diagonal block is not isolated")
+	}
+	if !(0 <= kbot && kbot < n) {
+		panic("lapack: invalid value of kbot")
+	}
+	if kbot < n-1 && h[(kbot+1)*ldh+kbot] != 0 {
+		panic("lapack: diagonal block is not isolated")
+	}
+	if nshfts < 0 || nshfts%2 != 0 {
```
kortschak: Use `nshfts&0x1 != 0` rather than `%2`? The compiler should optimise them to the same code, but I'm not sure it does.

native/dlaqr5.go
```diff
@@ -0,0 +1,609 @@
...
+		panic("lapack: invalid number of shifts")
+	}
+	if len(sr) != nshfts || len(si) != nshfts {
+		panic(badSlice) // TODO(vladimir-ch) Another message?
+	}
+	if wantz {
+		if !(0 <= iloz && iloz <= ihiz && ihiz < n) {
```
kortschak: `iloz < 0 || ihiz < iloz || ihiz >= n`

native/dlaqr5.go
```diff
@@ -0,0 +1,609 @@
...
+			panic("lapack: invalid value of iloz or ihiz")
```
kortschak: We can split the condition above and differentiate the two. This would clarify the message and make the condition nicer.

```
if ihiz >= n {
    panic("lapack: invalid value of ihiz")
}
if iloz <= 0 || ihiz < iloz {
    panic("lapack: invalid value of iloz")
}
```

native/dlaqr5.go
```diff
@@ -0,0 +1,609 @@
...
+	}
+
+	// Shuffle shifts into pairs of real shifts and pairs of complex
+	// conjugate shifts assuming complex conjugate shifts are already
+	// adjacent to one another.
+	for i := 0; i < nshfts-2; i += 2 {
+		if si[i] == -si[i+1] {
+			continue
+		}
+		swap := sr[i]
```
kortschak: Replace these with swap idiom? Or mark with TODO?

```
sr[i], sr[i+1], sr[i+2] = s[i+1], sr[i+2], sr[i]
sr[i], sr[i+1], sr[i+2] = s[i+1], sr[i+2], sr[i]
```

native/dlaqr5.go
```diff
@@ -0,0 +1,609 @@
...
+		si[i+1] = si[i+2]
+		si[i+2] = swap
+	}
+
+	// Note: lapack says that nshfts must be even but allows it to be odd
+	// anyway. At the moment, we panic above if nshfts is not even, so
+	// reducing it by one is unnecessary (it seems that Dlaqr0 and Dlaqr4
+	// indeed use only even nshfts).
+	//
+	// * NSHFTS is supposed to be even, but if it is odd,
```
kortschak: Be a little more explicit that this is the original comment and code in LAPACK.

native/dlaqr5.go
```diff
@@ -0,0 +1,609 @@
...
+			bmp22 := (mbot < nbmps-1) && (krcol+3*m22 == kbot-2)
+
+			// Generate reflections to chase the chain right one
+			// column. (The minimum value of k is ktop-1.)
+			for m := mtop; m <= mbot; m++ {
+				k := krcol + 3*m
+				if k == ktop-1 {
+					impl.Dlaqr1(3, h[ktop*ldh+ktop:], ldh,
+						sr[2*m], si[2*m], sr[2*m+1], si[2*m+1],
+						v[m*ldv:m*ldv+3])
```
kortschak: Nice.

native/dlaqr5.go
```diff
@@ -0,0 +1,609 @@
...
+					alpha := v[m*ldv]
+					_, v[m*ldv] = impl.Dlarfg(3, alpha, v[m*ldv+1:m*ldv+3], 1)
```
kortschak: Why is v sliced from 1 more here and in the next Dlarfg call than in the Dlaqr1 call above? They have the same index in the Fortran.

native/dlaqr5.go
```diff
@@ -0,0 +1,609 @@
...
+				// resulting from the new reflector is too large,
+				// then abandon it. Otherwise, use the new one.
+				var vt [3]float64
+				impl.Dlaqr1(3, h[(k+1)*ldh+k+1:], ldh, sr[2*m],
+					si[2*m], sr[2*m+1], si[2*m+1], vt[:])
+				alpha := vt[0]
+				_, vt[0] = impl.Dlarfg(3, alpha, vt[1:3], 1)
+				refsum := vt[0] * (h[(k+1)*ldh+k] + vt[1]*h[(k+2)*ldh+k])
+
+				tmp := math.Abs(h[k*ldh+k]) + math.Abs(h[(k+1)*ldh+k+1]) + math.Abs(h[(k+2)*ldh+k+2])
```
kortschak: Since this is a var here, can it get a better name than `tmp`?

native/dlaqr5.go
```diff
@@ -0,0 +1,609 @@
...
+				if math.Abs(h[(k+2)*ldh+k]-refsum*vt[1])+math.Abs(refsum*vt[2]) > ulp*tmp {
+					// Starting a new bulge here would create
+					// non-negligible fill. Use the old one with
+					// trepidation.
+					h[(k+1)*ldh+k] = beta
+					h[(k+2)*ldh+k] = 0
+					h[(k+3)*ldh+k] = 0
+				} else {
```
kortschak: Up to you, but you're in a loop here, so you could continue rather than else. The equivalent indentation and the alliteration of the comment sort of makes sense the way it is.

native/dlaqr5.go
```diff
@@ -0,0 +1,609 @@
...
+			case wantt:
+				jbot = n - 1
+			default:
+				jbot = kbot
+			}
+			for j := max(ktop, krcol); j <= jbot; j++ {
+				mend := min(mbot+1, (j-krcol+2)/3) - 1
+				for m := mtop; m <= mend; m++ {
+					k := krcol + 3*m
+					refsum := v[m*ldv] * (h[(k+1)*ldh+j] + v[m*ldv+1]*h[(k+2)*ldh+j] + v[m*ldv+2]*h[(k+3)*ldh+j])
```
kortschak: These long lines could be split (also below).

native/dlaqr5.go
```diff
@@ -0,0 +1,609 @@
...
+					refsum := v[m*ldv] * (h[j*ldh+k+1] + v[m*ldv+1]*h[j*ldh+k+2] + v[m*ldv+2]*h[j*ldh+k+3])
+					h[j*ldh+k+1] -= refsum
+					h[j*ldh+k+2] -= refsum * v[m*ldv+1]
+					h[j*ldh+k+3] -= refsum * v[m*ldv+2]
+				}
+				if accum {
+					// Accumulate U. (If necessary, update Z later with with an
+					// efficient matrix-matrix multiply.)
+					kms := k - incol
+					for j := max(1, ktop-incol) - 1; j < kdu; j++ {
```
kortschak: `max(0, ktop-incol-1)`?

kortschak: `max(0, ktop-incol-1)`

(second second thoughts)

Also below.

kortschak:
Got to L388 in the first commit.

native/dlaqr5.go
vladimir-ch: I'm on tip. `if nshfts < 0 || nshfts%2 != 0` produces
```
MOVQ    "".nshfts+912(FP), R8
TESTQ   R8, R8
JLT     $0, 22283
MOVQ    R8, R9
SARQ    $63, R8
MOVQ    R9, R10
SUBQ    R8, R9
MOVQ    R9, "".autotmp_1905+696(SP)
MOVQ    R9, R11
ANDQ    $1, R9
ADDQ    R9, R8
TESTQ   R8, R8
JNE     $0, 22283
```
while `if nshfts < 0 || nshfts&0x1 != 0` produces
```
MOVQ    "".nshfts+912(FP), R8
TESTQ   R8, R8
JLT     $0, 22269
TESTQ   $1, R8
JNE     $0, 22269
```
This check is not a performance-critical part of the code but the difference is surprising. I will change it to `&0x1`.

native/dlaqr5.go
vladimir-ch: I hope I haven't made a mistake but it seems that both versions produce exactly the same assembly. I'll go with the swap idiom.

native/dlaqr5.go
vladimir-ch: That's just Fortran.
`V( 1, M )` passed to `Dlaqr1` is an n-vector (n=3 here) and the result is stored in-place.
`V( 1, M )` passed to `Dlarfg` is just a reference to that single element that is modified in place (the second returned value `tau` in our Go version). `V( 2, M )` is an n-1-vector that is modified in-place together with alpha (the first returned value `beta` in our Go version).

native/dlaqr5.go
vladimir-ch: Renamed to `dsum`.

native/dlaqr5.go
vladimir-ch: I used `continue` in other place but here I left it as is for the exact reasons you state. I tried how it looks with continue and I prefer it this way.

vladimir-ch:
I'd rebase to replace the first commit with the updated one so that you can continue reviewing on per-commit basis but I'm afraid that your comments would get lost. Do you know if that still happens @kortschak ?

testlapack/dlaqr5.go
vladimir-ch: A possible explanation might be different implementations of Dgemm.

native/dlaqr5.go
kortschak: They should generate the same code.

kortschak:
I'm pretty confident that the per commit comments now are kept in the conversation - the links in the emails don't work, but they stay here.

native/dlaqr5.go
```diff
@@ -0,0 +1,609 @@
...
+				// improves accuracy in some examples. Falling back on an alternate
+				// convergence criterion when tst1 or tst2 is zero (as done here) is
+				// traditional but probably unnecessary.
+
+				if h[(k+1)*ldh+k] == 0 {
+					continue
+				}
+				tst1 := math.Abs(h[k*ldh+k]) + math.Abs(h[(k+1)*ldh+k+1])
+				if tst1 == 0 {
+					if k >= ktop+1 {
```
kortschak: Express this as a switch.

native/dlaqr5.go
```diff
@@ -0,0 +1,609 @@
...
+					}
+				}
+				if math.Abs(h[(k+1)*ldh+k]) <= math.Max(smlnum, ulp*tst1) {
+					h12 := math.Max(math.Abs(h[(k+1)*ldh+k]), math.Abs(h[k*ldh+k+1]))
+					h21 := math.Min(math.Abs(h[(k+1)*ldh+k]), math.Abs(h[k*ldh+k+1]))
+					h11 := math.Max(math.Abs(h[(k+1)*ldh+k+1]), math.Abs(h[k*ldh+k]-h[(k+1)*ldh+k+1]))
+					h22 := math.Min(math.Abs(h[(k+1)*ldh+k+1]), math.Abs(h[k*ldh+k]-h[(k+1)*ldh+k+1]))
+					scl := h11 + h12
+					tst2 := h22 * (h11 / scl)
+					if tst2 == 0 || h21*(h12/scl) <= math.Max(smlnum, ulp*tst2) {
```
kortschak: The corresponding Fortran for this is a mess.

native/dlaqr5.go
```diff
@@ -0,0 +1,609 @@
...
+		var jtop, jbot int
+		if wantt {
+			jtop = 0
+			jbot = n - 1
+		} else {
+			jtop = ktop
+			jbot = kbot
+		}
+		bi := blas64.Implementation()
+		if !blk22 || incol < ktop || ndcol > kbot || ns <= 2 {
```
kortschak: `!blk22 || kbot < ndcol || incol < ktop || ns <= 2` unless there's reason to believe the performance will be significantly hit.

native/dlaqr5.go
```diff
@@ -0,0 +1,609 @@
...
+			// Updates not exploiting the 2×2 block structure of U. k1 and nu keep track
+			// of the location and size of U in the special cases of introducing bulges
+			// and chasing bulges off the bottom. In these special cases and in case the
+			// number of shifts is ns = 2, there is no 2×2 block structure to exploit.
+
+			k1 := max(1, ktop-incol)
```
kortschak: Does it make sense to make zero-based and rename it `k0`? There's a lot of `k1-1` going on below.

native/dlaqr5.go
```diff
@@ -0,0 +1,609 @@
...
+			nu := kdu - max(0, ndcol-kbot) - k1 + 1
+
+			// Horizontal multiply.
+			for jcol := min(ndcol, kbot) + 1; jcol <= jbot; jcol += nh {
+				jlen := min(nh, jbot-jcol+1)
+				bi.Dgemm(blas.ConjTrans, blas.NoTrans, nu, jlen, nu,
```
kortschak: I know it's conjtrans in the Fortran, but why?

native/dlaqr5.go
```diff
@@ -0,0 +1,609 @@
...
+			// rows get multiplied by zero).
+			impl.Dlacpy(blas.All, knz, jlen, h[(incol+1+j2)*ldh+jcol:], ldh, wh[kzs*ldwh:], ldwh)
+
+			// Multiply by U21^T.
+			impl.Dlaset(blas.All, kzs, jlen, 0, 0, wh, ldwh)
+			bi.Dtrmm(blas.Left, blas.Upper, blas.ConjTrans, blas.NonUnit, knz, jlen,
+				1, u[j2*ldu+kzs:], ldu, wh[kzs*ldwh:], ldwh)
+
+			// Multiply top of H by U11^T.
+			bi.Dgemm(blas.ConjTrans, blas.NoTrans, i2, jlen, j2,
```
kortschak: Same question.

native/dlaqr5.go
```diff
@@ -0,0 +1,609 @@
...
+		for jcol := min(ndcol, kbot) + 1; jcol <= jbot; jcol += nh {
+			jlen := min(nh, jbot-jcol+1)
+
+			// Copy bottom of H to top+kzs of scratch (the first kzs
+			// rows get multiplied by zero).
+			impl.Dlacpy(blas.All, knz, jlen, h[(incol+1+j2)*ldh+jcol:], ldh, wh[kzs*ldwh:], ldwh)
+
+			// Multiply by U21^T.
+			impl.Dlaset(blas.All, kzs, jlen, 0, 0, wh, ldwh)
+			bi.Dtrmm(blas.Left, blas.Upper, blas.ConjTrans, blas.NonUnit, knz, jlen,
```
kortschak: And here.

native/dlaqr5.go
```diff
@@ -0,0 +1,609 @@
...
+				1, wh, ldwh)
+
+			// Copy top of H to bottom of WH.
+			impl.Dlacpy(blas.All, j2, jlen, h[(incol+1)*ldh+jcol:], ldh, wh[i2*ldwh:], ldwh)
+
+			// Multiply by U21^T.
+			bi.Dtrmm(blas.Left, blas.Lower, blas.ConjTrans, blas.NonUnit, j2, jlen,
+				1, u[i2:], ldu, wh[i2*ldwh:], ldwh)
+
+			// Multiply by U22 ====
```
kortschak: s/ ====/./

native/dlaqr5.go
```diff
@@ -0,0 +1,609 @@
...
+			bi.Dgemm(blas.NoTrans, blas.NoTrans, jlen, i4-i2, j4-j2,
+				1, h[jrow*ldh+incol+1+j2:], ldh, u[j2*ldu+i2:], ldu,
+				1, wv[i2:], ldwv)
+
+			// Copy it back.
+			impl.Dlacpy(blas.All, jlen, kdu, wv, ldwv, h[jrow*ldh+incol+1:], ldh)
+		}
+
+		// Multiply Z (also vertical).
+		if wantz {
```
kortschak:
```
if !wantz {
    continue
}
```

and required brace deletion below.

kortschak:
Ignoring second commit (reference Fortran files).

kortschak:
Third commit looks OK (test case generation). Ignoring zip file commit.

internal/testdata/dlaqr5test/main.go
```diff
@@ -0,0 +1,207 @@
+// Copyright ©2016 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package main
```
kortschak: Package level comment to explain what this does. Also comment on how you are handling row/col major between the two given that you're calling Fortran, but using the results in Go for the tests.

native/dlaqr5.go
vladimir-ch: I don't see how to convert this to a switch easily. All the if conditions can be true.

native/dlaqr5.go
vladimir-ch: I don't think this will affect performance. I changed to `incol < ktop || kbot < ndcol`, the order from lower to higher indices seems more natural.

native/dlaqr5.go
vladimir-ch: There is only one `k1-1` more than `k1` but I changed to `k0` to simply stick with 0-based indices.

native/dlaqr5.go
vladimir-ch: I don't know why Fortran uses conjtrans, I just followed Fortran because Dgemm handles both as equivalent. Now I changed all ConjTrans to just Trans.

native/dlaqr5.go
vladimir-ch: Sorry. Done.

internal/testdata/dlaqr5test/main.go
```diff
@@ -0,0 +1,216 @@
...
+	const tmpl = "wantt%v_n%v_nshfts%v_ktop%v_kbot%v.txt"
+
+	file, err := os.Create("dlaqr5data.zip")
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer file.Close()
+	zipfile := zip.NewWriter(file)
+
+	rnd := rand.New(rand.NewSource(1))
```
vladimir-ch: I added this common random generator to get higher variability in generated matrices. Indeed, it increases coverage by taking the branch at dlaqr5.go:416. It also required the equality tolerance to be increased to 1e-13.

internal/testdata/dlaqr5test/main.go
```diff
@@ -0,0 +1,216 @@
...
+
+	for _, wantt := range []bool{true, false} {
```
vladimir-ch: Using false for wantt also increases coverage slightly by triggering a couple of if conditions. It also increases the size of the test file but it's better to be safe.

native/dlaqr5.go
kortschak: Ah, sorry.

internal/testdata/dlaqr5test/main.go
kortschak: Nice work.

kortschak:
LGTM but wait for @btracey.

native/dlaqr5.go
```diff
@@ -0,0 +1,616 @@
...
+//  1: Dlaqr5 will accumulate reflections and use matrix-matrix multiply to
+//     update far-from-diagonal matrix entries.
+//  2: Dlaqr5 will accumulate reflections, use matrix-matrix multiply to update
+//     far-from-diagonal matrix entries, and take advantage of 2×2 block
+//     structure during matrix multiplies.
+// For other values of kacc2 Dlaqr5 will panic.
+//
+// n is the order of the Hessenberg matrix H.
+//
+// ktop and kbot are indices of the first and last row and column of an isolated
```
btracey: It might be easier in the long run if this is go indexing. That is [ktop:kbot], so kbot is as if it's one indexed. Usually the lapack routines specify a start and a length, so I'm not exactly sure.

native/dlaqr5.go
```diff
@@ -0,0 +1,616 @@
...
+		for krcol := incol; krcol <= min(incol+3*nbmps-3, kbot-2); krcol++ {
+			// Bulges number mtop to mbot are active double implicit
+			// shift bulges. There may or may not also be small 2×2
+			// bulge, if there is room. The inactive bulges (if any)
+			// must wait until the active bulges have moved down the
+			// diagonal to make room. The phantom matrix paradigm
+			// described above helps keep track.
+
+			mtop := max(0, ((ktop-1)-krcol+2)/3)
+			mbot := min(nbmps, (kbot-krcol)/3) - 1
```
btracey: If kbot is 1-indexed, this should change too.

native/dlaqr5.go
```diff
@@ -0,0 +1,616 @@
...
+//
+// iloz and ihiz specify the rows of Z to which transformations will be applied
+// if wantz is true. It must hold that 0 <= iloz <= ihiz < n, otherwise Dlaqr5
+// will panic.
+//
+// z and ldz represent the matrix Z of size n×n. If wantz is true, the QR sweep
+// orthogonal similarity transformation is accumulated into
+// z[iloz:ihiz,ilo:ihi] from the right, otherwise z not referenced.
+//
+// v and ldv represent an auxiliary matrix of size (nshfts/2)×3.
```
btracey: There should be a more clear note that V is transposed. Everywhere else we have kept the same data orientation.

Similarly, why did you decide to transpose it? You'll have to be clever in future callers of it.

btracey:
Main code LGTM looks good fol Dlaqr5. The only comment is about v being transposed, but it's your call.

native/dlaqr5.go
vladimir-ch: I transposed it because of Dlaqr1. In fortran it takes columns of V but without a stride which is assumed to be 1. So I transpose and pass rows to Dlaqr1.
I could use the array `vt` to pass to Dlaqr1 everywhere and then copy its elements out as on L245 but it would be slightly less efficient. The question is whether it's worth it. Dlaqr5 is a very internal and specific routine that is called only from Dlaqr0 and Dlaqr4. They both pass a work slice for v and in fact at the call site nothing will change, ldv is 3 both in Fortran and here.
I will make the transpose more explicit in the comment but I think it's ok not to do anything else.

native/dlaqr5.go
vladimir-ch: I thought about this too but in the end decided that keeping kbot inclusive is the least change both in meaning of the parameters and in the actual implementation. The same with mbot. I will keep it as is but feel free to bug me if you oppose.

PR #130, opened 2016-05-16
native: add Dlahr2

vladimir-ch:
This one is much easier than Dlaqr5. Dlahr2 is from a different branch of the dependency tree, it goes toward implementing Dgehrd, a routine for reducing general matrices to upper Hessenberg form.

Here I take the testing data approach from Dlaqr5 a bit further by storing the data in JSON format. I think that the little added markup is very useful, and writing and reading of the data becomes much simpler.

Also, I moved the necessary reference Fortran files to a new package netlib under internal, a step open to discussion. I agree with @kortschak that this could be later moved to blas so that both can benefit from it. It requires tip at the moment but this is an internal land, so it's ok.

This a lot of changes, but Dlahr2 is quite simple and I wanted to test the combination of netlib package, json format and related changes in the tests in real life. I think it worked out quite nicely, so if it is accepted, in the next PR I'll apply the same changes to Dlaqr5 (and one day maybe also to Dlasq and Dsterf).

PTAL


kortschak:
>I agree with @kortschak that this could be later moved to blas so that both can benefit from it.

I was just looking for where I wrote that and can't find it, can you post the link? (I suspect that we may be speaking about different things).

(I really need to adjust my internal cache invalidation policy).

vladimir-ch:
Private communication:
> I think a package that keeps the Fortran reference implantation and generates a set of golden tests sounds pretty reasonable.
The Fortran interface package I intended to include in the lapack repo but now think should go in blas would probably help though.

This doesn't exclude the possibility that we may be speaking about different things.

kortschak:
Yes, I thought that might be the case. That was two comments - the Fortran interface package was the thing that needs to be visible to blas. It's currently a languishing PR in lapack.

I agree though with your plan.

native/dlahr2.go
```diff
@@ -0,0 +1,143 @@
...
+//
+// tau will contain the scalar factors of the elementary reflectors. It must
+// have length at least nb.
+//
+// t represents the nb×nb upper triangular matrix T and y represents the n×nb
+// matrix Y.
+//
+// Dlahr2 is an internal routine. It is exported for testing purposes.
+func (impl Implementation) Dlahr2(n, k, nb int, a []float64, lda int, tau, t []float64, ldt int, y []float64, ldy int) {
+	checkMatrix(n, n-k+1, a, lda)
```
vladimir-ch: There are no safety checks in the netlib implementation, so I at least added these obvious four. However, not every combination of `n`, `k`, and `nb` is ok, but I'm not sure how strict we are with these very internal routines (it's called only by DGEHRD).

btracey:
Sorry, I don't understand why these fortran files would go into BLAS. Are we talking about functions like RowMajor? It would be nice to have a set of routines that are meant to be efficient, though we may need a different version for testing. We could use those to implement wrappers aroud cuBLAS

internal/testdata/dlahr2test/main.go
```diff
@@ -0,0 +1,100 @@
...
+	A        []float64
+
+	AWant   []float64
+	TWant   []float64
+	YWant   []float64
+	TauWant []float64
+}
+
+func main() {
+	file, err := os.Create("../../../testlapack/testdata/dlahr2data.json.gz")
```
btracey: Is this platform independent (with the "..")?

internal/testdata/dlahr2test/main.go
```diff
@@ -0,0 +1,100 @@
...
+
+	var tests []Dlahr2Test
+	for _, n := range []int{4, 5, 6, 7, 11} {
+		for k := 0; k <= n/2; k++ {
+			for nb := 1; nb <= k; nb++ {
+				ain := genrand(n, n-k+1, rnd)
+				a := make([]float64, len(ain))
+				copy(a, ain)
+
+				t := make([]float64, nb*nb)
```
btracey: I think in general it's better to fill these with random values as well as a general check.

internal/testdata/dlahr2test/main.go
kortschak: Two dots is, assuming "/" is not. Use `filepath.Join("..", "..", "..", "testlapack/testdata/dlahr2data.json.gz")` if we care.

kortschak:
I don't think the Fortran code here should be visible to blas. The code I'm talking about making visible to blas is the row/col major conversion code in #69. I'm pretty convinced that should move to blas and result in method addition to the types there. The thing that's delaying that is the implementation of packed type conversion and finding a naming system that's not revolting.

internal/testdata/dlahr2test/main.go
vladimir-ch: Used `filepath.FromSlash`. Done.

internal/testdata/dlahr2test/main.go
vladimir-ch: I added filling with random values but I don't know what you mean by the general check.

vladimir-ch:
> I don't think the Fortran code here should be visible to blas.

Agreed, I misunderstood one of your previous comments.

PTAL

internal/testdata/dlahr2test/main.go
btracey: Sorry, poorly phrased sentence. "It's a good general check to fill them with random numbers instead of zeros".

internal/testdata/dlahr2test/main.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+	json.NewEncoder(w).Encode(tests)
+
+	err = w.Close()
+	if err != nil {
+		log.Fatal(err)
+	}
+}
+
+// genrand returns a general r×c matrix with random entries.
```
btracey: There should be tests where stride != minimum necessary, I've found a lot of bugs with those tests.

testlapack/dlahr2.go
```diff
@@ -0,0 +1,83 @@
...
+	N, K, NB int
+	A        []float64
+
+	AWant   []float64
+	TWant   []float64
+	YWant   []float64
+	TauWant []float64
+}
+
+func Dlahr2Test(t *testing.T, impl Dlahr2er) {
```
btracey: It's fine to test against the Lapack implementation, but the primary test should be of the function behavior itself. Using the Fortran code directly in Dlasq was because I did not understand the purpose of the routine, and so could not write an independent test for it. The behavior of this function seems clear, and so there should be a direct test of function behavior. There is no sense in deleting the test you already have, but perhaps give the zip file test a different name.

testlapack/general.go
```diff
@@ -594,3 +594,55 @@ func isOrthonormal(q blas64.General) bool {
...
 }
+
+// copyMatrix copies an m×n matrix src of stride n into an m×n matrix dst of stride ld.
+func copyMatrix(m, n int, dst []float64, ld int, src []float64) {
+	for i := 0; i < m; i++ {
+		copy(dst[i*ld:i*ld+n], src[i*n:i*n+n])
+	}
+}
+
+// equalApprox returns whether the matrices A and B of order n are approximately
```
btracey: In general, we want to test not just that the matrix values are correct, but also that the off-matrix values haven't been modified either. We have found bugs in routines that didn't have these tests, for example a range statement that overwrites all values to zero, not just the in-matrix values. This is why I suggest having slices with random numbers. This fills the whole matrix with meaningful values and checks that only the correct ones have been modified.

native/dlahr2.go
btracey: If there are things that must be true, it's a good idea to add those checks. Maybe we would remove them when we know the routines are correct, but it's really good for testing purposes to make sure we've understood the function properly (for example, your implementation of Dgehrd may go against what you thought you knew was true).

btracey:
Actual implementation looks correct.

kortschak:
>Here I take the testing data approach from Dlaqr5 a bit further by storing the data in JSON format. I think that the little added markup is very useful, and writing and reading of the data becomes much simpler.

When we (you) have some experience with the various approaches it would be worth harmonising all the tests so there isn't a hodgepodge of approaches to wade through.

internal/testdata/dlahr2test/main.go
vladimir-ch: This is only for generating input data for the fortran routines. In the actual test I do create matrices with stride != minimum necessary.

testlapack/dlahr2.go
vladimir-ch: This is exactly what I was thinking yesterday when I started working on Dgehrd. I'll take a look at it  and see what I can do.

internal/testdata/dlahr2test/main.go
```diff
@@ -0,0 +1,101 @@
+// Copyright ©2016 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// This program generates test data for Dlahr2. Test cases are stored in
```
kortschak: s/This program/dlahr2test/

native/dlahr2.go
```diff
@@ -0,0 +1,143 @@
...
+
+	bi := blas64.Implementation()
+	var ei float64
+	for i := 0; i < nb; i++ {
+		if i > 0 {
+			// Update A[k:n,i].
+
+			// Update i-th column of A - Y * V^T.
+			bi.Dgemv(blas.NoTrans, n-k, i,
+				-1, y[k*ldy:], ldy, a[(k+i-1)*lda:], 1, 1, a[k*lda+i:], lda)
```
kortschak: Can we split these so slices and lds are more clearly grouped (also below).

vladimir-ch:
> When we (you) have some experience with the various approaches it would be worth harmonising all the tests so there isn't a hodgepodge of approaches to wade through.

Also this is something I've been thinking about recently. I'm still in the learning/exploring phase, I guess when I gain confidence I'll have a clearer idea. Using Fortran reference data is in some sense easy and comfortable but might create trouble in the future if the reference implementation changes. It is a desperately big amount of work.

testlapack/general.go
kortschak: Systematic rather than random - I used to do random, but I was convinced by your arguments, and have seen the benefits in my own debugging sessions.

kortschak:
Dlahr2 LGTM with minor comments, but the testing issues raised by @btracey need to be addressed.

>Using Fortran reference data is in some sense easy and comfortable but might create trouble in the future if the reference implementation changes.

Yes, we've seen enough errors in the backing cgo implementations (admittedly the C shims, but...) to not feel easy just testing against them.

>It is a desperately big amount of work.

I'm happy if an issue is filed to cover this work later.

vladimir-ch:
Ok, I added tests that don't use the reference data. I fill outside with NaNs and inside with random values. PTAL @kortschak @btracey

btracey:
LGTM.

In the future, you may find it easier to debug if you set the values to fixed unique values rather than NaNs, but this is good.

vladimir-ch:
Another reason for not using NaNs in matrices is the high-level LAPACKE which by default contains a check for NaN values. The middle-level interface does not contain such a check. Another benefit of the middle level interface is that work parameter becomes available. Was there a discussion in gonum about using high- or middle-level interface to LAPACK?

PR #131, opened 2016-05-31
testlapack: update test for Dlaqr5

vladimir-ch:
Follow the test for Dlahr2:
- Move Fortran files to netlib directory
- Test data in JSON format
- Add test that does not use reference data. There are not many known properties that can be tested, only the orthogonality of z but please let me know if you see anything else.


vladimir-ch:
I did find one more check: test if `Z^T * HOrig * Z` is (almost) equal to H from `Dlaqr5`. Together with the comparison to reference data I think Dlaqr5 is tested pretty thoroughly. PTAL

btracey:
I believe that testDlaqr5 can also set the wantX values to false, and then test that the output matches. This would be a nice test to add.

Otherwise LGTM (if you want to make that the next PR)

vladimir-ch:
Thanks, I'll do it in a future PR.

PR #134, opened 2016-06-02
native,cgo: add Dgehrd

vladimir-ch:


native/dgehrd.go
```diff
@@ -0,0 +1,178 @@
...
+//  [     v1  h   h   h   h   h ]
+//  [     v1  v2  h   h   h   h ]
+//  [     v1  v2  v3  h   h   h ]
+//  [                         a ]
+// where a denotes an element of the original matrix A, h denotes a
+// modified element of the upper Hessenberg matrix H, and vi denotes an
+// element of the vector defining H_i.
+//
+// ilo and ihi determine the block of A that will be reduced to upper Hessenberg
+// form. It must hold that 0 <= ilo <= ihi <= max(0, n-1), otherwise Dgehrd will
```
vladimir-ch: Here is the place with `ilo` and `ihi` when `n == 0` I was refering to before, @kortschak .

testlapack/dgehrd.go
```diff
@@ -0,0 +1,204 @@
...
+type Dgehrder interface {
+	Dgehrd(n, ilo, ihi int, a []float64, lda int, tau, work []float64, lwork int)
+}
+
+func DgehrdTest(t *testing.T, impl Dgehrder) {
+	rnd := rand.New(rand.NewSource(1))
+
+	// Randomized tests for small matrix sizes that will most likely
+	// use the unblocked algorithm.
+	for _, n := range []int{2, 3, 4, 5, 10, 34} {
```
vladimir-ch: I start with `n` at 2. For `n == 1` the cgo test (using OpenBLAS) panics with `panic: runtime error: index out of range [recovered]` and I don't know why. Even 'n == 0' should be ok according to the documentation.

kortschak: Can you write a non-testing program to reproduce this so we can see where the panic happens?

vladimir-ch: Got it. It happens in clapack.go:648. If n == 1, `tau` has zero length but we take the address of the first element.

kortschak: Is there anything I need to do there?

vladimir-ch: Not sure. Fortran says that tau is an array of size n-1 but also that n>=0.
It does not promise that tau is not referenced if n<=1 but it makes sense
to assume so. clapack.Dgehrd references tau always by taking the address of
its first element. One option is to pass nil from clapack to the C call if
the length of tau is zero. Another option is to pass a dummy slice from
cgo.Dgehrd to clapack in such situation. The first option is probably the
correct one but @btracey should confirm it.

btracey: This sounds like a bug in clapack. I would file an issue with lapack. For now, it seems easiest to shortcut the algorithm when n <= 1.

vladimir-ch: I think it is a bug in our clapack package that assumes that `tau` has always length at least 1 (which it has but only for n >= 2).

btracey: Ohh, I see. Yea, it seems easiest to just do a short-path return before calling into clapack

native/dgehrd.go
vladimir-ch: In the end I decided to follow fotran, using ilo = 0 and ihi = -1 if n == 0.

testlapack/dgehrd.go
vladimir-ch: Done.

clapack's Dgehrd should still be fixed I think, it is public and it does not follow the behavior of the reference.

cgo/lapack.go
```diff
@@ -459,6 +461,80 @@ func (impl Implementation) Dgeqrf(m, n int, a []float64, lda int, tau, work []fl
...
+// -1 if n == 0, otherwise Dgehrd will panic.
+//
+// On return, tau will contain the scalar factors of the elementary reflectors.
+// Elements tau[:ilo] and tau[ihi:] of tau will be set to zero. It must have
+// length equal to n-1 if n > 1, otherwise Dgehrd will panic.
+//
+// work must have length at least lwork, otherwise Dgehrd will panic, and lwork
+// must be at least max(1,n).
+//
+// The C interface does not support providing temporary storage. To provide
```
vladimir-ch: I think that if we exposed the middle-level interface in clapack we could remove this ugly exception everywhere in this package, and avoid the extra allocation and the nan checks in Lapacke.

vladimir-ch:
PTAL

cgo/lapack.go
kortschak: I don't see an easy way to do that; work is allocated in the clapack code.

vladimir-ch: It is if we use C.LAPACKE_dgehrd. But we can use C.LAPACKE_dgehrd_work.

testlapack/dgehrd.go
kortschak: OK, so what do we want to do?

It seems to me that the most general way of dealing with this is to check for zero length of any numeric slice type and only take the address if its length >= 0.

So `genlapack.pl` would generate for a parameter p of type []T:

```
...
var _p *T
if len(p) > 0 {
    _p = &p[0]
}
return isZero(C.LAPACKE_func(..., (*C.lapack_T)(_p), ...)
...
```

cgo/lapack.go
kortschak: That's true. do we care about the NaN checks?

vladimir-ch: I think that not much. We don't do them in `native`, fortran does not do them.
Is there any downside to using the `_work` version?

kortschak: I'm just looking at that now. The code generation should be pretty straightforward, the process of figuring out when is not. Not all functions take a work parameter, so I can't just reject `.h` lines without a work parameter. If I do two passes over the file then I can collect that information.

kortschak: That works. So, the next question is whether we keep the old ones and add FuncWork or just conform to the Fortran API more closely and always require work (I think the latter).

testlapack/dgehrd.go
btracey: Most of the cases we do have tests for a slice being of size zero in the other tests, i.e. panics of slice of insufficient length. We can probably just add them in as necessary, right?

native/dgehrd.go
```diff
@@ -0,0 +1,178 @@
...
+	var i int
+	if nb < nbmin || nh <= nb {
+		// Use unblocked code below.
+		i = ilo
+	} else {
+		// Use blocked code.
+
+		bi := blas64.Implementation()
+		iwt := n * nb // Size of the matrix Y and index where the matrix T starts in work.
+		for i = ilo; i < ihi-nx; i += nb {
```
vladimir-ch: I noticed that @btracey in `dgebrd.go` changed `nx` in a very similar loop to `nb`. `nx` is the crossover block size between blocked and unblocked code, its default value is 128, while the default block size is 32 (minimum 2). This does seem a bit strange because the loop is ever started to be executed only for matrices larger than 129 (even if the minimum block size of 2 is used). I'm not sure what the right action is here, I'm just point it out. I would like to merge this PR so that I can continue submitting what I have in the queue.

vladimir-ch:
_ping_

native/dgehrd.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+	ldwork := nb // work is used as an n×nb matrix.
+
+	var i int
+	if nb < nbmin || nh <= nb {
+		// Use unblocked code below.
+		i = ilo
+	} else {
+		// Use blocked code.
+
```
kortschak: Nit: this blank line is odd. Maybe use it down above L175.

native/dgehrd.go
```diff
@@ -0,0 +1,177 @@
...
+	const (
+		nbmax = 64
+		ldt   = nbmax + 1
+		tsize = ldt * nbmax
+	)
+	// Compute the workspace requirements.
+	nb := min(nbmax, impl.Ilaenv(1, "DGEHRD", " ", n, ilo, ihi, -1))
+	lwkopt := n*nb + tsize
+	work[0] = float64(lwkopt)
```
kortschak: Out of curiosity, what happens if this assignment is not done except in the following conditional block?

kortschak:
Implementation LGTM - I'll leave the tests for @btracey.

kortschak:
Note that if this goes in after #137 it will need to be updated to pass `work` and `lwork` to the clapack function.

native/dgehrd.go
vladimir-ch: I just removed it, I like the unblocked code not being separated from the blocked one.

native/dgehrd.go
vladimir-ch: Thanks, inside the block is the right place for this assignment. Done.

testlapack/dgehrd.go
```diff
@@ -0,0 +1,230 @@
...
+					if ilo > ihi {
+						ilo, ihi = ihi, ilo
+					}
+					testDgehrd(t, impl, n, ilo, ihi, extra, optwork, rnd)
+				}
+			}
+		}
+	}
+
+	// Selected tests for larger matrix sizes to test the blocked
```
kortschak: I read "Selected tests" as *noun* *verb* rather than *adj* *noun* (I think because of the normal Go documentation style - it seems to have broken my general English parsing). Maybe rephrase as "These are selected tests...".

lapack.go
```diff
@@ -24,6 +24,7 @@ type Complex128 interface{}
 // Float64 defines the public float64 LAPACK API supported by gonum/lapack.
 type Float64 interface {
 	Dgecon(norm MatrixNorm, n int, a []float64, lda int, anorm float64, work []float64, iwork []int) float64
+	Dgehrd(n, ilo, ihi int, a []float64, lda int, tau, work []float64, lwork int)
```
btracey: Does this need to be part of the interface? To date we've only exported those that are used in mat64 directly.

testlapack/dgehrd.go
btracey: Or just "Select tests"

btracey: I guess that has the same problem.

btracey:
LGTM

btracey:
With the interface issue addressed.

lapack.go
vladimir-ch: Probably not. Removed.

vladimir-ch:
I have also updated the call to clapack.Dgehrd

Thanks for the review.

By the way, I did not record any hard numbers but it seems to me that now with the _work calls the cgo tests are only slightly slower than our native. I remember that TestDbdsqr was taking about twice as long in cgo than in native.

btracey:
Is that pure go vs. pure c, or go + cblas vs. pure c?

btracey:
Oh, you're using the test suite, so it's pure go vs. pure c. I'd be interested in seeing how much of that is that our BLAS can be much slower.

vladimir-ch:
Yes, it's pure go vs. pure c+fortran via OpenBLAS. I was wrong, however, the tests in cgo not only don't execute faster than before, they take slightly more time than before. But native go is still faster.

Before switching to `*_work` routines:
```
 native git:(c254fff) go test -v
PASS
ok  	github.com/gonum/lapack/native	38.979s
 cgo git:(c254fff) LD_LIBRARY_PATH=$HOME/src/OpenBLAS CGO_LDFLAGS="-L$HOME/src/OpenBLAS -lopenblas" go test -v
PASS
ok  	github.com/gonum/lapack/cgo	50.992s
```
After switching to `*_work` routines:
```
 native git:(19c6723) go test -v
PASS
ok  	github.com/gonum/lapack/native	39.009s
 cgo git:(19c6723) LD_LIBRARY_PATH=$HOME/src/OpenBLAS CGO_LDFLAGS="-L$HOME/src/OpenBLAS -lopenblas" go test -v
PASS
ok  	github.com/gonum/lapack/cgo	55.688s
```

PR #135, opened 2016-06-02
lapack: fix typo in comment for NormFrob

vladimir-ch:


kortschak:
I think we may have discovered a new element. What all small things are made from.

LGTM

PR #139, opened 2016-06-08
native: add Dlasy2

vladimir-ch:


native/dlasy2.go
```diff
@@ -0,0 +1,272 @@
...
+// are not checked.
+//
+// Dlasy2 returns a scale factor that is chosen less than or equal to 1 to
+// prevent the solution overflowing, the infinity norm of the solution, and an
+// indicator of success. If ok is false, TL and TR have too close eigenvalues,
+// so TL or TR is perturbed to get a non-singular equation.
+//
+// Dlasy2 is an internal routine. It is exported for testing purposes.
+func (impl Implementation) Dlasy2(tranl, tranr bool, isgn, n1, n2 int, tl []float64, ldtl int, tr []float64, ldtr int, b []float64, ldb int, x []float64, ldx int) (scale, xnorm float64, ok bool) {
+	ok = true
```
vladimir-ch: I follow the fortran reference in not doing any checks. They are avoiding them for the sake of performance. Let me know if I should include them.

native/dlasy2.go
```diff
@@ -0,0 +1,272 @@
...
+			btmp [2]float64
+		)
+		if n1 == 1 && n2 == 2 {
+			// 1×2 case: TL11*[X11 X12] + sgn*[X11 X12]*op[TR11 TR12] = [B11 B12].
+			//                                            [TR21 TR22]
+			smin := math.Abs(tl[0])
+			smin = math.Max(smin, math.Max(math.Abs(tr[0]), math.Abs(tr[1])))
+			smin = math.Max(smin, math.Max(math.Abs(tr[ldtr]), math.Abs(tr[ldtr+1])))
+			smin = math.Max(eps*smin, smlnum)
+			tmp[0] = tl[0] + sgn*tr[0]
```
vladimir-ch: tmp is a 2x2 matrix, here following fortran in being column-major. I don't feel much like converting it to row-major, but again, let me know if I should do it.

native/dlasy2.go
```diff
@@ -0,0 +1,272 @@
...
+	// Set pivots less than smin to smin.
+
+	smin := math.Max(math.Abs(tr[0]), math.Abs(tr[1]))
+	smin = math.Max(smin, math.Max(math.Abs(tr[ldtr]), math.Abs(tr[ldtr+1])))
+	smin = math.Max(smin, math.Max(math.Abs(tl[0]), math.Abs(tl[1])))
+	smin = math.Max(smin, math.Max(math.Abs(tl[ldtl]), math.Abs(tl[ldtl+1])))
+	smin = math.Max(eps*smin, smlnum)
+
+	var t16 [16]float64
+	t16[0*4+0] = tl[0] + sgn*tr[0]
```
vladimir-ch: The same here, t16 is 4x4 col-major matrix.

vladimir-ch:
PTAL

native/dlasy2.go
kortschak: It could be done with a build constraint like we do with mat64 and the `At`/`Set` methods.

native/dlasy2.go
kortschak: I don't mind - maybe @btracey does, but if you do leave it as column major, leave a comment here saying so.

native/dlasy2.go
```diff
@@ -0,0 +1,272 @@
...
+		locl21 := [4]int{1, 0, 3, 2}
+		l21 := tmp[locl21[ipiv]] / u11
+		locu22 := [4]int{3, 2, 1, 0}
+		u22 := tmp[locu22[ipiv]] - u12*l21
+		if math.Abs(u22) <= smin {
+			ok = false
+			u22 = smin
+		}
+		bswpiv := [4]bool{false, true, false, true}
+		if bswpiv[ipiv] {
```
kortschak: Can't we just to `ipiv&0x1 != 0`?

native/dlasy2.go
```diff
@@ -0,0 +1,272 @@
...
+		scale = 1
+		if 2*smlnum*math.Abs(btmp[1]) > math.Abs(u22) || 2*smlnum*math.Abs(btmp[0]) > math.Abs(u11) {
+			scale = 0.5 / math.Max(math.Abs(btmp[0]), math.Abs(btmp[1]))
+			btmp[0] *= scale
+			btmp[1] *= scale
+		}
+		x21 := btmp[1] / u22
+		x20 := btmp[0]/u11 - (u12/u11)*x21
+		xswpiv := [4]bool{false, false, true, true}
+		if xswpiv[ipiv] {
```
kortschak: `ipiv&0x2 != 0`?

native/dlasy2.go
vladimir-ch: Is creating an issue for this enough for now?

native/dlasy2.go
vladimir-ch: Added.

native/dlasy2.go
vladimir-ch: A comment added also here.

native/dlasy2.go
vladimir-ch: Clever. Done.

native/dlasy2.go
vladimir-ch: Done.

native/dlasy2.go
kortschak: Yes, but add a TODO here to mark it as well.

native/dlasy2.go
```diff
@@ -0,0 +1,270 @@
...
+	if n1 == 0 || n2 == 0 {
+		return scale, xnorm, ok
+	}
+
+	// Set constants to control overflow.
+	eps := dlamchP
+	smlnum := dlamchS / eps
+	sgn := float64(isgn)
+
+	if n1 == 1 && n2 == 1 {
```
kortschak: The conditional operations here are *so* much nicer than in the Fortran.

native/dlasy2.go
```diff
@@ -0,0 +1,270 @@
...
+		// 1×2 or 2×1 case.
+		var (
+			smin float64
+			tmp  [4]float64 // tmp is used as a 2×2 column-major matrix.
+			btmp [2]float64
+		)
+		if n1 == 1 && n2 == 2 {
+			// 1×2 case: TL11*[X11 X12] + sgn*[X11 X12]*op[TR11 TR12] = [B11 B12].
+			//                                            [TR21 TR22]
+			smin = math.Abs(tl[0])
```
kortschak: Ugggh. The Fortran for this is a nightmare.

native/dlasy2.go
```diff
@@ -0,0 +1,270 @@
...
+		}
+		scale = 1
+		if 2*smlnum*math.Abs(btmp[1]) > math.Abs(u22) || 2*smlnum*math.Abs(btmp[0]) > math.Abs(u11) {
+			scale = 0.5 / math.Max(math.Abs(btmp[0]), math.Abs(btmp[1]))
+			btmp[0] *= scale
+			btmp[1] *= scale
+		}
+		x21 := btmp[1] / u22
+		x20 := btmp[0]/u11 - (u12/u11)*x21
+		if ipiv&0x2 != 0 { // true for ipiv equal to 2 and 3.
```
kortschak: Here the Fortran manages to be a little more informative. Partly because of the array I got you to delete, but also because it uses the `xswap` temporary var. In lieu of those, would you add a comment here about what this is doing, not just the truth of `ipiv&0x2 != 0` (which I'm not actually sure is needed, but you can leave).

(also for the `bswap` case above).

native/dlasy2.go
```diff
@@ -0,0 +1,270 @@
...
+		x[0] = b[0] * scale / tau1
+		xnorm = math.Abs(x[0])
+		return scale, xnorm, ok
+	}
+
+	if n1+n2 == 3 {
+		// 1×2 or 2×1 case.
+		var (
+			smin float64
+			tmp  [4]float64 // tmp is used as a 2×2 column-major matrix.
```
kortschak: Would you look at the performance impact of making this a `[2][2]float64`? If it's not too great it would greatly clarify this var's use.

native/dlasy2.go
```diff
@@ -0,0 +1,270 @@
...
+			} else {
+				tmp[1] = tl[ldtl]
+				tmp[2] = tl[1]
+			}
+			btmp[0] = b[0]
+			btmp[1] = b[ldb]
+		}
+		// Solve 2×2 system using complete pivoting.
+		// Set pivots less than smin to smin.
+		bi := blas64.Implementation()
```
kortschak: `bi` is used non-conditionally after we have reached L60, so maybe declare it there.

native/dlasy2.go
```diff
@@ -0,0 +1,270 @@
...
+	// Solve equivalent 4×4 system using complete pivoting.
+	// Set pivots less than smin to smin.
+
+	smin := math.Max(math.Abs(tr[0]), math.Abs(tr[1]))
+	smin = math.Max(smin, math.Max(math.Abs(tr[ldtr]), math.Abs(tr[ldtr+1])))
+	smin = math.Max(smin, math.Max(math.Abs(tl[0]), math.Abs(tl[1])))
+	smin = math.Max(smin, math.Max(math.Abs(tl[ldtl]), math.Abs(tl[ldtl+1])))
+	smin = math.Max(eps*smin, smlnum)
+
+	var t16 [16]float64 // t16 is used as a 4×4 column-major matrix.
```
kortschak: *This one* could be a `[4][4]float64`.

native/dlasy2.go
```diff
@@ -0,0 +1,270 @@
...
+			for jp := i; jp < 4; jp++ {
+				if math.Abs(t16[ip*4+jp]) >= xmax {
+					xmax = math.Abs(t16[ip*4+jp])
+					ipsv = ip
+					jpsv = jp
+				}
+			}
+		}
+		if ipsv != i {
+			bi.Dswap(4, t16[ipsv*4:], 1, t16[i*4:], 1)
```
kortschak: Please add a comment here and in the swap below saying which is swapping cols and which is swapping rows. It is obvious, but the interaction between the col-major t16 and the Fortran and the row-major blas leave a risk.

native/dlasy2.go
```diff
@@ -0,0 +1,270 @@
...
+	}
+	if math.Abs(t16[3*4+3]) < smin {
+		t16[3*4+3] = smin
+	}
+	scale = 1
+	if 8*smlnum*math.Abs(btmp[0]) > math.Abs(t16[0*4+0]) ||
+		8*smlnum*math.Abs(btmp[1]) > math.Abs(t16[1*4+1]) ||
+		8*smlnum*math.Abs(btmp[2]) > math.Abs(t16[2*4+2]) ||
+		8*smlnum*math.Abs(btmp[3]) > math.Abs(t16[3*4+3]) {
+		maxbtmp := math.Max(math.Abs(btmp[0]), math.Abs(btmp[1]))
```
kortschak: Add a blank line above this to help readability.

native/dlasy2.go
```diff
@@ -0,0 +1,270 @@
...
+	for i := 0; i < 4; i++ {
+		k := 3 - i
+		temp := 1 / t16[k*4+k]
+		tmp[k] = btmp[k] * temp
+		for j := k + 1; j < 4; j++ {
+			tmp[k] -= temp * t16[k*4+j] * tmp[j]
+		}
+	}
+	for i := 0; i < 3; i++ {
+		if jpiv[2-i] != 2-i {
```
kortschak: Maybe a quick note about `2-i` cf `4-1`. (I find these things harder to keep in my head than bit fiddling, which does get a comment).

testlapack/dlasy2.go
```diff
@@ -0,0 +1,102 @@
...
+	if tranr {
+		blas64.Gemm(blas.NoTrans, blas.Trans, float64(isgn), x, tr, 1, lhsWant)
+	} else {
+		blas64.Gemm(blas.NoTrans, blas.NoTrans, float64(isgn), x, tr, 1, lhsWant)
+	}
+	// Compare them.
+	for i := 0; i < n1; i++ {
+		for j := 0; j < n2; j++ {
+			diff := lhsWant.Data[i*lhsWant.Stride+j] - b.Data[i*b.Stride+j]
+			if math.Abs(diff) > 1e-11 && ok {
```
kortschak: I think we probably want a `const epsilon = 1e-11` at the outset of the func (do we not do this elsewhere?).

native/dlasy2.go
vladimir-ch: I think I misled you because of my (not unusual) momentary confusion, sorry. tmp and t16 are both row-major. Otherwise this would not work (Dswap just swaps vectors, does not care about row/col major), also at other places the indexing is row-major. I got confused by how they map b to btmp and tmp to x.

With this misconception cleared, maybe I can use [2][2]tmp and [4][4]t16 (renamed to just t?) after all if I replaced these calls to blas with loops and assignments.

native/dlasy2.go
vladimir-ch: I'll have to revisit this considering the row/col major issue. It is good, it will increase my understanding. I'll sprinkle the code with some comments.

native/dlasy2.go
vladimir-ch: Hopefully I disentangled it correctly.

native/dlasy2.go
vladimir-ch: Done. I hope the whole code is clearer now.

native/dlasy2.go
vladimir-ch: Let me know if I should do this, it's not a big change.

native/dlasy2.go
vladimir-ch: Now I use `bi` only in the 2x1/1x2 case. It could (must if we do [2][2]tmp) be replaced by a loop.

native/dlasy2.go
vladimir-ch: Actually, done. I think it is much cleaner and clearer this way.

native/dlasy2.go
vladimir-ch: Done.

native/dlasy2.go
vladimir-ch: Replaced by the loop going backwards which is hopefully clearer.

testlapack/dlasy2.go
vladimir-ch: Done. I haven't done this in my PRs so far, I will do from now on.

vladimir-ch:
Ok, I ended up redoing the 2x1/1x2 and 2x2 cases so that everything is row-major. Previously the matrices tmp and t16 were row-major, but the mapping from the original tl, tr, and b matrices to this equivalent system was column-major. As a consequence, the indexing details have departed from the netlib reference, but the algorithm and the operations are the same. PTAL

native/dlasy2.go
```diff
@@ -0,0 +1,289 @@
...
+		for k := i + 1; k < 4; k++ {
+			t[k][i] /= t[i][i]
+			btmp[k] -= t[k][i] * btmp[i]
+			for j := i + 1; j < 4; j++ {
+				t[k][j] -= t[k][i] * t[i][j]
+			}
+		}
+	}
+	if math.Abs(t[3][3]) < smin {
+		t[3][3] = smin
```
vladimir-ch: It is slightly surprising for me that the reference does not set `INFO = 1` here. Above in the 1x2 case they do set it if u22 is too small. Seems like a tiny bug to me.

kortschak:
LGTM, but wait for @btracey.

native/dlasy2.go
```diff
@@ -0,0 +1,289 @@
...
+
+// Dlasy2 solves the Sylvester matrix equation where the matrices are of order 1
+// or 2. It computes the unknown n1×n2 matrix X so that
+//  TL*X   + sgn*X*TR   = scale*B,  if tranl == false and tranr == false,
+//  TL^T*X + sgn*X*TR   = scale*B,  if tranl == true  and tranr == false,
+//  TL*X   + sgn*X*TR^T = scale*B,  if tranl == false and tranr == true,
+//  TL^T*X + sgn*X*TR^T = scale*B,  if tranl == true  and tranr == true,
+// where TL is n1×n1, TR is n2×n2, B is n1×n2, and 1 <= n1,n2 <= 2.
+//
+// isgn must be 1 or -1, and n1 and n2 must be 0, 1, or 2, but these conditions
```
btracey: Is there a good reason to not check them?

native/dlasy2.go
```diff
@@ -0,0 +1,289 @@
...
+// are not checked.
+//
+// Dlasy2 returns a scale factor that is chosen less than or equal to 1 to
```
btracey: Dlansy2 returns three parameters, a scale ...

(was having trouble parsing the sentence)

native/dlasy2.go
```diff
@@ -0,0 +1,289 @@
...
+// prevent the solution overflowing, the infinity norm of the solution, and an
+// indicator of success. If ok is false, TL and TR have too close eigenvalues,
```
btracey: I know this is what fortran says, but...

`If ok is false, TL and TR have eigenvalues that are very close, and so a slight perturbation of TL or TR will lead to an infinite number of solutions to the system of equations`

native/dlasy2.go
```diff
@@ -0,0 +1,289 @@
...
+			// 1×2 case: TL11*[X11 X12] + sgn*[X11 X12]*op[TR11 TR12] = [B11 B12].
+			//                                            [TR21 TR22]
+			smin = math.Abs(tl[0])
+			smin = math.Max(smin, math.Max(math.Abs(tr[0]), math.Abs(tr[1])))
+			smin = math.Max(smin, math.Max(math.Abs(tr[ldtr]), math.Abs(tr[ldtr+1])))
+			smin = math.Max(eps*smin, smlnum)
+			tmp[0] = tl[0] + sgn*tr[0]
+			tmp[3] = tl[0] + sgn*tr[ldtr+1]
+			if tranr {
+				tmp[1] = sgn * tr[1]
```
btracey: This looks backward. tr(1,2) should be tr[ldtr] I think.

native/dlasy2.go
```diff
@@ -0,0 +1,289 @@
...
+				tmp[2] = sgn * tr[ldtr]
+			} else {
+				tmp[1] = sgn * tr[ldtr]
```
btracey: Same here.

native/dlasy2.go
```diff
@@ -0,0 +1,289 @@
...
+			// 2×1 case: op[TL11 TL12]*[X11] + sgn*[X11]*TR11 = [B11].
+			//             [TL21 TL22]*[X21]       [X21]        [B21]
+			smin = math.Abs(tr[0])
+			smin = math.Max(smin, math.Max(math.Abs(tl[0]), math.Abs(tl[1])))
+			smin = math.Max(smin, math.Max(math.Abs(tl[ldtl]), math.Abs(tl[ldtl+1])))
+			smin = math.Max(eps*smin, smlnum)
+			tmp[0] = tl[0] + sgn*tr[0]
+			tmp[3] = tl[ldtl+1] + sgn*tr[0]
+			if tranl {
+				tmp[1] = tl[ldtl]
```
btracey: Same here. These look backward.

native/dlasy2.go
kortschak: Good catch.

native/dlasy2.go
```diff
@@ -0,0 +1,289 @@
...
+		bi := blas64.Implementation()
+		ipiv := bi.Idamax(len(tmp), tmp[:], 1)
+		// Compute the upper triangular matrix [u11 u12].
+		//                                     [  0 u22]
+		u11 := tmp[ipiv]
+		if math.Abs(u11) <= smin {
+			ok = false
+			u11 = smin
+		}
+		locu12 := [4]int{1, 0, 3, 2} // Index in tmp of the element on the same row as the pivot.
```
btracey: Why are these shifted from the fortran? I expected `locu12 := [4]int{2,3,0,1}`

native/dlasy2.go
```diff
@@ -0,0 +1,289 @@
...
+
+	smin := math.Max(math.Abs(tr[0]), math.Abs(tr[1]))
+	smin = math.Max(smin, math.Max(math.Abs(tr[ldtr]), math.Abs(tr[ldtr+1])))
+	smin = math.Max(smin, math.Max(math.Abs(tl[0]), math.Abs(tl[1])))
+	smin = math.Max(smin, math.Max(math.Abs(tl[ldtl]), math.Abs(tl[ldtl+1])))
+	smin = math.Max(eps*smin, smlnum)
+
+	var t [4][4]float64
+	t[0][0] = tl[0] + sgn*tr[0]
+	t[1][1] = tl[0] + sgn*tr[ldtr+1]
```
btracey: I don't understand these. Why are t[1][1] and t[2][2] swapped?

native/dlasy2.go
```diff
@@ -0,0 +1,289 @@
...
+		t[2][3] = sgn * tr[1]
+		t[3][2] = sgn * tr[ldtr]
+	} else {
+		t[0][1] = sgn * tr[ldtr]
+		t[1][0] = sgn * tr[1]
+		t[2][3] = sgn * tr[ldtr]
+		t[3][2] = sgn * tr[1]
+	}
+
+	var btmp [4]float64
```
btracey: Same here, why is b swapped?

native/dlasy2.go
```diff
@@ -0,0 +1,289 @@
...
+					xmax = math.Abs(t[ip][jp])
+					ipsv = ip
+					jpsv = jp
+				}
+			}
+		}
+		if ipsv != i {
+			// The pivot is not in the top row of the unprocessed
+			// block, swap rows ipsv and i of t and btmp.
+			t[ipsv], t[i] = t[i], t[ipsv]
```
btracey: Doesn't the dswap in Fortran swap 4 elements?

native/dlasy2.go
btracey: A sml bug you mean?

btracey:
The code looks correct given the swaps, but I don't really understand. The interior matrices are one thing, but all of the input matrices should be row major. I likely don't understand, but I'm confused.

Other than that, code looks good.

testlapack/dlasy2.go
```diff
@@ -0,0 +1,104 @@
...
+	"github.com/gonum/blas/blas64"
+)
+
+type Dlasy2er interface {
+	Dlasy2(tranl, tranr bool, isgn, n1, n2 int, tl []float64, ldtl int, tr []float64, ldtr int, b []float64, ldb int, x []float64, ldx int) (scale, xnorm float64, ok bool)
+}
+
+func Dlasy2Test(t *testing.T, impl Dlasy2er) {
+	rnd := rand.New(rand.NewSource(1))
+	for _, tranl := range []bool{true, false} {
```
btracey: We've almost hit the nesting of the dark regions of the compiler!

native/dlasy2.go
kortschak: `t` is `[4][4]float64`, so `t[i] = t[j]` is an assignment of a `[4]float64`. Array semantics not slice semantics.

native/dlasy2.go
vladimir-ch: I think they do it for the sake of performance.

testlapack/dlasy2.go
kortschak: It looks like real Fortran code.

native/dlasy2.go
vladimir-ch: I don't think so because TR is multiplied with X from the right. For example, in the non-transposed case one gets:
```
L11 [x11 x12] + [x11 x12] [R11 R12] = ( [L11   0] + [R11 R21] ) [x11]
                          [R21 R22] = ( [  0 L11]   [R12 R22] ) [x12]
```

native/dlasy2.go
vladimir-ch: That's because of the row/col ordering. In fortran, tmp[0] and tmp[2] are on the first row, here it is tmp[0] and tmp[1].

native/dlasy2.go
vladimir-ch: The difference here and with the b (and x) matrices is because fortran maps, for example, the matrix b to btmp columnwise, here I do everything rowwise - rowwise storage, rowwise mapping.

native/dlasy2.go
vladimir-ch: It does but I thought it is not necessary to call into blas if we can swap the rows so easily using just go.

testlapack/dlasy2.go
vladimir-ch: I know, I know ... I'll remove the two levels for tranl and tranr and write out the 4 combinations explicitly.

vladimir-ch:
The confusing differences are from the row/column-major thing. Fortran maps the matrices to vectors column-wise and stores them in column-major matrices, here I do the same mapping and storage row-wise. It helps to write out the identity on L164 as the equivalent 4x4 system.

testlapack/dlasy2.go
btracey: I don't actually mind, they're very clean for loops. I just thought it was funny.

native/dlasy2.go
vladimir-ch: A slight perturbation could also lead to a system with no solution. The current (fortran) formulation says that if ok is false, the returned solution is the solution of a slightly perturbed system. I think it is more informative.

native/dlasy2.go
vladimir-ch: :-)

native/dlasy2.go
vladimir-ch: Thanks, done.

PR #140, opened 2016-06-08
native: change loop condition for blocked part of Dgehrd

vladimir-ch:
I brought this up in #134. I feel there is a confusion about what the value 3 passed as `ispec` parameter to `Ilaenv` means.

netlib lapack code uses the returned crossover point as the size of the unprocessed block at which to switch from blocked to unblocked code. Take as an example Dgehrd, n = 130, block size 32, and crossover 128. With these values the blocked code processes 32 columns, remaining size is less than crossover, and the rest is processed by unblocked code. This seems very strange, why not process all 4 blocks and the remaining two columns with unblocked?

The following would make more sense to me: if n is less than crossover, the matrix is small, so don't bother with the blocked code. Once the matrix is larger, as in the above example, there are enough blocks (4), so process them with blocked code and the rest with unblocked code.

I copied the respective docs formulation from Lapack but it does not shed much light on this. It refers to N which suggests it is a decision made based on the matrix size, not on the remaining unprocessed size.

In our native, @btracey also noticed this strangeness and fixed it somewhere to the more sensible option (e.g., in Dgebrd) but not everywhere (e.g., in Dsytrd). It is not as simple as replacing nx with nb, see Dsytrd.go:110.

Perhaps this is something to be asked at the Lapack forum?

### DIFF
--- a/native/dgehrd.go
+++ b/native/dgehrd.go
@@ -137,7 +137,7 @@ func (impl Implementation) Dgehrd(n, ilo, ihi int, a []float64, lda int, tau, wo
 		// Use blocked code.
 		bi := blas64.Implementation()
 		iwt := n * nb // Size of the matrix Y and index where the matrix T starts in work.
-		for i = ilo; i < ihi-nx; i += nb {
+		for i = ilo; i < ihi-nb; i += nb {
 			ib := min(nb, ihi-i)
 
 			// Reduce columns [i:i+ib] to Hessenberg form, returning the
--- a/native/ilaenv.go
+++ b/native/ilaenv.go
@@ -6,10 +6,13 @@ package native
 
 // Ilaenv returns algorithm tuning parameters for the algorithm given by the
 // input string. ispec specifies the parameter to return:
-//  1: The optimal block size for a blocked algorithm.
-//  2: The minimum block size for a blocked algorithm.
-//  3: The block size of unprocessed data at which a blocked algorithm should
-//     crossover to an unblocked version.
+//  1: The optimal block size. If this value is 1, an unblocked algorithm will
+//     give the best performance.
+//  2: The minimum block size for which the block routine should be used. If the
+//     usable block size is less than this value, an unblocked routine should be
+//     used.
+//  3: The crossover point between a blocked and unblocked algorithm. In a block
+//     routine, for N less than this value, an unblocked routine should be used.
 //  4: The number of shifts.
 //  5: The minimum column dimension for blocking to be used.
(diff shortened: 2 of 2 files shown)


vladimir-ch:
Oops, obviously it is not as simple as replacing nx with nb.

btracey:
LGTM.

It's probably worth asking on the lapack forum, but I don't see any reason why you'd want one giant unblocked algorithm.

vladimir-ch:
As always, I've been thinking about this further and for the moment I will not merge this. First of all, the tests panic when the minimum work length is used. The reason is the slightly convoluted logic in Dgehrd that decides when blocked is used and if work is long enough for blocked. If in the loop I leave `ihi-nx`, the blocked loop is not started if nx >= nh (even though it is in the blocked branch of the relevant `if`, argh). If I change it to `ihi - nb`, the blocked loop is started even though above the corresponding work check did not take place. For them to take place I would have to change the logic there, effectively eliminating the need for nx.

The documentation for `ispec == 3` is not completely clear but I think that the reference lapack does what it should do, nx is just an additional tuning parameter. It seems that the right approach is to keep using nx and adjust the values returned by `Ilaenv` (e.g., by returning nx that is equal to the block size).

Interestingly, when experimenting with this I supplied to Dgehrd work slightly shorter than optimal. This results in shorter than optimal block size ... and the test completes in 18.762s instead of 20.924s. Playing with the numbers in `Ilaenv` might apparently bring some speed up.

If we decide to respect `Ilaenv(ispec=3)` and follow the reference, then we should also revert the modifications that have already been done at other places.

kortschak:
>Interestingly, when experimenting with this I supplied to Dgehrd work slightly shorter than optimal. This results in shorter than optimal block size ... and the test completes in 18.762s instead of 20.924s. Playing with the numbers in `Ilaenv` might apparently bring some speed up.

Be wary of test completion time over benchmarks. But worth looking into.

vladimir-ch:
> > Interestingly, when experimenting with this I supplied to Dgehrd work slightly shorter than optimal. This results in shorter than optimal block size ... and the test completes in 18.762s instead of 20.924s. Playing with the numbers in Ilaenv might apparently bring some speed up.

> Be wary of test completion time over benchmarks. But worth looking into.

I should definitely stop making such statements, now back on the master branch I cannot reproduce this. Who knows what I was doing. Sorry.

jonlawlor:
There is a bug in the new coverage code, working on it.

vladimir-ch:

The algorithmic part has been done in the new repository. I will think if and how to update the comment for Ilaenv.

PR #141, opened 2016-06-10
native,cgo: add Dlarfx

vladimir-ch:


cgo/lapack.go
```diff
@@ -158,6 +169,44 @@ func (impl Implementation) Dlantr(norm lapack.MatrixNorm, uplo blas.Uplo, diag b
...
+// v must have length equal to m if side == blas.Left, and equal to n if side ==
+// blas.Right, otherwise Dlarfx will panic.
+//
+// c and ldc represent the m×n matrix C. On return, C is overwritten by the
+// matrix H * C if side == blas.Left, or C * H if side == blas.Right.
+//
+// work must have length at least n if side == blas.Left, and at least m if side
+// == blas.Right, otherwise Dlarfx will panic. work is not referenced if H has
+// order < 11.
+func (impl Implementation) Dlarfx(side blas.Side, m, n int, v []float64, tau float64, c []float64, ldc int, work []float64) {
```
vladimir-ch: Do we have any policy for adding functions to the cgo package? For example, the other `Dlarf[bgt]` routines were not added.

native/dlarfx.go
```diff
@@ -0,0 +1,480 @@
...
+	if tau == 0 {
+		return
+	}
+
+	if side == blas.Left {
+		// Form H * C, where H has order m.
+		switch m {
+		default:
+			impl.Dlarf(side, m, n, v, 1, tau, c, ldc, work)
+		case 0:
```
kortschak: I think an explicit return here make this clearer.

native/dlarfx.go
```diff
@@ -0,0 +1,480 @@
...
+		panic(badSide)
+	}
+
+	if tau == 0 {
+		return
+	}
+
+	if side == blas.Left {
+		// Form H * C, where H has order m.
+		switch m {
```
kortschak: Can we have the case comments.

native/dlarfx.go
kortschak: Does the Fortran do nothing for m == 0? Is this determined in dlarf?

cgo/lapack.go
kortschak: We add the ones that are needed to implement the interface.

btracey: I disagree with @kortschak . Ideally, we add functions to cgo that exist in the clapack wrapping and aren't broken. The Dlarf[*] methods aren't there because they were some of the first added to the package, and so before Dan rightly started demanding comparison with clapack.

btracey: Other missing functions are oversights on my part

kortschak: That's fair. Policy now established!

btracey: And so it's fully established, the functions in the interface are those that are needed by mat64. I'm not sure if we should maintain that strictly... we at one point used Dgels but we don't any more.

kortschak: We can remove them prior to the stability guarantee, but not after.

native/dlarfx.go
vladimir-ch: > I think an explicit return here make this clearer. Possibly in the other cases below as well. It saves a scroll to find out what is happening at the cost of nine extra lines. (Both up to you though).

Definitely clearer, added everywhere.

> Does the Fortran do nothing for m == 0? Is this determined in dlarf?

In fortran Dlarfx, the zero case is handled by Dlarf. Dlarf does not say much about it but looking at the code it seems that the case is a no-op. If in our test I start the loops for m and n from 0, I get some panics. I tried to dig into it a little but the hole is too deep. That's why I added the explicit zero case to native Dlarfx.

kortschak: Maybe add a comment on the m == 0 case.

I think given the size of the case blocks, judicious addition of new lines after the returns is warranted. I looks nicer in this case I think.

vladimir-ch: Yes, it looks nicer. Done.

vladimir-ch:
_ping_

native/dlarfx.go
```diff
@@ -0,0 +1,527 @@
...
+		}
+		return
+
+	case 2: // Special code for 2×2 Householder matrix.
+		v0 := v[0]
+		t0 := tau * v0
+		v1 := v[1]
+		t1 := tau * v1
+		for j := 0; j < m; j++ {
+			sum := v0*c[j*ldc] + v1*c[j*ldc+1]
```
btracey: For all of these, consider `cs := c[j*ldc:]`. This code is then more easily vectorized. TODO is fine.

btracey:
LGTM

native/dlarfx.go
vladimir-ch: Done.

vladimir-ch:
Thanks for the review.

PR #145, opened 2016-06-13
native,cgo: add Dorghr

vladimir-ch:


cgo/lapack.go
```diff
@@ -821,6 +821,46 @@ func (impl Implementation) Dorgbr(vect lapack.DecompUpdate, m, n, k int, a []flo
...
+	case ilo < 0 || max(1, n) <= ilo:
+		panic(badIlo)
+	case ihi < min(ilo, n-1) || n <= ihi:
+		panic(badIhi)
+	case lwork < max(1, nh) && lwork != -1:
+		panic(badWork)
+	case len(work) < max(1, lwork):
+		panic(shortWork)
+	}
+	clapack.Dorghr(n, ilo+1, ihi+1, a, lda, tau, work, lwork)
```
vladimir-ch: This is where the cgo interface departs from the clapack interface (also occurs in Dgehrd and others): we assume ilo and ihi to be zero-based, clapack assumes them to be one-based. I'm just pointing it out, I think having one-based indices in Go (or C) is strange.

native/dorghr.go
```diff
@@ -0,0 +1,89 @@
...
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package native
+
+// Dorghr generates an n×n orthogonal matrix Q which is defined as the product
+// of ihi-ilo elementary reflectors, as returned by Dgehrd:
+//  Q = H_{ilo} H_{ilo+1} ... H_{ihi-1}.
+// ilo and ihi must have the same values as in the previous call of Dgehrd. Q
+// will be equal to the identity matrix except in the submatrix
```
btracey: Is this before or after the call to Dorghr. It's unclear.

btracey:
LGTM with the one documentation fix.

PR #146, opened 2016-06-13
native,cgo: add Dormhr

vladimir-ch:


native/dormhr.go
```diff
@@ -0,0 +1,117 @@
...
+		return
+	}
+
+	if m == 0 || n == 0 || nh == 0 {
+		work[0] = 1
+		return
+	}
+	if side == blas.Left {
+		impl.Dormqr(side, trans, nh, n, nh, a[(ilo+1)*lda+ilo:], lda,
+			tau[ilo:ihi], c[(ilo+1)*ldc:], ldc, work, lwork)
```
btracey: I wonder if in the future these should really be three-slice syntax. All the bugs I've encountered/caused were because of reslicing longer than the length, not accessing out of bounds. Two slice doesn't prevent those bugs.

@kortschak ?

btracey:
LGTM

native/dormhr.go
kortschak: Yes, that SGTM

PR #148, opened 2016-06-15
native: set ok to false in 2x2 case in Dlasy2

vladimir-ch:
Bug report to upstream was accepted:
http://icl.cs.utk.edu/lapack-forum/viewtopic.php?f=13&t=4975


kortschak:
LGTM

PR #149, opened 2016-06-15
testlapack: avoid panic in *OutsideAllNan for zero-sized matrices

vladimir-ch:
Simply remove the checking of NaNs after the matrix because everywhere where I use this function the slices are allocated tightly anyway.

Sorry that this slipped me through.

### DIFF
--- a/testlapack/general.go
+++ b/testlapack/general.go
@@ -134,7 +134,11 @@ func generalOutsideAllNaN(a blas64.General) bool {
 		}
 	}
 	// Check after last element.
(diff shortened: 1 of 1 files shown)


kortschak:
It's not harming anyone, and in the future someone may not be as careful as you. I'd leave it in.

vladimir-ch:
Ok, I'll leave it in but I have to do something about it anyway, it panics (now in DorghrTest) if Rows == 0.

vladimir-ch:
PTAL @kortschak

kortschak:
LGTM

PR #150, opened 2016-06-17
go 1.7beta1 -> go 1.7beta2

jonlawlor:


sbinet:
hum... it fails https://travis-ci.org/gonum/lapack/jobs/138444863#L505
(but this seems to be a pre-existing condition)

LGTM

jonlawlor:
Yes, I think that error already exists, because all of the versions are failing.

kortschak:
#149?

jonlawlor:
Looks like it.

PR #151, opened 2016-06-21
native: add Dlahqr

vladimir-ch:
Another PR on the slightly heavier side. PTAL @kortschak @btracey

### DIFF
--- a/native/dlahqr.go
+++ b/native/dlahqr.go
@@ -1,0 +1,418 @@
+// Copyright ©2016 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package native
+
+import (
+	"math"
+
+	"github.com/gonum/blas/blas64"
+)
+
+// Dlahqr computes the eigenvalues and Schur factorization of a block of an n×n
+// upper Hessenberg matrix H, using the double-shift/single-shift QR algorithm.
+//
+// h and ldh represent the matrix H. Its contents on return depends on the value
+// of info. Dlahqr works primarily with the Hessenberg submatrix
+// H[ilo:ihi+1,ilo:ihi+1], but applies transformations to all of H if wantt is
+// true. It is assumed that H[ihi+1:n,ihi+1:n] is already upper
+// quasi-triangular, although this is not checked.
+//
+// It must hold that
+//  0 <= ilo <= max(0,ihi), and ihi < n,
+// and that
+//  H[ilo,ilo-1] == 0,  if ilo > 0,
+// otherwise Dlahqr will panic.
+//
+// wr[ilo:ihi+1] and wi[ilo:ihi+1] will contain, on return, the real and
+// imaginary parts, respectively, of the computed eigenvalues ilo to ihi. If two
+// eigenvalues are computed as a complex conjugate pair, they are stored in
+// consecutive elements of wr and wi, say the i-th and (i+1)th, with wi[i] > 0
+// and wi[i+1] < 0. If wantt is true, the eigenvalues are stored in the same
+// order as on the diagonal of the Schur form returned in H, with wr[i] =
+// H[i,i], and, if H[i:i+2,i:i+2) is a 2×2 diagonal block, wi[i] =
+// sqrt(abs(H[i+1,i]*H[i,i+1])) and wi[i+1] = -wi[i].
+//
+// wr and wi must have length n.
+//
+// z and ldz represent an n×n matrix Z. If wantz is true, the transformations
+// will be applied to the submatrix Z[iloz:ihiz+1,ilo:ihi+1]. If wantz is false,
+// z is not referenced. It must hold that
+//  0 <= iloz <= ilo; ihi <= ihiz < n.
+//
+// If info == 0 and wantt is true, on return H[ilo:ihi+1,ilo:ihi+1] will be
+// overwritten by upper quasi-triangular full Schur form with any 2×2 diagonal
+// blocks in standard form.
+//
+// If info == 0 and wantt is false, the contents of h on return is unspecified.
+//
+// If info > 0, Dlahqr failed to compute all the eigenvalues ilo to ihi in a
+// total of 30 iterations per eigenvalue. wr[info+1:ihi+1] and wi[info+1:ihi+1]
+// contain those eigenvalues which have been successfully computed.
+//
+// If info > 0 and wantt is false, on return, the remaining unconverged
+// eigenvalues are the eigenvalues of the upper Hessenberg matrix
+// H[ilo:info+1,ilo:info+1].
+//
+// If info > 0 and wantt is true, then on return
+//  (initial H)*U = U*(final H),   (*)
+// where U is an orthogonal matrix. The final H is upper Hessenberg and
+// H[info+1:ihi+1,info+1:ihi+1] is upper triangular.
+//
+// If info > 0 and wantz is true, then on return
+//  (final Z) = (initial Z)*U,
+// where U is the orthogonal matrix in (*) regardless of the value of wantt.
+//
+// Dlahqr is an internal routine. It is exported for testing purposes.
+func (impl Implementation) Dlahqr(wantt, wantz bool, n, ilo, ihi int, h []float64, ldh int, wr, wi []float64, iloz, ihiz int, z []float64, ldz int) (info int) {
+	checkMatrix(n, n, h, ldh)
+	switch {
+	case ilo < 0 || max(0, ihi) < ilo:
+		panic(badIlo)
+	case n <= ihi:
+		panic(badIhi)
+	case len(wr) != n:
+		panic("lapack: bad length of wr")
+	case len(wi) != n:
+		panic("lapack: bad length of wi")
+	case ilo > 0 && h[ilo*ldh+ilo-1] != 0:
+		panic("lapack: block is not isolated")
+	}
+	if wantz {
+		checkMatrix(n, n, z, ldz)
+		switch {
+		case iloz < 0 || ilo < iloz:
+			panic("lapack: iloz out of range")
+		case ihiz < ihi || n <= ihiz:
+			panic("lapack: ihiz out of range")
+		}
+	}
+
+	// Quick return if possible.
+	if n == 0 {
+		return info
+	}
+	if ilo == ihi {
+		wr[ilo] = h[ilo*ldh+ilo]
+		wi[ilo] = 0
+		return info
+	}
+
+	// Clear out the trash.
+	for j := ilo; j < ihi-2; j++ {
+		h[(j+2)*ldh+j] = 0
+		h[(j+3)*ldh+j] = 0
+	}
+	if ilo <= ihi-2 {
+		h[ihi*ldh+ihi-2] = 0
+	}
+
+	nh := ihi - ilo + 1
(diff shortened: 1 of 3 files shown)


native/dlahqr.go
```diff
@@ -0,0 +1,421 @@
...
+	bi := blas64.Implementation()
+	i := ihi
+	for i >= ilo {
+		l := ilo
+
+		// Perform QR iterations on rows and columns ilo to i until a
+		// submatrix of order 1 or 2 splits off at the bottom because a
+		// subdiagonal element has become negligible.
+		converged := false
+		for its := 0; its < itmax; its++ {
```
kortschak: `its <= itmax`

native/dlahqr.go
```diff
@@ -0,0 +1,421 @@
...
+			// Look for a single small subdiagonal element.
+			var k int
+			for k = i; k >= l+1; k-- {
```
kortschak: `k > l`

native/dlahqr.go
```diff
@@ -0,0 +1,421 @@
...
+						tst += math.Abs(h[(k+1)*ldh+k])
+					}
+				}
+				// The following is a conservative small
+				// subdiagonal deflation criterion due to Ahues
+				// & Tisseur (LAWN 122, 1997). It has better
+				// mathematical foundation and improves accuracy
+				// in some cases.
+				if math.Abs(h[k*ldh+k-1]) <= ulp*tst {
+					ab := math.Max(math.Abs(h[k*ldh+k-1]), math.Abs(h[(k-1)*ldh+k]))
```
kortschak: Since you're using 0 as a typographic place holder in some places, use them in these two lines as well.

native/dlahqr.go
```diff
@@ -0,0 +1,421 @@
...
+			var m int
+			for m = i - 2; m >= l; m-- {
+				// Determine the effect of starting the
+				// double-shift QR iteration at row m, and see
+				// if this would make H[m,m-1] negligible. The
+				// following uses scaling to avoid overflows and
+				// most underflows.
+				h21s := h[(m+1)*ldh+m]
+				s := math.Abs(h[m*ldh+m]-rt2r) + math.Abs(rt2i) + math.Abs(h21s)
+				h21s /= s
```
kortschak: Nice.

native/dlahqr.go
```diff
@@ -0,0 +1,421 @@
...
+				v1 := h21s*h[m*ldh+m+1] + (h[m*ldh+m]-rt1r)*((h[m*ldh+m]-rt2r)/s) - rt2i/s*rt1i
```
kortschak: You use `v [3]float64` below. Why not use it here? No push, just curious.

native/dlahqr.go
```diff
@@ -0,0 +1,421 @@
...
+				v[0], t0 = impl.Dlarfg(nr, v[0], v[1:], 1)
+				if k > m {
+					h[(k+0)*ldh+k-1] = v[0]
+					h[(k+1)*ldh+k-1] = 0
+					if k < i-1 {
+						h[(k+2)*ldh+k-1] = 0
+					}
+				} else if m > l {
+					// Use the following instead of H[k,k-1]
+					// = -H[k,k-1] to avoid a bug when v(2)
```
kortschak: Avoid breaking in an expression?

native/dlahqr.go
```diff
@@ -0,0 +1,421 @@
...
+					// and v(3) underflow.
+					h[k*ldh+k-1] *= 1 - t0
+				}
+				v1 := v[1]
```
kortschak: What is the reason for this? and the v2 assignment below? I don't see v being modified here.

native/dlahqr.go
```diff
@@ -0,0 +1,421 @@
...
+//  (initial H)*U = U*(final H),   (*)
+// where U is an orthogonal matrix. The final H is upper Hessenberg and
+// H[info+1:ihi+1,info+1:ihi+1] is upper triangular.
+//
+// If info > 0 and wantz is true, then on return
+//  (final Z) = (initial Z)*U,
+// where U is the orthogonal matrix in (*) regardless of the value of wantt.
+//
+// Dlahqr is an internal routine. It is exported for testing purposes.
+func (impl Implementation) Dlahqr(wantt, wantz bool, n, ilo, ihi int, h []float64, ldh int, wr, wi []float64, iloz, ihiz int, z []float64, ldz int) (info int) {
```
kortschak: Do we use info?

vladimir-ch: We will, in `Dhseqr` and `Dlaqr3`.

native/dlahqr.go
vladimir-ch: Thanks, done.

native/dlahqr.go
vladimir-ch: Ah, of course, done.

native/dlahqr.go
vladimir-ch: I removed them.

native/dlahqr.go
```diff
@@ -0,0 +1,421 @@
...
+
+			// Now the active submatrix is in rows and columns l to
+			// i. If eigenvalues only are being computed, only the
+			// active submatrix need be transformed.
+			if !wantt {
+				i1 = l
+				i2 = i
+			}
+
+			const (
```
vladimir-ch: I'll try to find out.

native/dlahqr.go
vladimir-ch: This was something I wanted to revisit before submitting but apparently I did not. Moved `var v [3]float64` before this loop and started using it here.

native/dlahqr.go
vladimir-ch: Done.

native/dlahqr.go
vladimir-ch: Not sure. I removed `v1` and `v2` and use `v` directly.

vladimir-ch:
Done. PTAL

native/dlahqr.go
vladimir-ch: The most useful description of these low-level details that I found is http://www8.cs.umu.se/~kressner/kressner.pdf, page 31, but from a quick look I can't immediately connect the exceptional shift polynomial to those two constants. Having access to the references from that section would probably help, especially
```
D. Day. How the shifted QR algorithm fails to converge and how to fix it.
Tech. Report 96–0913, Sandia National Laboratories, Albuquerque, NM, 1996.
```
seems relevant but I can't find it online.

Interestingly, in previous versions of lapack (v2.0?) and in eispack the constants are `0.75` and `-0.4375`.

vladimir-ch: In
```
J. H. Wilkinson and C. Reinsch. Handbook for Automatic Computation. Vol. II
Linear Algebra. Springer-Verlag, New York, 1971.
```
on page 362 they use a simple exceptional shifts that also involves a parameter. They use 1.5 and write 
```
No deep significance should be attributed to the factor 1.5; it was included
since it appeared to diminish the probability of the choice being "unfortunate".
This strategy is used again after 20 unsuccessful iterations. If 30 unsuccessful
iterations are needed then a failure indication is given.
```
However, for example -0.4375 does not seem to be as arbitrary as 1.5.

native/dlahqr.go
```diff
@@ -0,0 +1,418 @@
...
+	// i1 and i2 are the indices of the first row and last column of H to
+	// which transformations must be applied. If eigenvalues only are being
+	// computed, i1 and i2 are set inside the main loop.
+	var i1, i2 int
+	if wantt {
+		i1 = 0
+		i2 = n - 1
+	}
+
+	itmax := 30 * max(10, nh) // Total number of QR iterations allowed.
```
kortschak: This is actually not true (now) since the for loop was changed. I don't know why the Fortran does what it does.

kortschak:
Implementation LGTM. I'll leave the tests for @btracey.

native/dlahqr.go
vladimir-ch: Ok, by going backwards and fitting the constants to the code and to the above references, I can get to something where these numbers come up, at least for the case of eispack and old lapack. 

A rough outline follows.

A shift polynomial is the characteristic polynomial of the 2x2 submatrix at the bottom right corner of the active block (the default, Francis' double shift, case here). If a 1x1 or 2x2 submatrix does not split off in 10 (and then in another 10) iterations, they use an exceptional shift polynomial `p_e` instead, for example,
`p_e(x) := x^2 - 3/2 s x + s^2`
where
`s := |h_{i,i-1}| + |h_{i-1,i-2}|.`
So they want to form a real 2x2 matrix `[h11 h12; h21 h22]` whose characteristic polynomial is `p_e`. They take the matrix in the form
`[ a b ]`
`[ b a ]`
Then its characteristic polynomial is
`det([ a-x b; b a-x]) = x^2 - 2ax + a^2-b^2`
Comparing coefficients in front of the same powers of `x` in `p_e`, we get
`3/2 s = 2 a`
`s^2 = a^2-b^2`
From the first equation we have that `a = 3/4 s`, and that's the 0.75 constant in, for example,
https://github.com/johannesgerer/jburkardt-f77/blob/master/eispack/eispack.f#L4389
Substitute `a` into the second equation and we get that `b^2 = -7/16 s^2`. And `-7/16 == -0.4375` in https://github.com/johannesgerer/jburkardt-f77/blob/master/eispack/eispack.f#L4391
But this means that b must be imaginary while they want a real matrix, so instead they form `[a b^2; b^2 a]`. I don't see an immediate justification as to why, maybe it will come later. Or maybe I'm missing something.

That's for the old lapack, the 10 and 20 cases in the switch here form the matrix in a different way. One possibility is that they use a different shift polynomial but I'm not sure.

kortschak: That's probably good enough for our purposes here. Thanks.

native/dlahqr.go
vladimir-ch: I think that it became true thanks to the change you suggested. The convergence criterion is checked at the beginning of the `its` loop, so at the time of checking it exactly `its` iterations have been done.

native/dlahqr.go
vladimir-ch: One final comment.

Dlahqr from Lapack v2.0 https://github.com/inducer/arpack/blob/master/LAPACK/dlahqr.f#L216
implies that the 2x2 matrix they form is simply any matrix `[a ?; ? a]` where the product of the off-diagonal elements gives `-0.4375 s^2`. Subtracting of the H[i,i] element happened already in the 2.0 version https://github.com/inducer/arpack/blob/master/LAPACK/dlahqr.f#L238
So the last (well, almost) remaining question to explain the form of `a` is the change of DAT1 from 0.75 to 3. I think I'll leave it here.

testlapack/dlahqr.go
```diff
@@ -0,0 +1,268 @@
...
+	"github.com/gonum/blas/blas64"
+)
+
+type Dlahqrer interface {
+	Dlahqr(wantt, wantz bool, n, ilo, ihi int, h []float64, ldh int, wr, wi []float64, iloz, ihiz int, z []float64, ldz int) int
+}
+
+func DlahqrTest(t *testing.T, impl Dlahqrer) {
+	rnd := rand.New(rand.NewSource(1))
+	for _, wantt := range []bool{true, false} {
```
btracey: Could you add a comment for the two different loop sets?

btracey:
LGTM

PR #153, opened 2016-06-30
native: add Dlaexc

vladimir-ch:


native/dlaexc.go
```diff
@@ -0,0 +1,267 @@
...
+		panic("lapack: invalid value of n1")
+	}
+	if n2 < 0 || 2 < n2 {
+		panic("lapack: invalid value of n2")
+	}
+
+	if n == 0 || n1 == 0 || n2 == 0 {
+		return true
+	}
+	if j1+n1 >= n {
```
vladimir-ch: It's strange that Lapack checks whether the first block fits into the matrix but not the second. And moreover return true in that case?

native/dlaexc.go
```diff
@@ -0,0 +1,267 @@
...
+		t22 := t[j2*ldt+j2]
+
+		// Determine the transformation to perform the interchange.
+		cs, sn, _ := impl.Dlartg(t[j1*ldt+j2], t22-t11)
+
+		// Apply transformation to the matrix T.
+		if j3 < n {
+			bi.Drot(n-j3, t[j1*ldt+j3:], 1, t[j2*ldt+j3:], 1, cs, sn)
+		}
+		if j1 > 0 {
```
vladimir-ch: I'm not sure why Lapack checks whether there are any columns to transform after the block but not whether there are any rows above it. Drot should be a no-op if the size is 0 anyway.

native/dlaexc.go
```diff
@@ -0,0 +1,267 @@
...
+		return true
+	}
+
+	// Swapping involves at least one 2×2 block.
+	//
+	// Copy the diagonal block of order n1+n2 to the local array d and
+	// compute its norm.
+	nd := n1 + n2
+	ldd := nd
+	d := make([]float64, nd*nd)
```
vladimir-ch: `d` does not escape, so it should be allocated on stack, shouldn't it? Either way, I can make it to be a 4x4 array like in Fortran, if desired.

native/dlaexc.go
```diff
@@ -0,0 +1,267 @@
...
+	impl.Dlacpy(blas.All, nd, nd, t[j1*ldt+j1:], ldt, d, ldd)
+	dnorm := impl.Dlange(lapack.MaxAbs, nd, nd, d, ldd, work)
+
+	// Compute machine-dependent threshold for test for accepting swap.
+	eps := dlamchP
+	thresh := math.Max(10*eps*dnorm, dlamchS/eps)
+
+	// Solve T11*X - X*T22 = scale*T12 for X.
```
vladimir-ch: Same here with `x`.

native/dlaexc.go
btracey: Doesn't it escape because it is used in Dlange? I don't fully understand the rules of escape analysis.

native/dlaexc.go
```diff
@@ -0,0 +1,267 @@
...
+
+		t11 := t[j1*ldt+j1]
+		t22 := t[j2*ldt+j2]
+
+		// Determine the transformation to perform the interchange.
+		cs, sn, _ := impl.Dlartg(t[j1*ldt+j2], t22-t11)
+
+		// Apply transformation to the matrix T.
+		if j3 < n {
+			bi.Drot(n-j3, t[j1*ldt+j3:], 1, t[j2*ldt+j3:], 1, cs, sn)
```
kortschak: `n-j3`? j2?

native/dlaexc.go
```diff
@@ -0,0 +1,267 @@
...
+		// Swap two 1×1 blocks.
+
+		t11 := t[j1*ldt+j1]
+		t22 := t[j2*ldt+j2]
+
+		// Determine the transformation to perform the interchange.
+		cs, sn, _ := impl.Dlartg(t[j1*ldt+j2], t22-t11)
+
+		// Apply transformation to the matrix T.
+		if j3 < n {
```
kortschak: `<=`?

native/dlaexc.go
kortschak: Ask the compiler.

native/dlaexc.go
```diff
@@ -0,0 +1,267 @@
...
+			// Accumulate transformation in the matrix Q.
+			impl.Dlarfx(blas.Right, n, 3, u[:], tau, q[j1:], ldq, work)
+		}
+
+	case n1 == 2 && n2 == 1:
+		//  Generate elementary reflector H so that:
+		//   H (  -X11 ) = ( * )
+		//     (  -X21 ) = ( 0 )
+		//     ( scale ) = ( 0 )
+		u[1] = -x[ldx]
```
kortschak: Nice deconvolution.

native/dlaexc.go
vladimir-ch: If by asking you mean something like `go build -gcflags -S 2>&1 | less`, then I asked. When `d` and `x` are created with `make`, there are 2 calls to `runtime.makeslice` in the output. When they are local arrays, `[16]float64` and `[4]float64`, respectively, there is one call to `runtime.newobject` for `d` but no such call for `x`. For `x` there is instead
```
MOVQ    $0, "".x+432(SP)
MOVQ    $0, "".x+440(SP)
MOVQ    $0, "".x+448(SP)
MOVQ    $0, "".x+456(SP)
```
`x` is obviously on the stack, I don't know about `d`, maybe allocated? Interestingly, there are also calls to `newobject` for each of `u`, `u1`, and `u2` below that are shorter than `x`.

I've changed both `d` and `x` to arrays.

native/dlaexc.go
vladimir-ch: Here we rotate the rows j1 and j2. The numbers at the diagonal are taken care of explicitly, so we need to rotate only the elements after the 2x2 block. `j3` is the index of the first column after the block. and `n-j3` then is the number of the elements.

native/dlaexc.go
vladimir-ch: This is correct, rotate only if there are any elements to rotate.

vladimir-ch:
PTAL

native/dlaexc.go
kortschak: That's one way. You can also do `go build -gcflags='-m'` if you make an executable that calls this func.

kortschak:
LGTM but wait for @btracey.

native/dlaexc.go
btracey: It is strange. Make a note of it?

native/dlaexc.go
btracey: j1 is 0-indexed, to be specific.

native/dlaexc.go
```diff
@@ -0,0 +1,253 @@
...
+	}
+
+	j2 := j1 + 1
+	j3 := j1 + 2
+
+	bi := blas64.Implementation()
+
+	if n1 == 1 && n2 == 1 {
+		// Swap two 1×1 blocks.
+
```
btracey: remove newline?

native/dlaexc.go
```diff
@@ -0,0 +1,253 @@
...
+	var x [4]float64
+	const ldx = 2
+	scale, _, _ := impl.Dlasy2(false, false, -1, n1, n2, d[:], ldd, d[n1*ldd+n1:], ldd, d[n1+1:], ldd, x[:], ldx)
+
+	// Swap the adjacent diagonal blocks.
+	switch {
+	case n1 == 1 && n2 == 2:
+		// Generate elementary reflector H so that
+		//  ( scale, X11, X12 ) H = ( 0, 0, * )
+		u := [3]float64{scale, x[0], 1}
```
btracey: x[1]?

btracey: Oh, nm.

native/dlaexc.go
kortschak: Thanks.

native/dlaexc.go
vladimir-ch: Done.

native/dlaexc.go
vladimir-ch: Done.

vladimir-ch:
PTAL

native/dlaexc.go
```diff
@@ -0,0 +1,259 @@
...
+		t[j2*ldt+j1] = 0
+		t[j3*ldt+j1] = 0
+
+		if wantq {
+			// Accumulate transformation in the matrix Q.
+			impl.Dlarfx(blas.Right, n, 3, u[:], tau, q[j1:], ldq, work)
+		}
+
+	default: // n1 == 2 && n2 == 2
+		// Generate elementary reflectors H(1) and H(2) so that:
```
btracey: Brackets?

native/dlaexc.go
```diff
@@ -0,0 +1,259 @@
...
+			impl.Dlarfx(blas.Right, n, 3, u2[:], tau2, q[j2:], ldq, work)
+		}
+	}
+
+	if n2 == 2 {
+		// Standardize new 2×2 block T11.
+		a, b := t[j1*ldt+j1], t[j1*ldt+j2]
+		c, d := t[j2*ldt+j1], t[j2*ldt+j2]
+		a, b, c, d, _, _, _, _, cs, sn := impl.Dlanv2(a, b, c, d)
+		t[j1*ldt+j1], t[j1*ldt+j2] = a, b
```
btracey: Combine these into above with a `var cs, sn float64`?

native/dlaexc.go
```diff
@@ -0,0 +1,259 @@
...
+		}
+	}
+	if n1 == 2 {
+		// Standardize new 2×2 block T22.
+		j3 := j1 + n2
+		j4 := j3 + 1
+		a, b := t[j3*ldt+j3], t[j3*ldt+j4]
+		c, d := t[j4*ldt+j3], t[j4*ldt+j4]
+		a, b, c, d, _, _, _, _, cs, sn := impl.Dlanv2(a, b, c, d)
+		t[j3*ldt+j3], t[j3*ldt+j4] = a, b
```
btracey: Same here.

native/dlaexc.go
```diff
@@ -0,0 +1,259 @@
...
+		t[j4*ldt+j3], t[j4*ldt+j4] = c, d
+		if j3+2 < n {
```
btracey: I think this is nicer as n - j3 - 2 > 0. Makes the mapping to the next line a lot clearer.

btracey:
LGTM with a couple minor fixes.

native/dlaexc.go
vladimir-ch: Underscore, we denote them as `H_i` at other places.

native/dlaexc.go
vladimir-ch: I wanted to avoid having a too long line but it is not a big deal here. Done.

native/dlaexc.go
vladimir-ch: Done.

native/dlaexc.go
vladimir-ch: Done, also at a couple of other lines above.

PR #156, opened 2016-07-23
populate OpenBLAS cache, unset -ex in scripts

jonlawlor:
Same as gonum/blas#184 - use a prebuilt cache of the OpenBLAS libraries instead of recompiling every time.

### DIFF
--- a/.travis/linux/OpenBLAS/install.sh
+++ b/.travis/linux/OpenBLAS/install.sh
@@ -19,7 +19,6 @@ if [ -e ${CACHE_DIR}/last_commit_id ]; then
     fi
 fi
 
-# check if cache directory exists
 if [ ! -e ${CACHE_DIR}/last_commit_id ]; then
     if [ -d ${CACHE_DIR} ]; then
         # Travis automatically creates the cache directory if it does not exist,
--- a/.travis/linux/OpenBLAS/install.sh
+++ b/.travis/linux/OpenBLAS/install.sh
@@ -20,11 +20,8 @@ if [ -e ${CACHE_DIR}/last_commit_id ]; then
 fi
 
 if [ ! -e ${CACHE_DIR}/last_commit_id ]; then
-    if [ -d ${CACHE_DIR} ]; then
-        # Travis automatically creates the cache directory if it does not exist,
-        # so this is needed for initialization.
-        rm -rf ${CACHE_DIR}
-    fi
+    # Clear cache.
+    rm -rf ${CACHE_DIR}
 
     # cache generation
     echo "Building cache at $CACHE_DIR"
--- a/.travis/linux/OpenBLAS/install.sh
+++ b/.travis/linux/OpenBLAS/install.sh
@@ -63,4 +63,12 @@ pushd cgo/clapack
 go install -v -x
 popd
 
+# travis compiles commands in script and then executes in bash.  By adding
+# set -e we are changing the travis build script's behavior, and the set
+# -e lives on past the commands we are providing it.  Some of the travis
+# commands are supposed to exit with non zero status, but then continue
+# executing.  set -x makes the travis log files extremely verbose and
+# difficult to understand.
+# 
+# see travis-ci/travis-ci#5120
 set +ex
--- a/.travis/linux/OpenBLAS/install.sh
+++ b/.travis/linux/OpenBLAS/install.sh
@@ -29,9 +29,6 @@ if [ ! -e ${CACHE_DIR}/last_commit_id ]; then
     sudo git clone --depth=1 git://github.com/xianyi/OpenBLAS
 
     pushd OpenBLAS
-    echo OpenBLAS $(git rev-parse HEAD)
-    # write commit id to cache location
-    echo $(git rev-parse HEAD) > ${CACHE_DIR}/last_commit_id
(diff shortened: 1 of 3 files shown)


jonlawlor:
PTAL @kortschak @btracey - built times dropped by 10 minutes, still 11+, but a lot better.

btracey:
This looks okay to me, but I would wait for @kortschak

.travis/linux/OpenBLAS/install.sh
```diff
@@ -1,31 +1,58 @@
...
-# make and install cblas
-pushd CBLAS
-sudo mv Makefile.LINUX Makefile.in
-sudo BLLIB=/usr/lib/libopenblas.a make alllib
-sudo mv lib/cblas_LINUX.a /usr/lib/libcblas.a
-popd
-popd
+# check if cache directory exists
+if [ ! -e ${CACHE_DIR}/last_commit_id ]; then
+    if [ -d ${CACHE_DIR} ]; then
```
vladimir-ch: I think it is simpler to `rm -rf ${CACHE_DIR}` unconditionally.

.travis/linux/OpenBLAS/install.sh
```diff
@@ -1,31 +1,58 @@
...
+
+    # cache generation
+    echo "Building cache at $CACHE_DIR"
+    mkdir ${CACHE_DIR}
+    sudo git clone --depth=1 git://github.com/xianyi/OpenBLAS
+
+    pushd OpenBLAS
+    echo OpenBLAS $(git rev-parse HEAD)
+    # write commit id to cache location
+    echo $(git rev-parse HEAD) > ${CACHE_DIR}/last_commit_id
```
vladimir-ch: Maybe do this at the end of the if block after all the actions have finished?

.travis/linux/OpenBLAS/install.sh
```diff
@@ -1,31 +1,58 @@
...
+fi
 
-# make and install cblas
-pushd CBLAS
-sudo mv Makefile.LINUX Makefile.in
-sudo BLLIB=/usr/lib/libopenblas.a make alllib
-sudo mv lib/cblas_LINUX.a /usr/lib/libcblas.a
-popd
-popd
+# check if cache directory exists
```
vladimir-ch: This comment is not accurate. Maybe it is ok to remove it, it is clear what the if condition is.

.travis/linux/OpenBLAS/install.sh
```diff
@@ -1,31 +1,58 @@
...
+    # cache generation
+    echo "Building cache at $CACHE_DIR"
+    mkdir ${CACHE_DIR}
+    sudo git clone --depth=1 git://github.com/xianyi/OpenBLAS
+
+    pushd OpenBLAS
+    echo OpenBLAS $(git rev-parse HEAD)
+    # write commit id to cache location
+    echo $(git rev-parse HEAD) > ${CACHE_DIR}/last_commit_id
+    sudo make FC=gfortran &> /dev/null && sudo make PREFIX=${CACHE_DIR} install
```
vladimir-ch: Not sure but use /usr/ as PREFIX here and in CBLAS below and then avoid copying the whole cache to /usr/?

.travis/linux/OpenBLAS/install.sh
```diff
@@ -1,31 +1,58 @@
...
+        rm -rf ${CACHE_DIR}
+    fi
+
+    # cache generation
+    echo "Building cache at $CACHE_DIR"
+    mkdir ${CACHE_DIR}
+    sudo git clone --depth=1 git://github.com/xianyi/OpenBLAS
+
+    pushd OpenBLAS
+    echo OpenBLAS $(git rev-parse HEAD)
```
kortschak: Don't do this twice?

.travis/linux/OpenBLAS/install.sh
kortschak: Agree.

jonlawlor: The files need to be in the cache directory to be persisted... maybe symlinks would be appropriate?

vladimir-ch: Or don't `make install` here, do it below instead of `cp -r`?

jonlawlor: That would require a clone of OpenBLAS in every build to get the make files, or copying the make files over as well.  Can you elaborate on the issue you see with this approach a little more?  I don't think I'm getting it right now.

My reasoning is that anything that `make install` would have placed in `/usr/...` could be used by gonum/lapack, and the exact files could change in different OpenBLAS versions (although, that is unlikely to happen).  The cache directory itself is only ~8MB, so it isn't going to take long to copy, either from AWS or from the cache dir to the `/usr/...` locations.  (There are actually 4 different versions of the cache, one for each go version, which each get separately stored in AWS, for a total space of ~33 MB, but only 8MB actually gets downloaded in each individual CI run).

jonlawlor:
I haven't resolved @vladimir-ch's issues.

.travis/linux/OpenBLAS/install.sh
vladimir-ch: Sorry, now I understand. For some reason I thought that OpenBLAS and CBLAS are subdirectories of CACHE_DIR.

vladimir-ch:
LGTM

PR #157, opened 2016-07-27
native: add Dlaln2

vladimir-ch:


native/dlaln2.go
```diff
@@ -0,0 +1,386 @@
...
+//
+// xnorm contains the infinity-norm of X when X is regarded as a na×nw real
+// matrix.
+//
+// ok will be false if (ca A - w D) had to be perturbed to make its smallest
+// singular value greater than smin, otherwise ok will be true.
+//
+// Dlaln2 is an internal routine. It is exported for testing purposes.
+func (impl Implementation) Dlaln2(trans bool, na, nw int, smin, ca float64, a []float64, lda int, d1, d2 float64, b []float64, ldb int, wr, wi float64, x []float64, ldx int) (scale, xnorm float64, ok bool) {
+	if na != 1 && na != 2 {
```
btracey: Is it worth making these typed constants?

Maybe better still is to add a TODO that maybe these should be split into different functions given we can have a complex type? Not sure what our future plan is (if any) on that.

native/dlaln2.go
```diff
@@ -0,0 +1,386 @@
+// Copyright ©2016 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package native
+
+import "math"
+
+// Dlaln2 solves a 1×1 or 2×2 linear system of equations of the form
```
btracey: Can I just comment how strange it is to have a function that solves a "1x1 system of linear equations". You mean division?

native/dlaln2.go
```diff
@@ -0,0 +1,386 @@
...
+		// Check scaling for X = B / C.
+		bnorm := math.Abs(b[0]) + math.Abs(b[1])
+		if cnorm < 1 && bnorm > math.Max(1, bignum*cnorm) {
+			scale = 1 / bnorm
+		}
+
+		// Compute X.
+		cx := complex(scale*b[0], scale*b[1]) / complex(csr, csi)
+		x[0], x[1] = real(cx), imag(cx)
+		xnorm = math.Abs(x[0]) + math.Abs(x[1])
```
btracey: I agree that's what Lapack does, but that's a strange definition of complex norm...

native/dlaln2.go
```diff
@@ -0,0 +1,386 @@
...
+	}
+
+	// 2×2 system.
+
+	// Compute the real part of
+	//  C = ca A   - w D
+	// or
+	//  C = ca A^T - w D.
+	crv := [4]float64{
+		ca*a[0] - wr*d1, ca * a[1],
```
btracey: Could you put these on newlines? I know it's like a 2x2, but the comma is subtle, and so it looks like only two entries are specified

native/dlaln2.go
```diff
@@ -0,0 +1,386 @@
...
+		ca * a[lda], ca*a[lda+1] - wr*d2,
+	}
+	if trans {
```
btracey: It looks like the layout is backward from Lapack. Just checking this is intentional.

native/dlaln2.go
```diff
@@ -0,0 +1,386 @@
...
+	xnorm = math.Max(math.Abs(xr1)+math.Abs(xi1), math.Abs(xr2)+math.Abs(xi2))
+
+	// Further scaling if norm(A)*norm(X) > overflow.
+	if xnorm > 1 && cmax > 1 && xnorm > bignum/cmax {
+		temp := cmax / bignum
+		x[0] *= temp
+		x[1] *= temp
+		x[ldx] *= temp
+		x[ldx+1] *= temp
+		xnorm = temp
```
btracey: xnorm and scale should be *=temp?

btracey:
LGTM with minor fixes

native/dlaln2.go
vladimir-ch: I'm not sure. `na` is the problem size, so making it a typed constant is not worth it I think, even though it can take only two values. Making `nw` a bool would be natural but again I'm not sure if it's worth it.

Splitting the function into two and using a complex type for w, b, x, is an interesting idea. Similar Go-ification would be applicable also to other routines. It touches on the question of how much we want to follow LAPACK API in such detail. With these internal routines we probably don't need to restict ourselves too much.

I'll add a TODO.

native/dlaln2.go
vladimir-ch: Good catch, thanks.

native/dlaln2.go
vladimir-ch: Yes, it is intentional, I convert to row-major even these internal computations.

native/dlaln2.go
vladimir-ch: Done.

native/dlaln2.go
vladimir-ch: Yes, in the end it is just a careful division but the statement as it is is correct. Between two alternatives that I saw:
`Dlaln2 solves a linear system of na equations of the form ...`
`Dlaln2 solves a linear equation or a system of 2 equations of the form ...`.
I went with the second one.

native/dlaln2.go
vladimir-ch: Unusual but still a norm.

PR #159, opened 2016-08-01
native,testlapack: document and test that Dlaexc return true when n1=n2=1

vladimir-ch:
`Dtrexc` uses (depends on) this behavior, so it makes sense to make it more formal by documenting and testing it. PTAL


btracey:
LGTM

PR #162, opened 2016-08-02
native,testlapack: fix Dlaexc and its test

vladimir-ch:
I'm sorry. If only someone could elucidate what exactly I was thinking when I wrote the original ...

Such a wholly unremarkable off-by-one error. Why this went (almost) undetected: because of the error, the matrix d did not have zeros where it should, therefore Dlaexc returned with a false in the `> thresh` checks, and therefore the tests exited early `if !ok`. I did notice that Dlaexc returned sometimes with a false but I didn't notice that it was always except for the trivial or 1x1 cases. Therefore I mistakenly attributed it to floating point effects.

The tests are a different story. I noticed **something** is wrong with them after fixing the indexing error. I don't know what I was thinking there.

PTAL @kortschak @btracey

### DIFF
--- a/native/dlaexc.go
+++ b/native/dlaexc.go
@@ -111,7 +111,7 @@ func (impl Implementation) Dlaexc(wantq bool, n int, t []float64, ldt int, q []f
 	// Solve T11*X - X*T22 = scale*T12 for X.
 	var x [4]float64
 	const ldx = 2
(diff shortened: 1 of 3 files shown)


vladimir-ch:
Added a couple of helper functions to simplify the tests and removed Dlanv2 from the Dlaexcer interface.

btracey:
Sorry for missing that earlier. Glad you corrected the mistake. It's generally pretty hard to review these as we all know. LGTM.

PR #165, opened 2016-08-03
native,cgo: add Dtrexc

vladimir-ch:
Needs #162 but submitting in advance to give time for review. The test for this covers Dlaexc completely, so it serves as a further confirmation of correctness of Dlaexc.

### DIFF
--- a/cgo/lapack.go
+++ b/cgo/lapack.go
@@ -1332,6 +1332,79 @@ func (impl Implementation) Dtrcon(norm lapack.MatrixNorm, uplo blas.Uplo, diag b
 	return rcond[0]
 }
 
+// Dtrexc reorders the real Schur factorization of a n×n real matrix
+//  A = Q*T*Q^T
+// so that the diagonal block of T with row index ifst is moved to row ilst.
+//
+// On entry, T must be in Schur canonical form, that is, block upper triangular
+// with 1×1 and 2×2 diagonal blocks; each 2×2 diagonal block has its diagonal
+// elements equal and its off-diagonal elements of opposite sign.
+//
+// On return, T will be reordered by an orthogonal similarity transformation Z
+// as Z^T*T*Z, and will be again in Schur canonical form.
+//
+// If compq is lapack.EigDecomp, on return the matrix Q of Schur vectors will be
+// updated by postmultiplying it with Z. If compq is lapack.EigValueOnly, the
+// matrix Q is not referenced and will not be updated. For other values of compq
+// Dtrexc will panic.
+//
+// ifst and ilst specify the reordering of the diagonal blocks of T. The block
+// with row index ifst is moved to row ilst, by a sequence of transpositions
+// between adjacent blocks.
+//
+// If ifst points to the second row of a 2×2 block, ifstOut will point to the
+// first row, otherwise it will be equal to ifst.
+//
+// ilstOut will point to the first row of the block in its final position. If ok
+// is true, ilstOut may differ from ilst by +1 or -1.
+//
+// It must hold that
+//  0 <= ifst < n, and  0 <= ilst < n,
+// otherwise Dtrexc will panic.
+//
+// If ok is false, two adjacent blocks were too close to swap because the
+// problem is very ill-conditioned. T may have been partially reordered, and
+// ilstOut will point to the first row of the block at the position to which it
+// has been moved.
+//
+// work must have length at least n, otherwise Dtrexc will panic.
+//
+// Dtrexc is an internal routine. It is exported for testing purposes.
+func (impl Implementation) Dtrexc(compq lapack.EigComp, n int, t []float64, ldt int, q []float64, ldq int, ifst, ilst int, work []float64) (ifstOut, ilstOut int, ok bool) {
+	checkMatrix(n, n, t, ldt)
+	switch compq {
+	default:
+		panic(badEigComp)
+	case lapack.EigValueOnly:
+		// q is not referenced but LAPACKE checks that ldq >= n always.
+		q = nil
+		ldq = max(1, n)
+	case lapack.EigDecomp:
+		checkMatrix(n, n, q, ldq)
+	}
+	if (ifst < 0 || n <= ifst) && n > 0 {
+		panic("lapack: ifst out of range")
+	}
+	if (ilst < 0 || n <= ilst) && n > 0 {
+		panic("lapack: ilst out of range")
+	}
+	if len(work) < n {
+		panic(badWork)
+	}
+
+	// Quick return if possible.
+	if n <= 1 {
+		return ifst, ilst, true
+	}
+
+	ifst32 := []int32{int32(ifst + 1)}
+	ilst32 := []int32{int32(ilst + 1)}
+	ok = clapack.Dtrexc(lapack.CompSV(compq), n, t, ldt, q, ldq, ifst32, ilst32, work)
+	ifst = int(ifst32[0] - 1)
+	ilst = int(ilst32[0] - 1)
+	return ifst, ilst, ok
+}
+
 // Dtrtri computes the inverse of a triangular matrix, storing the result in place
 // into a. This is the BLAS level 3 version of the algorithm which builds upon
 // Dtrti2 to operate on matrix blocks instead of only individual columns.
--- a/native/dtrexc.go
+++ b/native/dtrexc.go
@@ -1,0 +1,220 @@
+// Copyright ©2016 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package native
+
+import "github.com/gonum/lapack"
+
+// Dtrexc reorders the real Schur factorization of a n×n real matrix
+//  A = Q*T*Q^T
+// so that the diagonal block of T with row index ifst is moved to row ilst.
+//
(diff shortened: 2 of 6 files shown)


native/dtrexc.go
```diff
@@ -0,0 +1,217 @@
...
+				}
+				here += nbnext
+				// Test if 2×2 block breaks into two 1×1 blocks.
+				if nbf == 2 && t[(here+1)*ldt+here] == 0 {
+					nbf = 3
+				}
+				continue
+			}
+
+			// Current block consists of two 1×1 blocks each of
```
vladimir-ch: The rest of the loop body here and in the default case below is not covered by the tests. I don't know how to prepare a 2x2 block in Schur canonical form (with two conjugate eigenvalues) that after swapping with the neighboring block splits into two 1x1 blocks (that is, the block suddeny has two real eigenvalues). I see two explanations:

1. Either some weird, rare floating point effects could occur but I feel this is unlikely. In case of numerical trouble Dlaexc returns false. Or
2. It is to handle the case that the input 2x2 block is in fact not in Schur canonical form and has two real eigenvalues. In other words, the code handles more general case than what the documentation indicates. This is supported by the fact that if I prepare the matrix T to have 2x2 blocks not necessarily in Schur canonical form (and therefore with some probability some will have two real eigenvalues), these two sections become covered. The tests (mostly) succeed although I disabled the eigenvalue comparisons because they need the blocks to be in the Schur form. A couple (in the current setup 17) of cases fail because of numerical effects at `dlaexc.go:204`. The values involved in the check there are typically something like `5.7870375158586285e-15 6.071532165918825e-15 5.350671296697453e-15`.

I will try to ask LAPACK directly but independently of that, do we want to do anything about it?

vladimir-ch:
Ready for review, PTAL

cgo/lapack.go
```diff
@@ -1332,6 +1332,76 @@ func (impl Implementation) Dtrcon(norm lapack.MatrixNorm, uplo blas.Uplo, diag b
...
+// between adjacent blocks.
+//
+// On return, if ifst pointed to the second row of a 2×2 block, ifstOut will
+// point to the first row, otherwise it will be equal to ifst. ilstOut will
+// point to the first row of the block in its final position (if ok is true,
+// ilstOut may differ from ilst by +1 or -1). It must hold that
+//  0 <= ifst < n, and  0 <= ilst < n,
+// otherwise Dtrexc will panic.
+//
+// On return, if ok is false, two adjacent blocks were too close to swap (the
```
kortschak: Replace the parenthesis with "because the problem...".

vladimir-ch: Done (plus some splitting of the previous paragraph).

native/dtrexc.go
```diff
@@ -0,0 +1,217 @@
...
+// work must have length at least n, otherwise Dtrexc will panic.
+//
+// Dtrexc is an internal routine. It is exported for testing purposes.
+func (impl Implementation) Dtrexc(compq lapack.EigComp, n int, t []float64, ldt int, q []float64, ldq int, ifst, ilst int, work []float64) (ifstOut, ilstOut int, ok bool) {
+	checkMatrix(n, n, t, ldt)
+	var wantq bool
+	switch compq {
+	default:
+		panic(badEigComp)
+	case lapack.EigValueOnly:
```
kortschak: Add comment that there is nothing to do.

native/dtrexc.go
```diff
@@ -0,0 +1,220 @@
...
+	}
+
+	ok = true
+
+	// Quick return if possible.
+	if n <= 1 {
+		return ifst, ilst, true
+	}
+
+	// Determine the first row of specified block and find out it is 1×1 or 2×2.
```
kortschak: Break line for consistency?

native/dtrexc.go
```diff
@@ -0,0 +1,220 @@
...
+		nbl = 2
+	}
+
+	switch {
+	case ifst == ilst:
+		return ifst, ilst, true
+
+	case ifst < ilst:
+		// Update ilst.
+		if nbf == 2 && nbl == 1 {
```
kortschak: These conditions are mutually exclusive. Make them `...else if...` or a switch.

native/dtrexc.go
```diff
@@ -0,0 +1,220 @@
...
+
+			if nbf == 1 || nbf == 2 {
+				// Current block either 1×1 or 2×2.
+				nbnext := 1 // Size of the next block.
+				if here+nbf+1 < n && t[(here+nbf+1)*ldt+here+nbf] != 0 {
+					nbnext = 2
+				}
+				ok = impl.Dlaexc(wantq, n, t, ldt, q, ldq, here, nbf, nbnext, work)
+				if !ok {
+					return ifst, here, false
```
kortschak: Just for the sake of clarity, please assign `here` to `ilst` and return the canonically named vars.

native/dtrexc.go
```diff
@@ -0,0 +1,220 @@
...
+			}
+			here += 2
+		}
+		return ifst, here, true
+
+	default: // ifst > ilst
+		here := ifst
+		for here > ilst {
+			// Swap block with next one above.
+
```
kortschak: delete blank line

native/dtrexc.go
```diff
@@ -0,0 +1,220 @@
...
+		if nbf == 2 && nbl == 1 {
+			ilst--
+		}
+		if nbf == 1 && nbl == 2 {
+			ilst++
+		}
+		here := ifst
+		for here < ilst {
+			// Swap block with next one below.
+
```
kortschak: delete blank line

native/dtrexc.go
```diff
@@ -0,0 +1,220 @@
...
+				if nbf == 2 && t[(here+1)*ldt+here] == 0 {
+					nbf = 3
+				}
+				continue
+			}
+
+			// Current block consists of two 1×1 blocks each of
+			// which must be swapped individually.
+			nbnext := 1
+			if here-2 >= 0 && t[(here-1)*ldt+here-2] != 0 {
```
kortschak: `here > 1`? This shows the bounds check most clearly.

native/dtrexc.go
```diff
@@ -0,0 +1,220 @@
...
+				if nbf == 2 && t[(here+1)*ldt+here] == 0 {
+					nbf = 3
+				}
+				continue
+			}
+
+			// Current block consists of two 1×1 blocks each of
+			// which must be swapped individually.
+			nbnext := 1 // Size of the next block.
+			if here+3 < n && t[(here+3)*ldt+here+2] != 0 {
```
kortschak: `here < 3`

native/dtrexc.go
```diff
@@ -0,0 +1,220 @@
...
+
+	default: // ifst > ilst
+		here := ifst
+		for here > ilst {
+			// Swap block with next one above.
+
+			if nbf == 1 || nbf == 2 {
+				// Current block either 1×1 or 2×2.
+				nbnext := 1
+				if here-2 >= 0 && t[(here-1)*ldt+here-2] != 0 {
```
kortschak: `here > 1`

kortschak:
Comments are all take or leave (I don't feel strongly about any of them).

Code LGTM, but wait for @btracey on that and the tests.

native/dtrexc.go
vladimir-ch: Done.

native/dtrexc.go
vladimir-ch: Done.

native/dtrexc.go
vladimir-ch: Done with a switch.

native/dtrexc.go
vladimir-ch: Yeah, I thought about what is the best way when I wrote it and I'm also undecided. Maybe, just maybe, I prefer the current form.

native/dtrexc.go
vladimir-ch: Done.

native/dtrexc.go
vladimir-ch: Done.

native/dtrexc.go
vladimir-ch: Again, I'm undecided. I started using this form after recent @btracey's suggestion. In that case the index was more complex but I started to like this form because it directly links to the index used in `t[...]` (the smaller index must be non-negative). It also helps me in translating from Fortran. For the moment I'm leaving it.

native/dtrexc.go
kortschak: What to do really depends on what the answer from Netlib is. If there is a reasonable answer then there should be a comment block explaining the situation. If your 2. is correct, then adding the relevant tests would be worthwhile.

testlapack/dtrexc.go
```diff
@@ -0,0 +1,226 @@
...
+	if !isOrthonormal(q) {
+		t.Errorf("%v: Q is not orthogonal", prefix)
+	}
+	// Check that Q is unchanged outside of columns [modMin,modMax].
+	for i := 0; i < n; i++ {
+		for j := 0; j < n; j++ {
+			if modMin <= j && j < modMax {
+				continue
+			}
+			diff := q.Data[i*q.Stride+j] - qCopy.Data[i*qCopy.Stride+j]
```
btracey: Just call this directly in the if?

btracey:
LGTM as well

PR #166, opened 2016-08-08
native: add ok to returns from Dlahqr

vladimir-ch:
Previously, the zero value of the returned integer `info` was used to indicate
success. This is wrong because it is a 0-based index unlike the reference LAPACK
where it is a 1-based index and therefore 0 is available for such purpose. This
commit adds `ok bool` to the returned values from Dlahqr.

### DIFF
--- a/native/dlahqr.go
+++ b/native/dlahqr.go
@@ -13,8 +13,8 @@ import (
 // Dlahqr computes the eigenvalues and Schur factorization of a block of an n×n
 // upper Hessenberg matrix H, using the double-shift/single-shift QR algorithm.
 //
-// h and ldh represent the matrix H. Its contents on return depends on the value
-// of info. Dlahqr works primarily with the Hessenberg submatrix
+// h and ldh represent the matrix H. The contents of h on return depends on the
+// value of converged. Dlahqr works primarily with the Hessenberg submatrix
 // H[ilo:ihi+1,ilo:ihi+1], but applies transformations to all of H if wantt is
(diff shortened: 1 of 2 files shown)


native/dlahqr.go
```diff
@@ -14,7 +14,7 @@ import (
 // upper Hessenberg matrix H, using the double-shift/single-shift QR algorithm.
 //
 // h and ldh represent the matrix H. Its contents on return depends on the value
-// of info. Dlahqr works primarily with the Hessenberg submatrix
+// of ok. Dlahqr works primarily with the Hessenberg submatrix
```
kortschak: Could you add a rough precis of what `ok` actually means. From the list of possible outcomes below, it reads as though some times when `ok` is false, everything is actually OK. Having a conceptual peg to hang things on would help. There are two options, either add here "... of ok[, which indicates whether the factorization converged", or change `ok` to `converged`.

vladimir-ch: Great suggestion, thanks. You are right, everything is actually ok even if `ok` is false. Therefore `converged` is a much better name. 

I renamed the return to `converged` and slightly reordered the paragraphs about `converged`.

vladimir-ch:
PTAL

I'm preparing a PR that tests `Dlahqr` more thoroughly. There should be tests for H with known eigenvalues. What I don't know is how to prepare H for which Dlahqr will return converged=false. I'll have to study something about testing eigenvalue decomposition implementations.

kortschak:
That's nice.

LGTM

PR #168, opened 2016-08-15
native: change Dlahqr back to return single integer

vladimir-ch:
If the returned integer is a 1-based size and not a 0-based index, we can use 0 to indicate success (i.e., all eigenvalues have converged) and thus the `converged bool` is not necessary.

Also polished the docs a little.

PTAL

### DIFF
--- a/native/dlahqr.go
+++ b/native/dlahqr.go
@@ -13,10 +13,9 @@ import (
 // Dlahqr computes the eigenvalues and Schur factorization of a block of an n×n
 // upper Hessenberg matrix H, using the double-shift/single-shift QR algorithm.
 //
-// h and ldh represent the matrix H. The contents of h on return depends on the
-// value of converged. Dlahqr works primarily with the Hessenberg submatrix
-// H[ilo:ihi+1,ilo:ihi+1], but applies transformations to all of H if wantt is
-// true. It is assumed that H[ihi+1:n,ihi+1:n] is already upper
+// h and ldh represent the matrix H. Dlahqr works primarily with the Hessenberg
+// submatrix H[ilo:ihi+1,ilo:ihi+1], but applies transformations to all of H if
+// wantt is true. It is assumed that H[ihi+1:n,ihi+1:n] is already upper
 // quasi-triangular, although this is not checked.
 //
 // It must hold that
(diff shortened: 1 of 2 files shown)


native/dlahqr.go
```diff
@@ -37,37 +36,34 @@ import (
...
-// If converged is false and wantt is true, then on return
+// If unconverged is positive, some eigenvalues have not converged, and
+// wr[unconverged:ihi+1] and wi[unconverged:ihi+1] contain those eigenvalues
+// which have been successfully computed. If wantt is false, on return the
+// remaining unconverged eigenvalues are the eigenvalues of the upper Hessenberg
+// matrix H[ilo:unconverged,ilo:unconverged]. If wantt is true, then on return
 //  (initial H)*U = U*(final H),   (*)
 // where U is an orthogonal matrix. The final H is upper Hessenberg and
-// H[index+1:ihi+1,index+1:ihi+1] is upper triangular.
+// H[unconverged:ihi+1,unconverged:ihi+1] is upper quasi-triangular.
```
vladimir-ch: I think that reference LAPACK means _quasi_-triangular here, triangular does not make much sense.

native/dlahqr.go
```diff
@@ -25,7 +24,7 @@ import (
 //  H[ilo,ilo-1] == 0,  if ilo > 0,
 // otherwise Dlahqr will panic.
 //
-// wr[ilo:ihi+1] and wi[ilo:ihi+1] will contain, on return if converged is true,
+// If unconverged is zero, wr[ilo:ihi+1] and wi[ilo:ihi+1] will contain on return
```
kortschak:
```
// If unconverged is zero on return, wr[ilo:ihi+1] and wi[ilo:ihi+1] will contain
// respectively the real and imaginary parts of the computed eigenvalues ilo
// to ihi.
```

Keep the parse tree a little less branched.

kortschak:
LGTM with minor edit.

native/dlahqr.go
vladimir-ch: Ok, returned to the previous style.

vladimir-ch: Ah, reading the comment in email I missed your suggested edit and misunderstood the message. Will send as a separate PR.

PR #169, opened 2016-08-15
native: polish docs for Dlahqr

vladimir-ch:
Let me know if I should undo here also the splitting of the `If unconverged is ... and wantt is ...` paragraphs.


kortschak:
LGTM

The other paras parse reasonably linearly, so they are fine. I think the main issue was the wait on the "respectively".

PR #172, opened 2016-08-17
native: add Dlaqr2

vladimir-ch:
The documentation of reference DLAQR2 is quite loose or even incorrect in some places, so most of the differences in this PR come from that. Further, some of the documented restrictions on input values are in fact allowed and handled in the reference, I'm stricter and usually panic. That's another source of tiny differences.

Also, it seems that having a set of matrices with known eigenvalues will be soon very useful / necessary for testing. Any suggestions on where to get them, and how and where to store them? Using random matrices and computing their eigenvalues in e.g. Octave is of course possible but I'm after "difficult" eigenvalue problems that would trigger some exceptional paths in the code not taken by random matrices. Matrix Market is an option, their matrices for eigenvalue problems seem to be all sparse which should not be a problem for moderate sizes. Any other options?

### DIFF
--- a/native/dlaqr2.go
+++ b/native/dlaqr2.go
@@ -1,0 +1,382 @@
+// Copyright ©2016 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package native
+
+import (
+	"math"
+
+	"github.com/gonum/blas"
+	"github.com/gonum/blas/blas64"
+	"github.com/gonum/lapack"
+)
+
+// Dlaqr2 performs the orthogonal similarity transformation of a n×n upper
+// Hessenberg matrix to detect and deflate fully converged eigenvalues from a
+// trailing principal submatrix using aggressive early deflation [1].
+//
+// On return, H will be overwritten by a new Hessenberg matrix that is a
+// perturbation of an orthogonal similarity transformation of H. It is to be
+// hoped that on output H will have many zero subdiagonal entries.
+//
+// If wantt is true, the matrix H will be fully updated so that the
+// quasi-triangular Schur factor can be computed. If wantt is false, then only
+// enough of H will be updated to preserve the eigenvalues.
+//
+// If wantz is true, the orthogonal similarity transformation will be
+// accumulated into Z[iloz:ihiz+1,ktop:kbot+1], otherwise Z is not referenced.
+//
+// ktop and kbot determine an block [ktop:kbot+1,ktop:kbot+1] along the diagonal
+// of H. The block must be isolated, that is, it must hold that
+//  ktop == 0   or H[ktop,ktop-1] == 0,  and
+//  kbot == n-1 or H[kbot+1,kbot] == 0,
+// otherwise Dlaqr2 will panic.
+//
+// nw is the deflation window size. It must hold that
+//  1 <= nw <= kbot-ktop+1,
+// otherwise Dlaqr2 will panic.
+//
+// iloz and ihiz specify the rows of the n×n matrix Z to which transformations
+// will be applied if wantz is true. It must hold that
+//  0 <= iloz <= ktop,  and  kbot <= ihiz < n,
+// otherwise Dlaqr2 will panic.
+//
+// sr and si must have length kbot+1, otherwise Dlaqr2 will panic.
+//
+// v and ldv represent an nw×nw work matrix.
+//
+// t and ldt represent an nw×nw work matrix T. nh is the number of columns of T
+// available for workspace. It must hold that nh <= nw, otherwise Dlaqr2 will
+// panic.
+//
+// wv and ldwv represent an nw×nw work matrix W. nv is the number of rows of W
+// available for workspace. It must hold that nv <= nw, otherwise Dlaqr2 will
+// panic.
+//
+// work must have length at least lwork and lwork must be at least max(1,2*nw),
+// otherwise Dlaqr2 will panic. Larger values of lwork may result in greater
+// efficiency. On return, work[0] will contain the optimal value of lwork.
+//
+// If lwork is -1, instead of performing Dlaqr2, the function only estimates the
+// optimal workspace size and stores it into work[0]. Neither h nor z are
+// accessed.
+//
+// On return, ns and nd will contain respectively the number of unconverged
+// (i.e., approximate) eigenvalues and converged eigenvalues that are stored in
+// sr and si.
+//
+// On return, the real and imaginary parts of approximate eigenvalues that may
+// be used for shifts will be stored respectively in sr[kbot-nd-ns+1:kbot-nd+1]
+// and si[kbot-nd-ns+1:kbot-nd+1].
+//
+// On return, the real and imaginary parts of converged eigenvalues will be
+// stored respectively in sr[kbot-nd+1:kbot+1] and si[kbot-nd+1:kbot+1].
+//
+// References:
+//  [1] K. Braman, R. Byers, R. Mathias. The Multishift QR Algorithm. Part II:
+//      Aggressive Early Deflation. SIAM J. Matrix Anal. Appl 23(4) (2002), pp. 948—973
+//      URL: http://dx.doi.org/10.1137/S0895479801384585
+//
+func (impl Implementation) Dlaqr2(wantt, wantz bool, n, ktop, kbot, nw int, h []float64, ldh int, iloz, ihiz int, z []float64, ldz int, sr, si []float64, v []float64, ldv int, nh int, t []float64, ldt int, nv int, wv []float64, ldwv int, work []float64, lwork int) (ns, nd int) {
+	checkMatrix(n, n, h, ldh)
+	switch {
+	case ktop < 0 || ktop > max(0, n-1):
+		panic("lapack: invalid value of ktop")
+	case kbot < min(ktop, n-1) || n <= kbot:
+		panic("lapack: invalid value of kbot")
+	case ktop >= 1 && h[ktop*ldh+ktop-1] != 0:
+		panic("lapack: block not isolated")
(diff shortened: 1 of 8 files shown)


native/dlaqr2.go
```diff
@@ -0,0 +1,382 @@
...
+//  kbot == n-1 or H[kbot+1,kbot] == 0,
+// otherwise Dlaqr2 will panic.
+//
+// nw is the deflation window size. It must hold that
+//  1 <= nw <= kbot-ktop+1,
+// otherwise Dlaqr2 will panic.
+//
+// iloz and ihiz specify the rows of the n×n matrix Z to which transformations
+// will be applied if wantz is true. It must hold that
+//  0 <= iloz <= ktop,  and  kbot <= ihiz < n,
```
vladimir-ch: When n==0, ktop and kbot must have specific values. Is it worth documenting it?

kortschak: Yes, I think so.

native/dlaqr2.go
```diff
@@ -0,0 +1,382 @@
...
+
+import (
+	"math"
+
+	"github.com/gonum/blas"
+	"github.com/gonum/blas/blas64"
+	"github.com/gonum/lapack"
+)
+
+// Dlaqr2 performs the orthogonal similarity transformation of a n×n upper
```
kortschak: s/a/an/

native/dlaqr2.go
```diff
@@ -0,0 +1,382 @@
...
+// Hessenberg matrix to detect and deflate fully converged eigenvalues from a
+// trailing principal submatrix using aggressive early deflation [1].
+//
+// On return, H will be overwritten by a new Hessenberg matrix that is a
+// perturbation of an orthogonal similarity transformation of H. It is to be
```
kortschak: s/to be//

native/dlaqr2.go
```diff
@@ -0,0 +1,382 @@
...
+// hoped that on output H will have many zero subdiagonal entries.
+//
+// If wantt is true, the matrix H will be fully updated so that the
+// quasi-triangular Schur factor can be computed. If wantt is false, then only
+// enough of H will be updated to preserve the eigenvalues.
+//
+// If wantz is true, the orthogonal similarity transformation will be
+// accumulated into Z[iloz:ihiz+1,ktop:kbot+1], otherwise Z is not referenced.
+//
+// ktop and kbot determine an block [ktop:kbot+1,ktop:kbot+1] along the diagonal
```
kortschak: s/an/a/

native/dlaqr2.go
```diff
@@ -0,0 +1,382 @@
...
+// available for workspace. It must hold that nh <= nw, otherwise Dlaqr2 will
+// panic.
+//
+// wv and ldwv represent an nw×nw work matrix W. nv is the number of rows of W
+// available for workspace. It must hold that nv <= nw, otherwise Dlaqr2 will
+// panic.
+//
+// work must have length at least lwork and lwork must be at least max(1,2*nw),
+// otherwise Dlaqr2 will panic. Larger values of lwork may result in greater
+// efficiency. On return, work[0] will contain the optimal value of lwork.
```
kortschak: Is this always, or only when lwork == -1?

kortschak: Yes, it does. Why? Do they do this anywhere else? I don't recall this behaviour for any other function.

kortschak:
Another option for finding useful test matrices would be to use dvyukov's fuzzer. @btracey has used this AFAIK.

native/dlaqr2.go
```diff
@@ -0,0 +1,382 @@
...
+	}
+	if jw >= 3 {
+		t[(jw-1)*ldt+jw-3] = 0
+	}
+
+	ns = jw
+	ilst := infqr
+	// Deflation detection loop.
+	for ilst < ns {
+		var bulge bool
```
kortschak: `bulge := false`

native/dlaqr2.go
```diff
@@ -0,0 +1,382 @@
...
+		if ns >= 2 {
+			bulge = t[(ns-1)*ldt+ns-2] != 0
+		}
+		if !bulge {
+			// Real eigenvalue.
+			abst := math.Abs(t[(ns-1)*ldt+ns-1])
```
kortschak: They actually call this `foo`‽

native/dlaqr2.go
```diff
@@ -0,0 +1,382 @@
...
+
+	ns = jw
+	ilst := infqr
+	// Deflation detection loop.
+	for ilst < ns {
+		var bulge bool
+		if ns >= 2 {
+			bulge = t[(ns-1)*ldt+ns-2] != 0
+		}
+		if !bulge {
```
kortschak: The code duplication here is troubling, but I don't see a way to factor it cleanly. I guess that's why it's like this.

native/dlaqr2.go
```diff
@@ -0,0 +1,382 @@
...
+			work[0] = 1
+			impl.Dlaset(blas.Lower, jw-2, jw-2, 0, 0, t[2*ldt:], ldt)
+			impl.Dlarf(blas.Left, ns, jw, work[:ns], 1, tau, t, ldt, work[jw:])
+			impl.Dlarf(blas.Right, ns, ns, work[:ns], 1, tau, t, ldt, work[jw:])
+			impl.Dlarf(blas.Right, jw, ns, work[:ns], 1, tau, v, ldv, work[jw:])
+			impl.Dgehrd(jw, 0, ns-1, t, ldt, work[:jw-1], work[jw:], lwork-jw)
+		}
+
+		// Copy updated reduced window into place.
+		if kwtop >= 1 {
```
kortschak: `> 0`? I assume this is a 1->0 indexing transformation.

native/dlaqr2.go
```diff
@@ -0,0 +1,382 @@
...
+		}
+		impl.Dlacpy(blas.Upper, jw, jw, t, ldt, h[kwtop*ldh+kwtop:], ldh)
+		bi.Dcopy(jw-1, t[ldt:], ldt+1, h[(kwtop+1)*ldh+kwtop:], ldh+1)
+
+		// Accumulate orthogonal matrix in order to update H and Z, if
+		// requested.
+		if ns > 1 && s != 0 {
+			// work[:ns-1] contains the elementary reflectors stored
+			// by a call to Dgehrd above.
+			impl.Dormhr(blas.Right, blas.NoTrans, jw, ns, 0, ns-1, t, ldt, work[:ns-1], v, ldv, work[jw:], lwork-jw)
```
kortschak: Split this line.

native/dlaqr2.go
```diff
@@ -0,0 +1,382 @@
...
+					k = i + 2
+				}
+			}
+		}
+	}
+
+	// Restore shift/eigenvalue array from T.
+	for i := jw - 1; i >= infqr; {
+		if i == infqr || t[i*ldt+i-1] == 0 {
+			sr[kwtop+i] = t[i*ldt+i]
```
kortschak: The Fortran here is bizarre.

native/dlaqr2.go
vladimir-ch: Yes, they do (not sure why). Inexhaustive search gave me at least Dgehrd, Dorghr, Dormhr, Dorgbr.

native/dlaqr2.go
vladimir-ch: :-)

native/dlaqr2.go
vladimir-ch: Done.

native/dlaqr2.go
vladimir-ch: Done.

native/dlaqr2.go
vladimir-ch: Done.

native/dlaqr2.go
vladimir-ch: Done.

native/dlaqr2.go
vladimir-ch: Done.

native/dlaqr2.go
vladimir-ch: I wanted to include the border value of kwtop in the condition but with `> 0` it seems better. Done.

native/dlaqr2.go
vladimir-ch: Done.

kortschak:
Implementation LGTM, leaving tests for @btracey

vladimir-ch:
Reading through DLAQR4 I've noticed that from two possible interpretations of DLAQR2's documentation about the matrices T and WV I chose the wrong one. I've pushed a commit that should fix it. PTAL @kortschak , just in case.

vladimir-ch:
I've pushed a commit to check input slices in Dlaqr2 only after a workspace query is handled so that it is possible to pass all slices as nil. A workspace query should depend only on the input parameters, the matrices and vectors must not be accessed anyway or may not be available at the call site. Also, often the reference LAPACK uses arbitrary placeholders for arrays and relies on the fact that no checking is done (as is the case of workspace query to DLAQR2 in DLAQR4).

I also updated the test which in turn forced me to make the same change in Dgehrd and Dormhr. I'm sorry for such extensive change but I think it is for a good end.

kortschak:
>Moreover, often the reference LAPACK uses arbitrary placeholders for arrays and relies on the
fact that no checking is done.

Yes, this practice gives me the heebie-jeebies.

kortschak:
I think this last change is a nice one; it makes the intention of the code clearer (maybe NETLIB people could take the opportunity to have a named NULL matrix that is 1x1 and could be used as a place holder to the same effect).

LGTM, but again wait for @btracey.

btracey:
Test LGTM. I didn't check the implementation.

The eigenvalue tests are generally hard to do, because most randomly generated matrices are ill-conditioned. Maybe one could (in a combined test?) generate random eigenvalues and check that they look good to me.

PR #173, opened 2016-08-18
native: add Iparmq

vladimir-ch:


native/iparmq.go
```diff
@@ -0,0 +1,112 @@
...
+// n is the order of the Hessenberg matrix H.
+//
+// ilo and ihi specify the block [ilo:ihi+1,ilo:ihi+1] that is being processed.
+//
+// lwork is the amount of workspace available.
+//
+// Except for ispec input parameters are not checked.
+//
+// Iparmq is an internal routine. It is exported for testing purposes.
+func (Implementation) Iparmq(ispec int, name, _ string, n, ilo, ihi, lwork int) int {
```
kortschak: Will the opts parameter be added later? If not, leave it out now. If it will give it a name.

kortschak: Unless there is another reason I have missed.

native/iparmq.go
```diff
@@ -0,0 +1,112 @@
+// Copyright ©2016 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package native
+
+import "math"
+
+// Iparmq can return problem and machine dependent parameters useful for Dhseqr
```
kortschak: s/can//

kortschak: s/return/returns/

native/iparmq.go
vladimir-ch: `opts` is described as `concatenation of the string arguments to TTQRE`. From `[1]`, TTQRE was an experimental Fortran implementation of the multishift QR algorithm by Braman et al. and `opts` is not used in the reference LAPACK's IPARMQ. This suggests that it could be removed. However, DLAQR4 does assemble a string based on `wantt` and `wantz` so I can imagine that `opts` could be used to tune the returned values somehow (although I have no idea how). With this, let me know if I should remove it or keep it.

native/iparmq.go
vladimir-ch: Done.

native/iparmq.go
kortschak: Keep is and name it. Add a note in the docs that it exists for future use.

vladimir-ch: Done.

native/ilaenv.go
```diff
@@ -373,5 +375,8 @@ func (Implementation) Ilaenv(ispec int, s string, opts string, n1, n2, n3, n4 in
 	case 11:
 		// Go guarantees ieee
 		return 1
+	case 12, 13, 14, 15, 16:
+		// Dhseqr and related functions for eigenvalue problems.
```
btracey: Huh, sorry I missed those.

btracey:
LGTM

Issue #163, opened 2016-08-03
lapack: review type and constant declarations


vladimir-ch:
This is a reminder issue that arose in #160 

In `lapack.go`, maybe some types can be removed/merged, maybe there are better names for the types and constants, ...

vladimir-ch:
Looking at DGEEV, its list of parameters starts with
```
*> \param[in] JOBVL
*> \verbatim
*>          JOBVL is CHARACTER*1
*>          = 'N': left eigenvectors of A are not computed;
*>          = 'V': left eigenvectors of A are computed.
*> \endverbatim
*> \param[in] JOBVR
*> \verbatim
*>          JOBVR is CHARACTER*1
*>          = 'N': right eigenvectors of A are not computed;
*>          = 'V': right eigenvectors of A are computed.
*> \endverbatim
```
Using `lapack.Job` for both parameters seems dangerous:
```
const (
   ComputeLeftEigVec Job = 'V'
   ComputeRightEigVec Job = 'V'
)
```
because they would be interchangeable without the compiler complaining. This led me to believe that we should probably avoid using generic `Job` and `Comp` types in `native` and `cgo` which leads me to ask whether we need to use `lapack.Job` and `lapack.Comp` in the `lapacke` package. If its sole purpose is to serve the `cgo` package, then using just `byte` would be enough and `lapacke` wouldn't need to import `lapack`.

Issue #160, opened 2016-08-01
clapack: ?trexc use wrong type for compq


vladimir-ch:
`genLapack.pl` gives `comp?` named variables always the type `lapack.CompSV`. `?trexc` routines are not related to SVD, so a better type would be `lapack.EigComp`.

kortschak:
Would `lapack.Comp` suffice? It would be much simpler to make that change safely than to try to differentiate all the ED and SVD uses.

vladimir-ch:
I think that `lapack.Comp` is a good solution. Also, we should review all the byte types and consts in lapack before the stability promise.

btracey:
File a reminder issue?

vladimir-ch:
Resolved by #174

PR #174, opened 2016-08-22
lapack,cgo: rename CompSV to Comp

vladimir-ch:
Partly addresses #160 and #163

### DIFF
--- a/lapack.go
+++ b/lapack.go
@@ -10,13 +10,7 @@ const None = 'N'
 
 type Job byte
 
-// CompSV determines if the singular values are to be computed in compact form.
-type CompSV byte
-
-const (
-	Compact  CompSV = 'P'
-	Explicit CompSV = 'I'
-)
+type Comp byte
 
 // Complex128 defines the public complex128 LAPACK API supported by gonum/lapack.
 type Complex128 interface{}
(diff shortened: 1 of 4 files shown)


lapack.go
```diff
@@ -10,13 +10,7 @@ const None = 'N'
 
 type Job byte
 
-// CompSV determines if the singular values are to be computed in compact form.
-type CompSV byte
-
-const (
-	Compact  CompSV = 'P'
```
vladimir-ch: These two were unused.

btracey:
Good thing you filed a reminder :)

LGTM

vladimir-ch:
I'm also thinking of removing `type SVDJob byte` and using just `Job` for the SVD constants, and similarly use `Comp` for `type EigComp byte`. What do you think?

kortschak:
LGTM

Other changes SGTM

btracey:
Me too. It would be nicer to have a name per type, but Lapack plays pretty loose.

vladimir-ch:
> It would be nicer to have a name per type, but Lapack plays pretty loose.

And distinguishing them in `clapack` would be troublesome.

Thanks.

PR #175, opened 2016-08-22
cgo: rename clapack package to lapacke

vladimir-ch:


README.md
```diff
@@ -39,7 +39,7 @@ adjust the `CGO_LDFLAGS` variable:
 
 Defines the LAPACK API based on http://www.netlib.org/lapack/lapacke.html
 
-### lapack/clapack
+### lapack/lapacke
```
vladimir-ch: We should do something about the README, it is lacking. I made this minimal change just to have a place to hang this comment.

kortschak: LGTM

PR #177, opened 2016-08-23
native: fix computation of ldwork in Dormqr

### LINKED ISSUE #176: native: bad computation of work matrix size in Dormqr
`Dormqr` computes the maximum number of columns of a work matrix nwxnb that fits into available workspace. If lwork is not long enough to hold the optimum block size nb, nb is reduced. The code however doesn't update ldwork, leaving it at the previous, unnecessarily large value. If the workspace is long enough to use the blocked code but not long enough to hold the nwxnb matrix, Dlarfb will panic later.
```
	ldwork := nb // nb is the optimal block size as returned from Ilaenv.
	if nb > 1 && nb < k {
		iws := nw * nb
		if lwork < iws {
			nb = lwork / nw
			// ldwork = nb // <--- Uncommenting fixes the issue.
			nbmin = max(2, impl.Ilaenv(2, "DORMQR", opts, m, n, k, -1))
		}
	}
```

Speaking of `Dormqr`, why is `t` allocated in Dormqr and not taken from `work` as in the reference LAPACK, @btracey ?

vladimir-ch:
Fixes #176

### DIFF
--- a/cgo/lapack.go
+++ b/cgo/lapack.go
@@ -1213,36 +1213,34 @@ func (impl Implementation) Dormlq(side blas.Side, trans blas.Transpose, m, n, k 
 // If lwork is -1, instead of performing Dormqr, the optimal workspace size will
 // be stored into work[0].
 func (impl Implementation) Dormqr(side blas.Side, trans blas.Transpose, m, n, k int, a []float64, lda int, tau, c []float64, ldc int, work []float64, lwork int) {
-	left := side == blas.Left
-	if left {
-		checkMatrix(m, k, a, lda)
-	} else {
-		checkMatrix(n, k, a, lda)
+	var nq, nw int
+	switch side {
+	default:
+		panic(badSide)
+	case blas.Left:
+		nq = m
+		nw = n
+	case blas.Right:
+		nq = n
+		nw = m
 	}
-	checkMatrix(m, n, c, ldc)
-
-	if len(tau) < k {
-		panic(badTau)
-	}
-
-	if lwork == -1 {
-		if left {
-			work[0] = float64(n)
-			return
-		}
-		work[0] = float64(m)
-		return
-	}
-	if len(work) < lwork {
+	switch {
+	case trans != blas.NoTrans && trans != blas.Trans:
+		panic(badTrans)
+	case m < 0 || n < 0:
+		panic(negDimension)
+	case k < 0 || nq < k:
+		panic("lapack: invalid value of k")
+	case len(work) < lwork:
+		panic(shortWork)
+	case lwork < max(1, nw) && lwork != -1:
 		panic(badWork)
 	}
-	if left {
-		if lwork < n {
-			panic(badWork)
-		}
-	} else {
-		if lwork < m {
-			panic(badWork)
+	if lwork != -1 {
+		checkMatrix(nq, k, a, lda)
+		checkMatrix(m, n, c, ldc)
+		if len(tau) != k {
+			panic(badTau)
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


kortschak:
Does your test panic if the change is not made?

vladimir-ch:
Without the 032a556 commit, yes.

btracey:
Did you want to fix `t` while you're here?

vladimir-ch:
I hope I have not taken it too far. Here's what I've done:

- I removed the call to make and t is now taken from work slice. The reference is a little wasteful, t is always nbmax x nbmax even if nb is later reduced when non-optimal lwork. Solving a quadratic equation we could find larger nb that fits into the workspace but that would be overkill. Also, I put t at the beginning of work unlike the reference. It's convenient to use already available tsize as the index.

- I cleaned up the code, mainly the input checking part and I copied the checks to cgo. I use the style on which I have settled recently. I updated the test to pass nil for slices when doing workspace query.

- I made the check of tau's length stricter, in this way we don't have to worry about writing past tau[k-1]. The lead to some adjustments in Dgels and Dormbr.

- I updated the docs in native, cgo and lapack64

- Finally, I fixed a bug in how the test was (not) called in cgo.

PTAL

lapack64/lapack64.go
```diff
@@ -293,24 +293,33 @@ func Ormlq(side blas.Side, trans blas.Transpose, a blas64.General, tau []float64
...
+//  Q = H_0 * H_1 * ... * H_{k-1}.
+//
+// If side == blas.Left, A is an m×k matrix and 0 <= k <= m.
+// If side == blas.Right, A is an n×k matrix and 0 <= k <= n.
+// The ith column of A contains the vector which defines the elementary
+// reflector H_i and tau[i] contains its scalar factor. tau must have length k
+// and Ormqr will panic otherwise. Geqrf returns A and tau in the required
+// form.
+//
+// work must have length at least max(1,lwork), and lwork must be at least n if
```
kortschak: This para seems like it should be after/within the following one.

cgo/lapack.go
```diff
@@ -1185,21 +1185,33 @@ func (impl Implementation) Dormlq(side blas.Side, trans blas.Transpose, m, n, k
...
+//  Q = H_0 * H_1 * ... * H_{k-1}.
+//
+// If side == blas.Left, A is an m×k matrix and 0 <= k <= m.
+// If side == blas.Right, A is an n×k matrix and 0 <= k <= n.
+// The ith column of A contains the vector which defines the elementary
+// reflector H_i and tau[i] contains its scalar factor. tau must have length k
+// and Dormqr will panic otherwise. Dgeqrf returns A and tau in the required
+// form.
+//
+// work must have length at least max(1,lwork), and lwork must be at least n if
```
kortschak: Same as for the doc for lapack64.

cgo/lapack_test.go
```diff
@@ -156,7 +156,7 @@ func TestDorghr(t *testing.T) {
 }
 
 func TestDormqr(t *testing.T) {
-	testlapack.Dorm2rTest(t, blockedTranslate{impl})
+	testlapack.DormqrTest(t, blockedTranslate{impl})
```
btracey: This wasn't a typo. The test for Dormqr just compares to the output of Dorm2r. The cgo dorm2r is just a call to Dormqr, so this change just confirms that Dormqr = Dormqr. We want to call Dorm2rTest to actually test that the routine computes a QR factorization.

btracey:
LGTM with the test change reverted and the comments by @kortschak  accepted

PR #179, opened 2016-08-29
native,cgo: add Dgebal

vladimir-ch:
PTAL

### DIFF
--- a/cgo/lapack.go
+++ b/cgo/lapack.go
@@ -24,6 +24,7 @@ const (
 	badIlo          = "lapack: ilo out of range"
 	badIhi          = "lapack: ihi out of range"
 	badIpiv         = "lapack: insufficient permutation length"
+	badJob          = "lapack: bad Job"
 	badLdA          = "lapack: index of a out of range"
 	badNorm         = "lapack: bad norm"
 	badPivot        = "lapack: bad pivot"
--- a/cgo/lapack.go
+++ b/cgo/lapack.go
@@ -227,6 +227,74 @@ func (impl Implementation) Dpotrf(ul blas.Uplo, n int, a []float64, lda int) (ok
 	return lapacke.Dpotrf(ul, n, a, lda)
 }
 
+// Dgebal balances an n×n matrix A. Balancing consists of two stages, permuting
+// and scaling. Both steps are optional and depend on the value of job.
+//
+// Permuting consists of applying a permutation matrix P such that the matrix
+// that results from P^T*A*P takes the upper block triangular form
+//            [ T1  X  Y  ]
+//  P^T A P = [  0  B  Z  ],
+//            [  0  0  T2 ]
+// where T1 and T2 are upper triangular matrices and B contains at least one
+// nonzero off-diagonal element in each row and column. The indices ilo and ihi
+// mark the starting and ending columns of the submatrix B. The eigenvalues of A
+// isolated in the first 0 to ilo-1 and last ihi+1 to n-1 elements on the
+// diagonal can be read off without any roundoff error.
+//
+// Scaling consists of applying a diagonal similarity transformation D such that
+// D^{-1}*B*D has the 1-norm of each row and its corresponding column nearly
+// equal. The output matrix is
+//  [ T1     X*D          Y    ]
+//  [  0  inv(D)*B*D  inv(D)*Z ].
+//  [  0      0           T2   ]
+// Scaling may reduce the 1-norm of the matrix, and improve the accuracy of
+// the computed eigenvalues and/or eigenvectors.
+//
+// job specifies the operations that will be performed on A.
+// If job is lapack.None, Dgebal sets scale[i] = 1 for all i and returns ilo=0, ihi=n-1.
+// If job is lapack.Permute, only permuting will be done.
+// If job is lapack.Scale, only scaling will be done.
+// If job is lapack.PermuteScale, both permuting and scaling will be done.
+//
+// On return, if job is lapack.Permute or lapack.PermuteScale, it will hold that
+//  A[i,j] == 0,   for i > j and j ∈ {0, ..., ilo-1, ihi+1, ..., n-1}.
+// If job is lapack.None or lapack.Scale, or if n == 0, it will hold that
+//  ilo == 0 and ihi == n-1.
+//
+// On return, scale will contain information about the permutations and scaling
+// factors applied to A. If π(j) denotes the index of the column interchanged
+// with column j, and D[j,j] denotes the scaling factor applied to column j,
+// then
+//  scale[j] == π(j),     for j ∈ {0, ..., ilo-1, ihi+1, ..., n-1},
+//           == D[j,j],   for j ∈ {ilo, ..., ihi}.
+// scale must have length equal to n, otherwise Dgebal will panic.
+//
+// Dgebal is an internal routine. It is exported for testing purposes.
+func (impl Implementation) Dgebal(job lapack.Job, n int, a []float64, lda int, scale []float64) (ilo, ihi int) {
+	switch job {
+	default:
+		panic(badJob)
+	case lapack.None, lapack.Permute, lapack.Scale, lapack.PermuteScale:
+	}
+	checkMatrix(n, n, a, lda)
+	if len(scale) != n {
+		panic("lapack: bad length of scale")
+	}
+
+	ilo32 := make([]int32, 1)
+	ihi32 := make([]int32, 1)
+	lapacke.Dgebal(job, n, a, lda, ilo32, ihi32, scale)
+	ilo = int(ilo32[0]) - 1
+	ihi = int(ihi32[0]) - 1
+	for j := 0; j < ilo; j++ {
+		scale[j]--
+	}
+	for j := ihi + 1; j < n; j++ {
+		scale[j]--
+	}
+	return ilo, ihi
+}
+
 // Dbdsqr performs a singular value decomposition of a real n×n bidiagonal matrix.
 //
 // The SVD of the bidiagonal matrix B is
--- a/native/dgebal.go
+++ b/native/dgebal.go
@@ -1,0 +1,228 @@
+// Copyright ©2016 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package native
+
+import (
+	"math"
+
+	"github.com/gonum/blas/blas64"
+	"github.com/gonum/lapack"
+)
+
+// Dgebal balances an n×n matrix A. Balancing consists of two stages, permuting
+// and scaling. Both steps are optional and depend on the value of job.
+//
+// Permuting consists of applying a permutation matrix P such that the matrix
+// that results from P^T*A*P takes the upper block triangular form
+//            [ T1  X  Y  ]
+//  P^T A P = [  0  B  Z  ],
+//            [  0  0  T2 ]
+// where T1 and T2 are upper triangular matrices and B contains at least one
+// nonzero off-diagonal element in each row and column. The indices ilo and ihi
+// mark the starting and ending columns of the submatrix B. The eigenvalues of A
+// isolated in the first 0 to ilo-1 and last ihi+1 to n-1 elements on the
+// diagonal can be read off without any roundoff error.
+//
+// Scaling consists of applying a diagonal similarity transformation D such that
+// D^{-1}*B*D has the 1-norm of each row and its corresponding column nearly
+// equal. The output matrix is
+//  [ T1     X*D          Y    ]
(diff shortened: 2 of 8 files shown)


native/dgebal.go
```diff
@@ -0,0 +1,227 @@
...
+	if job == lapack.Scale {
+		goto Scaling
+	}
+
+	// Permutation to isolate eigenvalues if possible.
+	//
+	// Search for rows isolating an eigenvalue and push them down.
+	for swapped {
+		swapped = false
+	RowLoop:
```
vladimir-ch: Let me know if you see a better way how to disentangle the fortran's goto labyrinth.

native/dgebal.go
```diff
@@ -0,0 +1,227 @@
...
+			// Row i has only zero off-diagonal elements in the
+			// block A[ilo:ihi+1,ilo:ihi+1].
+			scale[ihi] = float64(i)
+			if i != ihi {
+				bi.Dswap(ihi+1, a[i:], lda, a[ihi:], lda)
+				bi.Dswap(n, a[i*lda:], 1, a[ihi*lda:], 1)
+			}
+			ihi--
+			if ihi == 0 {
+				scale[0] = 1
```
vladimir-ch: Fortran here is very subtle. They don't set the scale explicitly, they do one extra iteration with ihi=1 and rely on the fact that that value of ihi is also the right value for scale[ihi].

native/dgebal.go
```diff
@@ -0,0 +1,227 @@
...
+			ra := math.Abs(a[i*lda+ilo+ira])
+
+			// Guard against zero c or r due to underflow.
+			if c == 0 || r == 0 {
+				continue
+			}
+			g := r / sclfac
+			f := 1.0
+			s := c + r
+			for c < g && math.Max(f, math.Max(c, ca)) < sfmax2 && math.Min(r, math.Min(g, ra)) > sfmin2 {
```
vladimir-ch: What about the maintainability of the reference fortran here? :-)

native/dgebal.go
kortschak: This seems pretty reasonable, though I'd prefer just rows and columns for label names.

native/dgebal.go
```diff
@@ -0,0 +1,227 @@
...
+			scale[i] = 1
+		}
+		return ilo, ihi
+	}
+
+	bi := blas64.Implementation()
+	swapped := true
+
+	if job == lapack.Scale {
+		goto Scaling
```
kortschak: Lower case.

testlapack/dgebal.go
```diff
@@ -0,0 +1,168 @@
...
+type Dgebaler interface {
+	Dgebal(job lapack.Job, n int, a []float64, lda int, scale []float64) (int, int)
+}
+
+func DgebalTest(t *testing.T, impl Dgebaler) {
+	rnd := rand.New(rand.NewSource(1))
+
+	for _, job := range []lapack.Job{lapack.None, lapack.Permute, lapack.Scale, lapack.PermuteScale} {
+		for _, n := range []int{0, 1, 2, 3, 4, 5, 6, 10, 18, 31, 53, 100} {
+			p := math.Ceil(0.02*float64(n)) / float64(n)
```
vladimir-ch: This amount of nonzeros gives reasonable (not only 0,n-1) values of ilo,ihi across the matrix sizes.

native/dgebal.go
kortschak: Interestingly this is less bad for me than the previous case; I learned to program with BASIC in the early 80s.

vladimir-ch: No no, definitely worse :-) The `GO TO 170` is attached to the condition like in the previous case _and_ the alignment of `IF` on line 336 is as if the body belonged to `IF` on line 334.

kortschak: Yeah, I got that, but having an expressionish next to a conditional beats this IMHO.

native/dgebal.go
```diff
@@ -0,0 +1,227 @@
...
+	if len(scale) != n {
+		panic("lapack: bad length of scale")
+	}
+
+	ilo = 0
+	ihi = n - 1
+
+	if n == 0 || job == lapack.None {
+		for i := range scale {
+			scale[i] = 1
```
kortschak: Is this scale needed for the n == 0 case?

kortschak: Make the range `scale[:n]`?

native/dgebal.go
```diff
@@ -0,0 +1,227 @@
...
+	for i := ilo; i <= ihi; i++ {
+		scale[i] = 1
+	}
+
+	if job == lapack.Permute {
+		return ilo, ihi
+	}
+
+	// Balance the submatrix in rows ilo to ihi.
+
```
kortschak: Delete blank line.

native/dgebal.go
```diff
@@ -0,0 +1,227 @@
...
+		// norms that is considered non-negligible. It must be less than 1.
+		factor = 0.95
+	)
+	sfmin1 := dlamchS / dlamchP
+	sfmax1 := 1 / sfmin1
+	sfmin2 := sfmin1 * sclfac
+	sfmax2 := 1 / sfmin2
+
+	// Iterative loop for norm reduction.
+	conv := false
```
kortschak: `var conv bool`.

native/dgebal.go
vladimir-ch: Changed to rows and columns. Good suggestion, `continue rows` reads nicely.

native/dgebal.go
vladimir-ch: Done.

native/dgebal.go
vladimir-ch: Comment added. I can split this into two `if`s if you want, I just didn't want to waste too much space with this uninteresting case.

native/dgebal.go
vladimir-ch: I like to separate comments by a blank line if they relate to a larger piece of code below, not only to the subsequent block. I'll keep it here.

native/dgebal.go
vladimir-ch: Done.

vladimir-ch:
PTAL

native/dgebal.go
kortschak: Yeah, agreed.

cgo/lapack.go
```diff
@@ -226,6 +227,73 @@ func (impl Implementation) Dpotrf(ul blas.Uplo, n int, a []float64, lda int) (ok
 	return lapacke.Dpotrf(ul, n, a, lda)
 }
 
+// Dgebal balances an n×n matrix A. Balancing consists of two stages, permuting
+// and scaling. Both steps are optional and depend on the value of job.
+//
+// Permuting consists of applying a permutation matrix P such that P^T*A*P
```
btracey: the matrix that results from P^T A P

cgo/lapack.go
```diff
@@ -226,6 +227,73 @@ func (impl Implementation) Dpotrf(ul blas.Uplo, n int, a []float64, lda int) (ok
...
+// the computed eigenvalues and/or eigenvectors.
+//
+// job specifies the operations that will be performed on A.
+// If job is lapack.None, Dgebal sets scale[i] = 1 for all i and returns ilo=0, ihi=n-1.
+// If job is lapack.Permute, only permuting will be done.
+// If job is lapack.Scale, only scaling will be done.
+// If job is lapack.PermuteScale, both permuting and scaling will be done.
+//
+// On return, if job is lapack.Permute or lapack.PermuteScale, it will hold that
+//  A[i,j] == 0,   for i > j and j∈{0, ..., ilo-1, ihi+1, ..., n-1}.
```
btracey: spaces around ∈?

cgo/lapack.go
```diff
@@ -226,6 +227,73 @@ func (impl Implementation) Dpotrf(ul blas.Uplo, n int, a []float64, lda int) (ok
...
+// If job is lapack.None or lapack.Scale, or if n == 0, it will hold that
+//  ilo == 0 and ihi == n-1.
+//
+// On return, scale will contain information about the permutations and scaling
+// factors applied to A. If π(j) denotes the index of the column interchanged
+// with column j, and D[j,j] denotes the scaling factor applied to column j,
+// then
+//  scale[j] == π(j),     for j∈{0, ..., ilo-1, ihi+1, ..., n-1},
```
btracey: same as above.

native/dgebal.go
```diff
@@ -0,0 +1,228 @@
...
+			if c == 0 || r == 0 {
+				continue
+			}
+			g := r / sclfac
+			f := 1.0
+			s := c + r
+			for c < g && math.Max(f, math.Max(c, ca)) < sfmax2 && math.Min(r, math.Min(g, ra)) > sfmin2 {
+				if math.IsNaN(c + f + ca + r + g + ra) {
+					// Panic if NaN to avoid infinite loop.
+					panic("lapack: NaN")
```
btracey: Are we sure this should panic? NaNs can enter matrices in a lot of ways.

testlapack/dgebal.go
btracey: call it pZero. Should there be some check for the n == 1 case?

testlapack/dgebal.go
```diff
@@ -0,0 +1,168 @@
...
+			blas64.Swap(n,
+				blas64.Vector{p.Stride, p.Data[j:]},
+				blas64.Vector{p.Stride, p.Data[int(scale[j]):]})
+		}
+		for j := 0; j < ilo; j++ {
+			blas64.Swap(n,
+				blas64.Vector{p.Stride, p.Data[j:]},
+				blas64.Vector{p.Stride, p.Data[int(scale[j]):]})
+		}
+		// Compute P^T A P.
```
btracey: Add a "checked below"

btracey: Actually, sorry, probably "Compute P^T * A * P and store into want"

testlapack/dgebal.go
```diff
@@ -0,0 +1,168 @@
...
+		ap := eye(n, n)
+		blas64.Gemm(blas.NoTrans, blas.NoTrans, 1, want, p, 0, ap)
+		blas64.Gemm(blas.Trans, blas.NoTrans, 1, p, ap, 0, want)
+	}
+	if job == lapack.Scale || job == lapack.PermuteScale {
+		d := eye(n, n)
```
btracey: "// modify want by D"

testlapack/dgebal.go
```diff
@@ -0,0 +1,168 @@
...
+		dinv := eye(n, n)
+		for i := ilo; i <= ihi; i++ {
+			d.Data[i*d.Stride+i] = scale[i]
+			dinv.Data[i*dinv.Stride+i] = 1 / scale[i]
+		}
+		ad := eye(n, n)
```
btracey: why `eye` and not zeros?

testlapack/general.go
```diff
@@ -986,3 +986,29 @@ func isUpperHessenberg(h blas64.General) bool {
 	}
 	return true
 }
+
+func randomSparseGeneral(r, c, stride int, p float64, rnd *rand.Rand) blas64.General {
```
btracey: Add a comment that marks p is the probability of a zero, and probably make some comment about it being unstructured sparse.

btracey:
LGTM with the minor fixes.

btracey:
(but wait for @kortschak )

native/dgebal.go
kortschak: xerbla calls stop, which is pretty much equivalent.

kortschak:
LGTM after comments addressed.

testlapack/dgebal.go
vladimir-ch: Calling it `pNonzero`, it's the probability of an element being nonzero. Also added a check for n == 0 and n == 1 (pNonzero = 0.5).

cgo/lapack.go

cgo/lapack.go
vladimir-ch: Added.

testlapack/dgebal.go
vladimir-ch: Done.

testlapack/dgebal.go
vladimir-ch: Added `// Modify want by D and D^{-1}.`

testlapack/dgebal.go
vladimir-ch: Because we had no `zeros` and `eye` was the simplest way to get a matrix to work with. I added `zeros` and use it here.

testlapack/general.go
vladimir-ch: Comment added.

vladimir-ch:
Thanks for the review.

PR #180, opened 2016-08-29
native: change Dlaqr2 into Dlaqr23

vladimir-ch:


testlapack/dlaqr23.go
```diff
@@ -27,20 +27,23 @@ type dlaqr2Test struct {
...
-						if ktop > kbot {
-							ktop, kbot = kbot, ktop
+			for _, n := range []int{1, 2, 3, 4, 5, 6, 10, 18, 31, 100} {
+				for _, extra := range []int{0, 11} {
+					for cas := 0; cas < 30; cas++ {
+						var nw int
+						if nw <= 75 {
+							nw = rnd.Intn(n) + 1
+						} else {
+							nw = 76 + rnd.Intn(n-75)
```
vladimir-ch: I changed the logic here to make sure that for sufficiently large matrices we generate large enough window to trigger the Dlaqr4 path. For small matrices any window size works because they will always use Dlahqr inside Dlaqr23.

native/dlaqr23.go
```diff
@@ -98,6 +101,9 @@ func (impl Implementation) Dlaqr2(wantt, wantz bool, n, ktop, kbot, nw int, h []
 			panic("lapack: invalid value of ihiz")
 		}
 	}
+	if recurse < 0 {
```
kortschak: s/recurse/recur/g

Does it need to be an int? Can't it be a bool?

vladimir-ch: > Does it need to be an int? Can't it be a bool?

I wasn't sure here. As far as LAPACK is concerned, it can be a bool because it (Dlaqr04) would be called only with 1 from Dhseqr. I thought that someone might want to play with higher levels of recursion although it has been reported that no significant gain comes from it, so this use case is rather unlikely. Let me know and I can change it.

kortschak: OK. That's fine. Leave it as is.

kortschak:
Just checking that I'm not missing anything in thinking that there is no case where Dlaqr3 is called yet.

LGTM after the [English is fixed](https://books.google.com/ngrams/graph?content=recur%2Crecursive%2Crecurse&year_start=1800&year_end=2000&corpus=15&smoothing=3&share=&direct_url=t1%3B%2Crecur%3B%2Cc0%3B.t1%3B%2Crecursive%3B%2Cc0%3B.t1%3B%2Crecurse%3B%2Cc0).

vladimir-ch:
Dlaqr3 would be called by Dlaqr0 which will come next dressed as Dlaqr04.

> ... after the English is fixed.

Ten out of ten for creativity, but minus several million for good grammar. Fixed.

kortschak:
It's reasonably commonly used, but we don't need to encourage that.

PR #181, opened 2016-08-31
native,testlapack: change Dlaqr4 into Dlaqr04

vladimir-ch:
PTAL

### DIFF
--- a/testlapack/dlaqr04.go
+++ b/testlapack/dlaqr04.go
@@ -14,13 +14,13 @@ import (
 	"github.com/gonum/blas/blas64"
 )
 
-type Dlaqr4er interface {
-	Dlaqr4(wantt, wantz bool, n, ilo, ihi int, h []float64, ldh int, wr, wi []float64, iloz, ihiz int, z []float64, ldz int, work []float64, lwork int) int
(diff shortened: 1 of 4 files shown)


testlapack/dlaqr04.go
```diff
@@ -29,52 +29,73 @@ type dlaqr4Test struct {
...
-						testDlaqr4(t, impl, test, false)
-						testDlaqr4(t, impl, test, true)
 					}
 				}
 			}
 		}
 	}
+
+	// Tests for matrices large enough to possibly use the recursion (but it
+	// doesn't seem to be the case).
```
kortschak: ‽

kortschak:
LGTM

testlapack/dlaqr04.go
vladimir-ch: The recursion takes place only under certain conditions which are not satisfied with the random matrices that are generated. I'll keep this in mind and check the coverage when the whole eigendecomposition solver is finished. Then we'll be able to test it with some standard (hard) test matrices, now we would have to transform them into upper Hessenberg form which would be too troublesome.

PR #182, opened 2016-08-31
native,cgo: add Dgebak

vladimir-ch:
PTAL

### DIFF
--- a/native/dgebak.go
+++ b/native/dgebak.go
@@ -1,0 +1,79 @@
+// Copyright ©2016 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package native
+
+import (
+	"github.com/gonum/blas"
+	"github.com/gonum/blas/blas64"
+	"github.com/gonum/lapack"
+)
+
+// Dgebak updates an n×m matrix V as
+//  V = P D V,        if side == blas.Right,
+//  V = P D^{-1} V,   if side == blas.Left,
+// where P and D are n×n permutation and scaling matrices, respectively,
+// implicitly represented by job, scale, ilo and ihi as returned by Dgebal.
+//
+// Typically, columns of the matrix V contain the right or left (determined by
+// side) eigenvectors of the balanced matrix output by Dgebal, and Dgebak forms
+// the eigenvectors of the original matrix.
+//
+// Dgebak is an internal routine. It is exported for testing purposes.
+func (impl Implementation) Dgebak(job lapack.Job, side blas.Side, n, ilo, ihi int, scale []float64, m int, v []float64, ldv int) {
+	switch job {
+	default:
+		panic(badJob)
+	case lapack.None, lapack.Permute, lapack.Scale, lapack.PermuteScale:
+	}
+	switch side {
+	default:
+		panic(badSide)
+	case blas.Left, blas.Right:
+	}
+	checkMatrix(n, m, v, ldv)
+	switch {
+	case ilo < 0 || max(0, n-1) < ilo:
+		panic(badIlo)
+	case ihi < min(ilo, n-1) || n <= ihi:
+		panic(badIhi)
+	}
+
+	// Quick return if possible.
+	if n == 0 || m == 0 || job == lapack.None {
+		return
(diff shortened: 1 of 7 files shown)


native/dgebak.go
```diff
@@ -0,0 +1,79 @@
...
+package native
+
+import (
+	"github.com/gonum/blas"
+	"github.com/gonum/blas/blas64"
+	"github.com/gonum/lapack"
+)
+
+// Dgebak transforms an n×m matrix V as
+//  V = P D V,        if side == blas.Right,
```
vladimir-ch: Let me know if I should write as an implementation comment why these two formulas.

native/dgebak.go
```diff
@@ -0,0 +1,79 @@
...
+			for i := ilo; i <= ihi; i++ {
+				bi.Dscal(m, 1/scale[i], v[i*ldv:], 1)
+			}
+		}
+	}
+	if job == lapack.Scale {
+		return
+	}
+	// Backward permutation.
+	for i := ilo - 1; i >= 0; i-- {
```
vladimir-ch: I don't know why Lapack distinguishes between sides. The relations in the comment show it and the lapack code confirms it that the permutations are done identically for both sides.

testlapack/dgebal.go
```diff
@@ -24,15 +23,9 @@ func DgebalTest(t *testing.T, impl Dgebaler) {
 
 	for _, job := range []lapack.Job{lapack.None, lapack.Permute, lapack.Scale, lapack.PermuteScale} {
 		for _, n := range []int{0, 1, 2, 3, 4, 5, 6, 10, 18, 31, 53, 100} {
-			var pNonzero float64
```
vladimir-ch: Oops, this somehow leaked into this PR. Sorry. Anyway, specifying the number of non-zeros is much better than the probability.

testlapack/general.go
```diff
@@ -1016,16 +1016,12 @@ func isUpperTriangular(a blas64.General) bool {
...
-			} else {
-				a.Data[i*stride+j] = 0
-			}
-		}
+func randomSparseGeneral(m, n, stride int, nonzeros int, rnd *rand.Rand) blas64.General {
+	a := zeros(m, n, stride)
+	for k := 0; k < nonzeros; k++ {
+		i := rnd.Intn(n)
+		j := rnd.Intn(n)
+		a.Data[i*stride+j] = float64(j+1) * rnd.NormFloat64()
```
vladimir-ch: Multiplying by column index unbalances the matrix so that the scales are not all 1.

native/dgebak.go
```diff
@@ -0,0 +1,79 @@
...
+		} else {
+			for i := ilo; i <= ihi; i++ {
+				bi.Dscal(m, 1/scale[i], v[i*ldv:], 1)
+			}
+		}
+	}
+	if job == lapack.Scale {
+		return
+	}
+	// Backward permutation.
```
kortschak: The Fortran is amazing here. They have a comment, which is Fortran to explain the Fortran.

kortschak:
LGTM, but wait for @btracey.

native/dgebak.go
btracey: Can you use "updates"? I was thinking "transform" was some kind of eigensolution (finds the vector V such that...)

testlapack/general.go
btracey: The function name is that it is quite general, but it seems you are tailoring it for a specific use case. That's fine, but the name should adjust accordingly. Otherwise, shouldn't this be something random instead, rather than forcing the columns to increase in scale?

btracey:
LGTM with the few updates.

native/dgebak.go
vladimir-ch: Agreed, `updates` is better. Done.

testlapack/general.go
vladimir-ch: One doesn't get away with anything here :-) Renamed to `unbalancedSparseGeneral`.

Now that I specify the number of nonzeros the unbalancing is not strictly necessary, the row and column norms differ enough just by absence or presence of zeros. I help it a bit by multiplying with row or column index randomly.

PR #184, opened 2016-09-06
cgo/lapacke: make *hseqr return int

### LINKED ISSUE #183: lapacke: Dhseqr discards necessary information by returning bool instead of int

kortschak:

Fixes #183.


vladimir-ch:
LGTM

PR #185, opened 2016-09-06
native,cgo: add Dhseqr

vladimir-ch:


lapack.go
```diff
@@ -127,3 +127,12 @@ const (
 	Scale        Job = 'S'
 	PermuteScale Job = 'B'
 )
+
+// Jobs and Comps for Dhseqr.
+const (
+	EigenvaluesOnly     Job = 'E'
```
vladimir-ch: Better(=shorter) names here?

It seems that there arises some conflicting, or at least confusing, naming here. There is
`EigValueOnly EigComp = 'N'`
above from symmetric eigenvalue problems, here we have
`EigenvaluesOnly Job = 'E'` ...

native/dhseqr.go
```diff
@@ -0,0 +1,255 @@
...
+
+	if n > nmin {
+		// Dlaqr0 for big matrices.
+		unconverged = impl.Dlaqr04(wantt, wantz, n, ilo, ihi, h, ldh, wr[:ihi+1], wi[:ihi+1],
+			ilo, ihi, z, ldz, work, lwork, 1)
+	} else {
+		// Dlahqr for small matrices.
+		unconverged = impl.Dlahqr(wantt, wantz, n, ilo, ihi, h, ldh, wr[:ihi+1], wi[:ihi+1],
+			ilo, ihi, z, ldz)
+		if unconverged > 0 {
```
vladimir-ch: I hope that when `Dgeev` is finished, some test matrix (for instance, from Matrix Market) will actually trigger this failure. At the moment this branch is not covered by the tests.

native/dhseqr.go
```diff
@@ -0,0 +1,255 @@
...
+
+	// Quick return in case of a workspace query.
+	impl.Dlaqr04(wantt, wantz, n, ilo, ihi, nil, 0, nil, nil, ilo, ihi, nil, 0, work, -1, 1)
+	lwkopt := max(max(1, n), int(work[0]))
+	if lwork == -1 {
+		work[0] = float64(lwkopt)
+		return 0
+	}
+
+	// Copy eigenvalues isolaged by Dgebal.
```
kortschak: isolated

native/dhseqr.go
```diff
@@ -0,0 +1,255 @@
...
+	)
+
+	// Quick return if possible.
+	if n == 0 {
+		work[0] = 1
+		return 0
+	}
+
+	// Quick return in case of a workspace query.
+	impl.Dlaqr04(wantt, wantz, n, ilo, ihi, nil, 0, nil, nil, ilo, ihi, nil, 0, work, -1, 1)
```
kortschak: Move these into the condition below and use `work[0] = max(max(1, n), int(work[0])) `when you get to the bottom?

testlapack/dlahqr.go
```diff
@@ -312,11 +312,11 @@ func testDlahqr(t *testing.T, impl Dlahqrer, test dlahqrTest) {
 		}
 	}
 
-	// Check that wr and wi have not been modified outside [start:ihi+1].
```
kortschak: Can you give a brief explanation of the reason for these changes?

native/dhseqr.go
vladimir-ch: Thanks. Done.

native/dhseqr.go
vladimir-ch: I would like to clean up and make consistent these workspace queries but the change you suggest is not possible, I think, because work may be overwritten below. Of course, another question is why lapack wants to return the optimal workspace size always, not only when doing a query.

testlapack/dlahqr.go
vladimir-ch: Lapack's documentation is not consistent in places, so some routines document wr and wi to be of length `n`, some only `ihi+1`. I noticed this as I propagated upwards in the dependency tree and things had to be made to work together. Dhseqr is the first routine that writes into the whole wr and wi, Dlahqr and DlaqrX operate only on `[ilo:ihi+1]`, so recently I changed Dlahqr (and maybe others) to require wr and wi have length exactly `ihi+1`. I changed the documentation and the routine but forgot to remove the checking of the (empty) slice after `ihi`. I included it here to avoid juggling too many branches or forgetting about it.

native/dhseqr.go
kortschak: The Fortran work the way I suggested, which is why I suggested it. It should work because `work` is always mutated by `Dlaqr04` in this function, thus always leaving `work[0]` with the optimum `lwork`.

kortschak: The change that would be necessary to make the approach work would be to add `work[0] = workl[0]` after https://github.com/gonum/lapack/pull/185/files#diff-b371c19c3285421b26fa5f3dc61b3747R241.

vladimir-ch: Oh, I see what you mean. Will do.

vladimir-ch: Done. PTAL

testlapack/dlahqr.go
kortschak: Thanks.

kortschak:
LGTM

PR #186, opened 2016-09-13
native: add Dtrevc3

vladimir-ch:
For some reason this PR has been extremely onerous. Sometimes one ends up wishing the left eigenvectors didn't exist. On the other hand, cries of hosanna for slice bounds checking and our strict panics.

The tests seem solid and give me some confidence, but I still feel some tripidation. Please review carefully.

### DIFF
--- a/lapack.go
+++ b/lapack.go
@@ -136,3 +136,23 @@ const (
 	InitZ   Comp = 'I'
 	UpdateZ Comp = 'V'
 )
+
+// EigVecSide specifies what eigenvectors will be computed.
+type EigVecSide byte
+
+// EigVecSide constants for Dtrevc3.
+const (
+	RightEigVec     EigVecSide = 'R' // Compute right eigenvectors only.
+	LeftEigVec      EigVecSide = 'L' // Compute left eigenvectors only.
+	RightLeftEigVec EigVecSide = 'B' // Compute both right and left eigenvectors.
+)
+
+// HowMany specifies which eigenvectors will be computed.
+type HowMany byte
+
+// HowMany constants for Dhseqr.
+const (
+	AllEigVec      HowMany = 'A' // Compute all right and/or left eigenvectors.
+	AllEigVecMulQ  HowMany = 'B' // Compute all right and/or left eigenvectors multiplied by an input matrix.
+	SelectedEigVec HowMany = 'S' // Compute selected right and/or left eigenvectors.
+)
--- a/native/dtrevc3.go
+++ b/native/dtrevc3.go
@@ -1,0 +1,866 @@
+// Copyright ©2016 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package native
+
+import (
+	"math"
+
+	"github.com/gonum/blas"
+	"github.com/gonum/blas/blas64"
+	"github.com/gonum/lapack"
+)
+
+// Dtrevc3 computes some or all of the right and/or left eigenvectors of an n×n
+// upper quasi-triangular matrix T in Schur canonical form. Matrices of this
+// type are produced by the Schur factorization of a real general matrix A
+//  A = Q T Q^T,
+// as computed by Dhseqr.
+//
+// The right eigenvector x of T corresponding to an
+// eigenvalue λ is defined by
+//  T x = λ x,
+// and the left eigenvector is defined by
+//  y^H T = λ y^H,
+// where y^H is the conjugate transpose of y.
+//
+// The eigenvalues are read directly from the diagonal blocks of T.
+//
+// This routine returns the matrices X and/or Y of right and left eigenvectors
+// of T, or the products Q*X and/or Q*Y, where Q is an input matrix. If Q is the
+// orthogonal factor that reduces a matrix A to Schur form T, then Q*X and Q*Y
+// are the matrices of right and left eigenvectors of A.
+//
+// If side == lapack.RightEigVec, only right eigenvectors will be computed.
+// If side == lapack.LeftEigVec, only left eigenvectors will be computed.
+// If side == lapack.RightLeftEigVec, both right and left eigenvectors will be computed.
+// For other values of side, Dtrevc3 will panic.
+//
+// If howmny == lapack.AllEigVec, all right and/or left eigenvectors will be
+// computed.
+// If howmny == lapack.AllEigVecMulQ, all right and/or left eigenvectors will be
+// computed and multiplied from left by the matrices in VR and/or VL.
+// If howmny == lapack.SelectedEigVec, right and/or left eigenvectors will be
+// computed as indicated by selected.
+// For other values of howmny, Dtrevc3 will panic.
+//
+// selected specifies which eigenvectors will be computed. It must have length n
+// if howmny == lapack.SelectedEigVec, and it is not referenced otherwise.
+// If w_j is a real eigenvalue, the corresponding real eigenvector will be
+// computed if selected[j] is true.
+// If w_j and w_{j+1} are the real and imaginary parts of a complex eigenvalue,
+// the corresponding complex eigenvector is computed if either selected[j] or
+// selected[j+1] is true, and on return selected[j] will be set to true and
+// selected[j+1] will be set to false.
+//
+// VL and VR are n×mm matrices. If howmny is lapack.AllEigVec or
+// lapack.AllEigVecMulQ, mm must be at least n. If howmny ==
+// lapack.SelectedEigVec, mm must be large enough to store the selected
+// eigenvectors. Each selected real eigenvector occupies one column and each
+// selected complex eigenvector occupies two columns. If mm is not sufficiently
+// large, Dtrevc3 will panic.
+//
+// On entry, if howmny == lapack.AllEigVecMulQ, it is assumed that VL (if side
+// is lapack.LeftEigVec or lapack.RightLeftEigVec) contains an n×n matrix QL,
+// and that VR (if side is lapack.LeftEigVec or lapack.RightLeftEigVec) contains
+// an n×n matrix QR. QL and QR are typically the orthogonal matrix Q of Schur
+// vectors returned by Dhseqr.
+//
+// On return, if side is lapack.LeftEigVec or lapack.RightLeftEigVec,
+// VL will contain:
+//  if howmny == lapack.AllEigVec,      the matrix Y of left eigenvectors of T,
+//  if howmny == lapack.AllEigVecMulQ,  the matrix Q*Y,
+//  if howmny == lapack.SelectedEigVec, the left eigenvectors of T specified by
+//                                      selected, stored consecutively in
+//                                      the columns of VL, in the same order
+//                                      as their eigenvalues.
+// VL is not referenced if side == lapack.RightEigVec.
+//
+// On return, if side is lapack.RightEigVec or lapack.RightLeftEigVec,
+// VR will contain:
+//  if howmny == lapack.AllEigVec,      the matrix X of right eigenvectors of T,
+//  if howmny == lapack.AllEigVecMulQ,  the matrix Q*X,
+//  if howmny == lapack.SelectedEigVec, the left eigenvectors of T specified by
+//                                      selected, stored consecutively in
+//                                      the columns of VR, in the same order
+//                                      as their eigenvalues.
+// VR is not referenced if side == lapack.LeftEigVec.
+//
+// Complex eigenvectors corresponding to a complex eigenvalue are stored in VL
+// and VR in two consecutive columns, the first holding the real part, and the
+// second the imaginary part.
+//
+// Each eigenvector will be normalized so that the element of largest magnitude
+// has magnitude 1. Here the magnitude of a complex number (x,y) is taken to be
+// |x| + |y|.
+//
+// work must have length at least lwork and lwork must be at least max(1,3*n),
+// otherwise Dtrevc3 will panic. For optimum performance, lwork should be at
+// least n+2*n*nb, where nb is the optimal blocksize.
+//
+// If lwork == -1, instead of performing Dtrevc3, the function only estimates
+// the optimal workspace size based on n and stores it into work[0].
+//
+// Dtrevc3 returns the number of columns in VL and/or VR actually used to store
+// the eigenvectors.
+//
+// Dtrevc3 is an internal routine. It is exported for testing purposes.
+func (impl Implementation) Dtrevc3(side lapack.EigVecSide, howmny lapack.HowMany, selected []bool, n int, t []float64, ldt int, vl []float64, ldvl int, vr []float64, ldvr int, mm int, work []float64, lwork int) (m int) {
+	switch side {
+	default:
+		panic(badSide)
+	case lapack.RightEigVec, lapack.LeftEigVec, lapack.RightLeftEigVec:
+	}
+	switch howmny {
+	default:
+		panic(badHowMany)
+	case lapack.AllEigVec, lapack.AllEigVecMulQ, lapack.SelectedEigVec:
+	}
+	switch {
+	case n < 0:
+		panic(nLT0)
+	case len(work) < lwork:
+		panic(shortWork)
+	case lwork < max(1, 3*n) && lwork != -1:
+		panic(badWork)
+	}
+	if lwork != -1 {
+		if howmny == lapack.SelectedEigVec {
+			if len(selected) != n {
+				panic("lapack: bad selected length")
+			}
+			// Set m to the number of columns required to store the
+			// selected eigenvectors, and standardize the slice
+			// selected.
+			for j := 0; j < n; {
+				if j == n-1 || t[(j+1)*ldt+j] == 0 {
+					// Diagonal 1×1 block corresponding to a
+					// real eigenvalue.
+					if selected[j] {
+						m++
+					}
+					j++
+				} else {
+					// Diagonal 2×2 block corresponding to a
+					// complex eigenvalue.
+					if selected[j] || selected[j+1] {
+						selected[j] = true
+						selected[j+1] = false
+						m += 2
+					}
+					j += 2
+				}
+			}
+		} else {
+			m = n
+		}
+		if m > mm {
+			panic("lapack: insufficient number of columns")
+		}
+		checkMatrix(n, n, t, ldt)
+		if (side == lapack.RightEigVec || side == lapack.RightLeftEigVec) && m > 0 {
+			checkMatrix(n, m, vr, ldvr)
+		}
+		if (side == lapack.LeftEigVec || side == lapack.RightLeftEigVec) && m > 0 {
+			checkMatrix(n, m, vl, ldvl)
+		}
+	}
+
+	// Quick return if possible.
+	if n == 0 {
+		work[0] = 1
+		return m
+	}
+
+	const (
+		nbmin = 8
+		nbmax = 128
+	)
+	nb := impl.Ilaenv(1, "DTREVC", string(side)+string(howmny), n, -1, -1, -1)
+
+	// Quick return in case of a workspace query.
+	if lwork == -1 {
+		work[0] = float64(n + 2*n*nb)
+		return m
+	}
+
+	// Use blocked version of back-transformation if sufficient workspace.
+	// Zero-out the workspace to avoid potential NaN propagation.
+	if howmny == lapack.AllEigVecMulQ && lwork >= n+2*n*nbmin {
+		nb = min((lwork-n)/(2*n), nbmax)
+		impl.Dlaset(blas.All, n, 1+2*nb, 0, 0, work[:n+2*nb*n], 1+2*nb)
+	} else {
+		nb = 1
+	}
+
+	// Set the constants to control overflow.
+	ulp := dlamchP
+	smlnum := float64(n) / ulp * dlamchS
+	bignum := (1 - ulp) / smlnum
+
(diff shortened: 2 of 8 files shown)


lapack.go
```diff
@@ -136,3 +136,23 @@ const (
 	InitZ   Comp = 'I'
 	UpdateZ Comp = 'V'
 )
+
+// EigVecSide specifies what eigenvectors will be computed.
+type EigVecSide byte
```
vladimir-ch: Already settling on the names of this type, the one below, and of the constant values was a struggle. `blas.Side` cannot be used because it doesn't cover the `'B'` value and also the semantics is different. I'm very open to other suggestions.

lapack.go
```diff
@@ -136,3 +136,23 @@ const (
...
+	RightLeftEigVec EigVecSide = 'B' // Compute both right and left eigenvectors.
+)
+
+// HowMany specifies which eigenvectors will be computed.
+type HowMany byte
+
+// HowMany constants for Dhseqr.
+const (
+	AllEigVec      HowMany = 'A' // Compute all right and/or left eigenvectors.
+	AllEigVecMulQ  HowMany = 'B' // Compute all right and/or left eigenvectors multiplied by an input matrix.
```
vladimir-ch: This one's been the worst. Not ideal but cannot think of anything better.

native/dtrevc3.go
```diff
@@ -0,0 +1,861 @@
...
+// upper quasi-triangular matrix T in Schur canonical form. Matrices of this
+// type are produced by the Schur factorization of a real general matrix A
+//  A = Q T Q^T,
+// as computed by Dhseqr.
+//
+// The right eigenvector x of T corresponding to an
+// eigenvalue λ is defined by
+//  T x = λ x,
+// and the left eigenvector is defined by
+//  y^H T = λ y^H.
```
vladimir-ch: Reference lapack says just "transpose" here, but it should say "conjugate transpose" (the code confirms it).

kortschak:
Do you want me to add `*trevc` to lapacke?

native/dtrevc3.go
```diff
@@ -0,0 +1,862 @@
...
+	const (
+		nbmin = 8
+		nbmax = 128
+	)
+
+	// Use blocked version of back-transformation if sufficient workspace.
+	// Zero-out the workspace to avoid potential NaN propagation.
+	if howmny == lapack.AllEigVecMulQ && lwork >= n+2*n*nbmin {
+		nb = min((lwork-n)/(2*n), nbmax)
+		impl.Dlaset(blas.All, n, 1+2*nb, 0, 0, work, 1+2*nb)
```
kortschak: Why using `1+2*nb` instead of `n`?

native/dtrevc3.go
```diff
@@ -0,0 +1,862 @@
...
+	norms := work[:n]
+	ldb := 2 * nb
+	b := work[n : n+n*ldb]
+
+	// Compute 1-norm of each column of strictly upper triangular part of T
+	// to control overflow in triangular solver.
+	norms[0] = 0
+	for j := 1; j < n; j++ {
+		norms[j] = 0
+		for i := 0; i < j; i++ {
```
kortschak: `for i := range norms[:j-1] {`

native/dtrevc3.go
```diff
@@ -0,0 +1,862 @@
...
+	} else {
+		nb = 1
+	}
+
+	// Set the constants to control overflow.
+	ulp := dlamchP
+	smlnum := float64(n) / ulp * dlamchS
+	bignum := (1 - ulp) / smlnum
+
+	// Split work into a vector of column norms and an n×2*nb matrix b.
```
kortschak: Nice.

native/dtrevc3.go
```diff
@@ -0,0 +1,862 @@
...
+	ldb := 2 * nb
+	b := work[n : n+n*ldb]
+
+	// Compute 1-norm of each column of strictly upper triangular part of T
+	// to control overflow in triangular solver.
+	norms[0] = 0
+	for j := 1; j < n; j++ {
+		norms[j] = 0
+		for i := 0; i < j; i++ {
+			norms[j] += math.Abs(t[i*ldt+j])
```
kortschak: Does it make a difference to to generated code/performance if you use an intermediate for this - the intention being to avoid the `i` loop repeated bounds checks in `norms[j]`.

native/dtrevc3.go
```diff
@@ -0,0 +1,862 @@
...
+
+		// ip is used below to specify the real or complex eigenvalue:
+		//  ip == 0, real eigenvalue,
+		//        1, first  of conjugate complex pair (wr,wi),
+		//       -1, second of conjugate complex pair (wr,wi).
+		ip        int
+		iscomplex [nbmax]int // Stores ip for each column in current block.
+	)
+
+	if side == lapack.LeftEigVec {
```
kortschak: I'm torn here.

native/dtrevc3.go
```diff
@@ -0,0 +1,862 @@
...
+					continue
+				}
+			} else if !selected[ki-1] {
+				continue
+			}
+		}
+
+		// Compute the ki-th eigenvalue (wr,wi).
+		wr := t[ki*ldt+ki]
+		wi := 0.0
```
kortschak: `var wi float64` unless you have objections.

native/dtrevc3.go
```diff
@@ -0,0 +1,862 @@
...
+
+			b[ki*ldb+iv] = 1
+			// Form right-hand side.
+			for k := 0; k < ki; k++ {
+				b[k*ldb+iv] = -t[k*ldt+ki]
+			}
+			// Solve upper quasi-triangular system:
+			//  [ T[0:ki,0:ki] - wr ]*X = scale*b.
+			for j := ki - 1; j >= 0; {
+				if j == 0 || t[j*ldt+j-1] == 0 {
```
kortschak: Does the Fortran really do all that work for this? Why‽

native/dtrevc3.go
```diff
@@ -0,0 +1,862 @@
...
+					// 1×1 diagonal block.
+					scale, xnorm, _ := impl.Dlaln2(false, 1, 1, smin, 1, t[j*ldt+j:], ldt,
+						1, 1, b[j*ldb+iv:], ldb, wr, 0, x[:], 2)
+					// Scale X[0,0] to avoid overflow when updating the
```
kortschak: Why does the Fortran scale x(1,2) and the other doublings below?

native/dtrevc3.go
```diff
@@ -0,0 +1,862 @@
...
+					}
+					b[j*ldb+iv] = x[0]
+					// Update right-hand side.
+					bi.Daxpy(j, -x[0], t[j:], ldt, b[iv:], ldb)
+					j--
+				} else {
+					// 2×2 diagonal block.
+					scale, xnorm, _ := impl.Dlaln2(false, 2, 1, smin, 1, t[(j-1)*ldt+j-1:], ldt,
+						1, 1, b[(j-1)*ldb+iv:], ldb, wr, 0, x[:], 2)
+					// Scale X[0,0] and X[1,0] to avoid overflow
```
kortschak: I'm similarly confused by the disparity here. (Above I don't understand the Fortran, here I don't understand the Go - in all likelihood I don't understand either, so would you explain).

native/dtrevc3.go
```diff
@@ -0,0 +1,862 @@
...
+					// when updating the right-hand side.
+					if xnorm > 1 {
+						beta := math.Max(norms[j-1], norms[j])
+						if beta > bignum/xnorm {
+							x[0] /= xnorm
```
kortschak: Same here and below.

vladimir-ch:
> Do you want me to add *trevc to lapacke?

I think it is not necessary. Or do you see a reason why we would want it there?

native/dtrevc3.go
```diff
@@ -0,0 +1,862 @@
...
+//
+// This routine returns the matrices X and/or Y of right and left eigenvectors
+// of T, or the products Q*X and/or Q*Y, where Q is an input matrix. If Q is the
+// orthogonal factor that reduces a matrix A to Schur form T, then Q*X and Q*Y
+// are the matrices of right and left eigenvectors of A.
+//
+// If side == lapack.RightEigVec, only right eigenvectors will be computed.
+// If side == lapack.LeftEigVec, only left eigenvectors will be computed.
+// If side == lapack.RightLeftEigVec, both right and left eigenvectors will be computed.
+// For other values of side Dtrevc3 will panic.
```
vladimir-ch: Done.

native/dtrevc3.go
```diff
@@ -0,0 +1,862 @@
...
+//
+// If howmny == lapack.AllEigVec, all right and/or left eigenvectors will be
+// computed.
+// If howmny == lapack.AllEigVecMulQ, all right and/or left eigenvectors will be
+// computed and multiplied from left by the matrices in VR and/or VL.
+// If howmny == lapack.SelectedEigVec, right and/or left eigenvectors will be
+// computed as indicated by selected.
+// For other values of howmny Dtrevc3 will panic.
```
vladimir-ch: Done.

native/dtrevc3.go
```diff
@@ -0,0 +1,862 @@
...
+	}
+	if lwork != -1 {
+		if howmny == lapack.SelectedEigVec {
+			if len(selected) != n {
+				panic("lapack: bad selected length")
+			}
+			// Set m to the number of columns required to store the
+			// selected eigenvectors, and standardize the slice
+			// selected.
+			for j := 0; j < n; {
```
vladimir-ch: Yeah, I've written this real/complex eigenvalue distingushing loop so many times by now that it's become a second nature. I've tried to follow the fortran original more closely and keep the pair variable but it looks worse than what I have here. I've added two comments and removed the `continue` statement so that now the two branches are directly comparable. Let me know what you think.

native/dtrevc3.go
```diff
@@ -0,0 +1,862 @@
...
+		}
+		if (side == lapack.LeftEigVec || side == lapack.RightLeftEigVec) && m > 0 {
+			checkMatrix(n, m, vl, ldvl)
+		}
+	}
+
+	// Quick return if possible.
+	if n == 0 {
+		work[0] = 1
+		return
```
vladimir-ch: Sorry. Done.

native/dtrevc3.go
```diff
@@ -0,0 +1,862 @@
...
+
+	nb := impl.Ilaenv(1, "DTREVC", string(side)+string(howmny), n, -1, -1, -1)
+
+	// Quick return in case of a workspace query.
+	if lwork == -1 {
+		work[0] = float64(n + 2*n*nb)
+		return m
+	}
+
+	const (
```
vladimir-ch: I've been thinking about a nice place for these constants, L172 is a scenic spot.

kortschak:
>I think it is not necessary. Or do you see a reason why we would want it there?

No reason to want it except perhaps to cross test between lapacke and native.

native/dtrevc3.go
vladimir-ch: `work` consist of n elements for the column norms and 2*n*nb elements for the b matrix. Here we pretend that work is an nx(1+2*nb) matrix, and since Fortran is column-major, it uses n for the leading dimension, while we use 1+2*nb.

native/dtrevc3.go
vladimir-ch: Mmm, I'd rather keep the current version. For me what you propose suggests that i is somehow used to access the norms slide which is not true.

native/dtrevc3.go
vladimir-ch: I didn't check but using an intermediate variable is much more reasonable. Done.

native/dtrevc3.go
kortschak: That's nice, thanks.

native/dtrevc3.go
vladimir-ch: Yeah, me too. I did it to reduce the indentation, it made the painful work a tiny bit more convenient. I can remove it now, I really have no preference.

native/dtrevc3.go
vladimir-ch: Done.

native/dtrevc3.go
vladimir-ch: I think it is because you cannot modify the loop variable within the loop body in Fortran?

native/dtrevc3.go
kortschak: No worries.

native/dtrevc3.go
kortschak: Leave it as it is.

native/dtrevc3.go
vladimir-ch: I'm not sure I fully understand the question and what is unclear.

A possible answer: We multiply the corresponding column(s) of T by x[0] (x[1], ...) and subtract it from the right-hand side (in Daxpy below). `norms` holds the maximum norm of each column of T. So if the norm of x times the norm of the column of T is greater than bignum, then there is an element of T in that column that will overflow. Therefore in that case we normalize x to avoid the overflow. Scaling by `scale` comes from Dlaln2. Scaling of `scale` by `xnorm` is inevitable once we scale x[0] (x[1], ...) because we are scaling one component of the eigenvector so we must scale the previously computed components too and the right hand side (to scale the components not yet computed). And we are free to multiply the eigenvector by any non-zero number.

kortschak: Ahhhh. I've figured it out. Sorry, I had phase shifted on the Fortran and was aligning the code here with the code some sections down (~ at line 645). This was due to scroll down and up to see where continues were jumping to. It's clear now.

vladimir-ch: That was my alternative hypothesis (it happened to me too) but I prefered the above reply because it forced me to actually look at what the scaling is doing.

native/dtrevc3.go
```diff
@@ -0,0 +1,866 @@
...
+				// Version 1: back-transform each vector with GEMV, Q*x.
+				if ki > 0 {
+					bi.Dgemv(blas.NoTrans, n, ki, 1, vr, ldvr, b[iv:], ldb,
+						b[ki*ldb+iv], vr[ki:], ldvr)
+				}
+				ii := bi.Idamax(n, vr[ki:], ldvr)
+				remax := 1 / math.Abs(vr[ii*ldvr+ki])
+				bi.Dscal(n, remax, vr[ki:], ldvr)
+			default:
+				// Version 2: back-transform block of vectors with GEMM.
```
kortschak: Where is this from?

native/dtrevc3.go
```diff
@@ -0,0 +1,866 @@
...
+						b[(ki+1)*ldb+iv+1], vl[ki+1:], ldvl)
+				} else {
+					bi.Dscal(n, b[ki*ldb+iv], vl[ki:], ldvl)
+					bi.Dscal(n, b[(ki+1)*ldb+iv+1], vl[ki+1:], ldvl)
+				}
+				emax := 0.0
+				for k := 0; k < n; k++ {
+					emax = math.Max(emax, math.Abs(vl[k*ldvl+ki])+math.Abs(vl[k*ldvl+ki+1]))
+				}
+				remax := 1 / emax
```
kortschak: Is it worth putting these types of cases in conditions for `emax != 1`? It's done elsewhere here. Either it is and it should be everywhere, or it's not and it shouldn't be done.

Is it the case that the presence or absence of the test for unity is based on probability of the scalar being one?

native/dtrevc3.go
```diff
@@ -0,0 +1,866 @@
...
+					b[k*ldb+iv] = 0
+					b[k*ldb+iv+1] = 0
+				}
+				iscomplex[iv] = ip
+				iscomplex[iv+1] = -ip
+				iv++
+				// Back-transform and normalization is done below.
+			}
+		}
+		if nb > 1 {
```
kortschak: Follow up: where is this from?

native/dtrevc3.go
vladimir-ch: Yes, it is based on probability of the scalar being one. In the checks `if scale != 1 {`, `scale` comes from `Dlaln2` and is less than one only if `x` would overflow without scaling it with `scale`. I think the assumption that this won't happen frequently  is reasonable. On the other hand, for a general matrix it is reasonable to assume that its largest element in a column will not be one.

native/dtrevc3.go
vladimir-ch: From L190-195. `nb` is greater than one only if the backtransform is being done and there is enough workspace, otherwise the `case nb == 1` above are taken.

kortschak: Sorry, I didn't mean the condition, I mean the blocked approach. I missed it in the Fortran.

testlapack/dtrevc3.go
```diff
@@ -0,0 +1,209 @@
...
+			evim := blas64.Vector{vr.Stride, vr.Data[k+1:]}
+			var evmax float64
+			for i := 0; i < n; i++ {
+				evmax = math.Max(evmax, math.Abs(evre.Data[i*evre.Inc])+
+					math.Abs(evim.Data[i*evim.Inc]))
+			}
+			if math.Abs(evmax-1) > tol {
+				t.Errorf("%v: magnitude of largest element of VR[:,%v] not 1", prefix, k)
+			}
+			if !isRightEigenvectorOf(tmat, evre, evim, complex(re, im), tol) {
```
vladimir-ch: Oh, I'm forgetting to check the other eigenvector `evre - i*evim` corresponding to `re-i*im`. Will fix.

vladimir-ch: Ok, done.

vladimir-ch:
PTAL

lapack.go
btracey: We've in the past just modified the blas type to help lapack. I don't think blas actually uses uplo.All. I'm okay with such a modification again.

lapack.go
btracey: ugh, but yes.

vladimir-ch: At least the `Q` refers to the documentation for Dhseqr where `Q` denotes the matrix. I was also thinking about `AllEigVecUpdate` but I'm not sure if it's better. Lapack uses "backtransformed" but that's even worse.

I will revisit these constants again later. I'd like to use `EV` instead of `EigVec` but that's ambiguous with EigenValues.

lapack.go
vladimir-ch: If we were to modify blas, I would make it
```
const (
    Left Side = 141 + iota
    Right
    Both   // <- add this
)
```
instead of mixing it with `Uplo.All`.

vladimir-ch: GitHub puts the comments here in a non-chronological order for me. Strange.

Anyway, I think I misunderstood you comment, then, @btracey . Do you want me to do anything about the `EigVecSide` type?

testlapack/general.go
```diff
@@ -1038,3 +1038,85 @@ func unbalancedSparseGeneral(m, n, stride int, nonzeros int, rnd *rand.Rand) bla
...
+		panic("bad column index")
+	}
+	col := make([]float64, a.Rows)
+	for i := range col {
+		col[i] = a.Data[i*a.Stride+j]
+	}
+	return col
+}
+
+func isRightEigenvectorOf(a blas64.General, xRe, xIm []float64, lambda complex128, tol float64) bool {
```
vladimir-ch: I added a function comment, some comments in the code, and made the code less dense. I think it is better now. Done.

testlapack/general.go
```diff
@@ -1038,3 +1038,85 @@ func unbalancedSparseGeneral(m, n, stride int, nonzeros int, rnd *rand.Rand) bla
...
+	if a.Rows != a.Cols {
+		panic("matrix not square")
+	}
+
+	n := a.Rows
+	axRe := make([]float64, n)
```
vladimir-ch: Changed to `xReAns`. Done.

testlapack/general.go
```diff
@@ -1038,3 +1038,85 @@ func unbalancedSparseGeneral(m, n, stride int, nonzeros int, rnd *rand.Rand) bla
...
+	blas64.Gemv(blas.NoTrans, 1, a, blas64.Vector{1, xRe}, 0, blas64.Vector{1, axRe})
+	if imag(lambda) == 0 {
+		// Real eigenvalue and eigenvector.
```
vladimir-ch: Great observation. I was so submerged in `Dtrevc3` that I didn't consider this situation because it doesn't arise with Dtrevc3

Real matrix cannot have a real eigenvector corresponding to complex eigenvalue. I added a check for this to `is{Left,Right}EigenvectorOf`.

Real eigenvalue and complex eigenvector is possible (for example, multiply the real eigenvector by i) although a real eigenvector always exists in such case and that's what Dtrevc3 computes. I modified the checking routines to cover this.

Thanks. Done.

testlapack/dtrevc3.go
```diff
@@ -0,0 +1,210 @@
...
+	const tol = 1e-14
+
+	n := tmat.Rows
+	extra := tmat.Stride - tmat.Cols
+	right := side != lapack.LeftEigVec
+	left := side != lapack.RightEigVec
+
+	var selected, selectedWant []bool
+	var mWant int
+	if howmny == lapack.SelectedEigVec {
```
vladimir-ch: Added some comments that hopefully clarify it. The loop just mimics what Dtrevc3 does: it takes `selected`, standardizes it based on whether the eigenvalues of `T` are real or complex, and computes how many columns are necessary for storing the eigenvectors. For real eigenvalues it just sets the corresponding element of `selectedWant` to true. For complex eigenvalues, it sets to true only the element corresponding to the first column of the pair.

PR #188, opened 2016-09-13
cgo/lapacke: make *geev,*geevx return int

### LINKED ISSUE #187: lapacke: Dgeev discards necessary information by returning bool instead of int

kortschak:

Fixes #187.


vladimir-ch:
LGTM

PR #190, opened 2016-09-28
native,testlapack: fix Dlartg for g==0 and update its test

vladimir-ch:


vladimir-ch:
Testing Dgeev on the matrix
```
[  0 1 ]
[ -1 0 ]
```
was failing due to the eigenvectors not being eigenvectors and not having norm 1. After checking that Dtrevc3 gives the correct eigenvectors, I narrowed it to Dgeev where the eigenvectors are rotated so that their largest component is real. What I saw that Dlartg returns cosine and sine both equal to 1, the culprit was identified.

PR #191, opened 2016-09-30
native,cgo: add Dgeev

vladimir-ch:
I still want to extend the testing a bit but you can start looking at this, especially the testing.

### DIFF
--- a/cgo/lapack.go
+++ b/cgo/lapack.go
@@ -308,16 +308,20 @@ func (impl Implementation) Dgebal(job lapack.Job, n int, a []float64, lda int, s
 // the eigenvectors of the original matrix.
 //
 // Dgebak is an internal routine. It is exported for testing purposes.
-func (impl Implementation) Dgebak(job lapack.Job, side blas.Side, n, ilo, ihi int, scale []float64, m int, v []float64, ldv int) {
+func (impl Implementation) Dgebak(job lapack.Job, side lapack.EigVecSide, n, ilo, ihi int, scale []float64, m int, v []float64, ldv int) {
 	switch job {
 	default:
 		panic(badJob)
 	case lapack.None, lapack.Permute, lapack.Scale, lapack.PermuteScale:
 	}
+	var bside blas.Side
 	switch side {
 	default:
 		panic(badSide)
-	case blas.Left, blas.Right:
+	case lapack.LeftEigVec:
+		bside = blas.Left
+	case lapack.RightEigVec:
+		bside = blas.Right
 	}
 	checkMatrix(n, m, v, ldv)
 	switch {
@@ -334,7 +338,7 @@ func (impl Implementation) Dgebak(job lapack.Job, side blas.Side, n, ilo, ihi in
 	for j := ihi + 1; j < n; j++ {
 		scale[j]++
 	}
-	lapacke.Dgebak(job, side, n, ilo+1, ihi+1, scale, m, v, ldv)
+	lapacke.Dgebak(job, bside, n, ilo+1, ihi+1, scale, m, v, ldv)
 	// Convert permutation indices back to 0-based.
 	for j := 0; j < ilo; j++ {
 		scale[j]--
--- a/cgo/lapack.go
+++ b/cgo/lapack.go
@@ -1716,3 +1716,113 @@ func (impl Implementation) Dhseqr(job lapack.Job, compz lapack.Comp, n, ilo, ihi
 
 	return lapacke.Dhseqr(job, compz, n, ilo+1, ihi+1, h, ldh, wr, wi, z, ldz, work, lwork)
 }
+
+// Dgeev computes the eigenvalues and, optionally, the left and/or right
+// eigenvectors for an n×n real nonsymmetric matrix A.
+//
+// The right eigenvector v_j of A corresponding to an eigenvalue λ_j
+// is defined by
+//  A v_j = λ_j v_j,
+// and the left eigenvector u_j corresponding to an eigenvalue λ_j is defined by
+//  u_j^H A = λ_j u_j^H,
+// where u_j^H is the conjugate transpose of u_j.
+//
+// On return, A will be overwritten and the left and right eigenvectors will be
+// stored, respectively, in the columns of the n×n matrices VL and VR in the
+// same order as their eigenvalues. If the j-th eigenvalue is real, then
+//  u_j = VL[:,j],
+//  v_j = VR[:,j],
+// and if it is not real, then j and j+1 form a complex conjugate pair and the
+// eigenvectors can be recovered as
+//  u_j     = VL[:,j] + i*VL[:,j+1],
+//  u_{j+1} = VL[:,j] - i*VL[:,j+1],
+//  v_j     = VR[:,j] + i*VR[:,j+1],
+//  v_{j+1} = VR[:,j] - i*VR[:,j+1].
+// where i is the imaginary unit. The computed eigenvectors are normalized to
+// have Euclidean norm equal to 1 and largest component real.
+//
+// Left eigenvectors will be computed only if jobvl == lapack.ComputeLeftEV,
+// otherwise jobvl must be lapack.None. Right eigenvectors will be computed
+// only if jobvr == lapack.ComputeRightEV, otherwise jobvr must be lapack.None.
+// For other values of jobvl and jobvr Dgeev will panic.
+//
+// wr and wi contain the real and imaginary parts, respectively, of the computed
+// eigenvalues. Complex conjugate pairs of eigenvalues appear consecutively with
+// the eigenvalue having the positive imaginary part first.
+// wr and wi must have length n, and Dgeev will panic otherwise.
+//
+// work must have length at least lwork and lwork must be at least max(1,4*n) if
+// the left or right eigenvectors are computed, and at least max(1,3*n) if no
+// eigenvectors are computed. For good performance, lwork must generally be
+// larger.  On return, optimal value of lwork will be stored in work[0].
+//
+// If lwork == -1, instead of performing Dgeev, the function only calculates the
+// optimal vaule of lwork and stores it into work[0].
+//
+// On return, first is the index of the first valid eigenvalue. If first == 0,
+// all eigenvalues and eigenvectors have been computed. If first is positive,
+// Dgeev failed to compute all the eigenvalues, no eigenvectors have been
+// computed and wr[first:] and wi[first:] contain those eigenvalues which have
+// converged.
+func (impl Implementation) Dgeev(jobvl lapack.JobLeftEV, jobvr lapack.JobRightEV, n int, a []float64, lda int, wr, wi []float64, vl []float64, ldvl int, vr []float64, ldvr int, work []float64, lwork int) (first int) {
+	var wantvl bool
+	switch jobvl {
+	default:
+		panic("lapack: invalid JobLeftEV")
+	case lapack.ComputeLeftEV:
+		wantvl = true
+	case lapack.None:
+		wantvl = false
+	}
+	var wantvr bool
+	switch jobvr {
+	default:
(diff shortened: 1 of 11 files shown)


btracey:
Wooooooo!

btracey:
This looks like a real failure.

native/dgeev.go
```diff
@@ -0,0 +1,283 @@
...
+	var side lapack.EigVecSide
+	if wantvl {
+		side = lapack.LeftEigVec
+		// Copy Householder vectors to VL.
+		impl.Dlacpy(blas.Lower, n, n, a, lda, vl, ldvl)
+		// Generate orthogonal matrix in VL.
+		impl.Dorghr(n, ilo, ihi, vl, ldvl, tau, work[iwrk:], lwork-iwrk)
+		// Perform QR iteration, accumulating Schur vectors in VL.
+		iwrk = n
+		firstok = impl.Dhseqr(lapack.EigenvaluesAndSchur, lapack.UpdateZ, n, ilo, ihi,
```
btracey: Make this firstOk or firstOK? I was reading it as "first tok" for most of the code and was really confused (yes, I realize that is nonsensical)

native/dgeev.go
```diff
@@ -0,0 +1,283 @@
...
+			impl.Dlascl(lapack.General, 0, 0, cscale, anrm, ilo, 1, wr, 1)
+			impl.Dlascl(lapack.General, 0, 0, cscale, anrm, ilo, 1, wi, 1)
+		}
+		work[0] = float64(maxwrk)
+		return firstok
+	}
+
+	if wantvl || wantvr {
+		// Compute left and/or right eigenvectors.
+		impl.Dtrevc3(side, lapack.AllEigVecMulQ, nil, n,
```
btracey: What does the fortran keyword do here?

native/dgeev.go
kortschak: `validFrom` or even just `first`?

lapack.go
```diff
@@ -156,3 +156,15 @@ const (
 	AllEigVecMulQ  HowMany = 'B' // Compute all right and/or left eigenvectors multiplied by an input matrix.
 	SelectedEigVec HowMany = 'S' // Compute selected right and/or left eigenvectors.
 )
+
+// Job types for Dgeev.
+type (
+	JobLeftEV  byte
```
kortschak: What's the rationale for these additions?

native/dgeev.go
```diff
@@ -0,0 +1,283 @@
...
+		return 0
+	}
+
+	maxwrk := 2*n + n*impl.Ilaenv(1, "DGEHRD", " ", n, 1, n, 0)
+	if wantvl || wantvr {
+		maxwrk = max(maxwrk, 2*n+(n-1)*impl.Ilaenv(1, "DORGHR", " ", n, 1, n, -1))
+		impl.Dhseqr(lapack.EigenvaluesAndSchur, lapack.UpdateZ, n, 0, n-1,
+			nil, 1, nil, nil, nil, 1, work, -1)
+		maxwrk = max(maxwrk, max(n+1, n+int(work[0])))
+		if wantvl {
```
kortschak:
```
job := lapack.RightEigVec
if wantvl {
    job = lapack.LeftEigVec
}
impl.Dtrevc3(job, lapack.AllEigVecMulQ, nil, n,
    nil, 1, nil, 1, nil, 1, n, work, -1)
```

native/dgeev.go
```diff
@@ -0,0 +1,283 @@
...
+		return 0
+	}
+
+	// Get machine constants.
+	smlnum := math.Sqrt(dlamchS) / dlamchP
+	bignum := 1 / smlnum
+
+	// Scale A if max element outside range [smlnum,bignum].
+	anrm := impl.Dlange(lapack.MaxAbs, n, n, a, lda, nil)
+	scalea := false
```
kortschak: `var scalea bool`?

native/dgeev.go
kortschak: `SELECT`? It's not a keyword here, it's a []bool equivalent.

testlapack/dgeev.go
```diff
@@ -0,0 +1,693 @@
...
+	}
+
+	// If the eigenvectors were not computed, check the computed eigenvalues
+	// against provided known eigenvalues.
+	if first > 0 || (jobvl == lapack.None && jobvr == lapack.None) {
+		if test.evWant == nil {
+			return
+		}
+		used := make([]bool, n)
+		for i := first; i < n; i++ {
```
btracey: Does this test actually work? If any of the eigenvalues match, idx != -1, so the error check below won't work as far as I can tell. Seems like the test should be about if all of the values of `used` are set. It's also weird that `i` indexes from first, but evWant indexes from 0.

testlapack/dgeev.go
```diff
@@ -0,0 +1,693 @@
...
+			if idx == -1 {
+				t.Log(test.evWant)
+				t.Log(dist)
+				t.Errorf("%v: unexpected eigenvalue %v", prefix, evGot)
+			}
+		}
+		return
+	}
+
+	for k := 0; k < n; {
```
btracey: Add comment.

cgo/lapack.go
```diff
@@ -1712,3 +1716,111 @@ func (impl Implementation) Dhseqr(job lapack.Job, compz lapack.Comp, n, ilo, ihi
...
 	return lapacke.Dhseqr(job, compz, n, ilo+1, ihi+1, h, ldh, wr, wi, z, ldz, work, lwork)
 }
+
+// Dgeev computes the eigenvalues and, optionally, the left and/or right
+// eigenvectors for an n×n real nonsymmetric matrix A.
+//
+// The right eigenvector v_j of A corresponding to an eigenvalue λ_j
+// is defined by
+//  A v_j = λ_j v_j,
+// and the left eigenvector u_j is defined by
```

native/dgeev.go
```diff
@@ -0,0 +1,283 @@
...
+// and the left eigenvector u_j is defined by
+//  u_j^H A = λ_j u_j^H,
+// where u_j^H is the conjugate transpose of u_j.
+//
+// On return, A will be overwritten and the left and right eigenvectors will be
+// stored, respectively, in the columns of the n×n matrices VL and VR in the
+// same order as their eigenvalues. If the j-th eigenvalue is real, then
+//  u_j = VL[:,j],
+//  v_j = VR[:,j],
+// and if the j-th and (j+1)-th eigenvalues form a complex conjugate pair, then
```
vladimir-ch: Done.

native/dgeev.go
vladimir-ch: Went with `first`. Done.

native/dgeev.go
vladimir-ch: Done.

native/dgeev.go
vladimir-ch: Done.

lapack.go
vladimir-ch: Both jobs take the same values ('V' and 'N'), so if they are just the generic `lapack.Job`, I think it would be too easy for the caller to get them in the wrong order without any compiler error or runtime panic.

testlapack/dgeev.go
```diff
@@ -0,0 +1,693 @@
...
+				for _, extra := range []int{0, 11} {
+					for _, optwork := range []bool{false, true} {
+						testDgeev(t, impl, strconv.Itoa(i), test, jobvl, jobvr, extra, optwork)
+					}
+				}
+			}
+		}
+	}
+
+	// for _, n := range []int{10, 30, 50} {
```
vladimir-ch: I expected that commenting this out would trigger questions which in turn would provide a place for explanation and discussion.

It depends. The test matrices above are nice but many of them have too simple structure that will not trigger the inner parts of the algorithm. I wanted to have a couple of tests on (random) matrices with known eigenvalues that are a mix of real and complex conjugate pair numbers. Starting with the real Schur decomposition theorem that for every nxn matrix there exists an orthogonal matrix Q such that 'Q^T A Q = T' where T is in real Schur form (quasi upper triangular with 1x1 and 2x2 blocks on the diagonal), I generate a random T and a random Q and multiply them to get my random A with known eigenvalues. It works but the computed eigenvalues are rather inaccurate, I would have to use tolerance 1e-4 or even 1e-2 (it gets worse with increasing n and affects both real and complex eigenvalues). That's suspiciously loose and that's why I commented it out.

Possible places to look at:
- I'm not using the real Schur decomposition theorem (for every matrix there exists ...), rather "for any T in real Schur form and Q orthogonal, `Q T Q^T` has the same eigenvalues as T". I guess it's fine but I don't have a proof that it holds. I have seen using this approach to generate random symmetric matrices with given eigenvalues, but not for general non-symmetric matrices.
- There might be something wrong with `reflector` and `applyReflector` functions although it does seem that the matrices are indeed orthogonal.
- The randomness creates bad matrices so that `Q T Q^T` has slightly different eigenvalues than T. I suspect this is the reason.

I'll try to think about it further but any comments are appreciated anyway.

testlapack/dgeev.go
```diff
@@ -0,0 +1,693 @@
...
+
+	var vr blas64.General
+	if jobvr == lapack.ComputeRightEV {
+		vr = nanGeneral(n, n, n)
+	}
+
+	wr := make([]float64, n)
+	wi := make([]float64, n)
+
+	lwork := max(1, 3*n)
```
vladimir-ch: I'll try to do something about it.

testlapack/dgeev.go
vladimir-ch: The idea is just to take each computed eigenvalue and see if there is an unused known eigenvalue sufficiently close to it. If there is not (i.e., idx == -1), then error. The order of the eigenvalues is unknown and if first > 0, then it is not known which eigenvalues were actually computed, so I can't think of anything better than this. This is also why `i` indexes from first (the index of the first valid eigenvalue), which evWant from 0 (the computed eigenvalue can be any element of `test.evWant`).

Of course, this test assumes that the known eigenvalues are well separated.

lapack.go
kortschak: OK

testlapack/dgeev.go
vladimir-ch: Done.

testlapack/dgeev.go
btracey: Why don't you generate random eigenvalues and eigenvectors, multiply the matrices, and then run the algorithm?

testlapack/dgeev.go
btracey: I see, okay

testlapack/dgeev.go
vladimir-ch: I've tried using the Jordan canonical form with the same result. I've also tried replacing `reflector` and `applyReflector` with generating a random orthogonal matrix and passing it through `Dgeqrf` and `Dorgqr` with the same result. Also, `cgo` behaves identically. All this kind of confirms my initial suspicion that `Q T Q^T` simply has slightly different eigenvalues from those of `T`  due to the matrix multiplication . Without known eigenvalues this test becomes worthless so I suggest removing it. @btracey ?

testlapack/dgeev.go
vladimir-ch: Ok, done. PTAL @btracey

vladimir-ch:
I think this is getting stabilized. Coverage is almost complete (I still can't get any non-converged eigenvalue), I've added testing with an intermediate workspace size, and I've added Geev to `lapack64`. PTAL

btracey:
LGTM, but wait for @kortschak

kortschak:
LGTM with a minimal at the tests.

As an aside, the naming of the implementation type is a little odd. Presumably eventually we will autogenerate a float32 implementation as well. What is the plan for naming when that happens?

vladimir-ch:
What implementation type do you mean?

kortschak:
Sorry. Brain fart.

PR #194, opened 2016-10-07
testlapack: fix lwork in test for Dgeev

vladimir-ch:
- Fix wrong calculation of `lwork` for intermediate workspace size
- ~~Use positive stride even if the corresponding matrix is not referenced or has zero size. While we don't care, cgo implementations sometimes do.~~

### DIFF
--- a/testlapack/dgeev.go
+++ b/testlapack/dgeev.go
@@ -571,17 +571,18 @@ func testDgeev(t *testing.T, impl Dgeever, tc string, test dgeevTest, jobvl lapa
 	var lwork int
 	switch wl {
 	case minimumWork:
-		lwork = max(1, 3*n)
 		if jobvl == lapack.ComputeLeftEV || jobvr == lapack.ComputeRightEV {
 			lwork = max(1, 4*n)
+		} else {
+			lwork = max(1, 3*n)
 		}
 	case mediumWork:
 		work := make([]float64, 1)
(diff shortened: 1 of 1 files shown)


testlapack/dgeev.go
```diff
@@ -556,11 +556,13 @@ func testDgeev(t *testing.T, impl Dgeever, tc string, test dgeevTest, jobvl lapa
 	n := a.Rows
 
 	var vl blas64.General
+	vl.Stride = 1
 	if jobvl == lapack.ComputeLeftEV {
 		vl = nanGeneral(n, n, n)
```
kortschak: Does n ever == 0 here?

vladimir-ch:
> Is there a test where this fails?

No, thanks for watching over what I do. I forgot that I already take care of LAPACKE inconsistent stride checks in our `cgo` calls. I reverted that part of the change. I keep the change to `lwork`, I want to make sure `lwork` is between the minimum and optimum size.

testlapack/dgeev.go
```diff
@@ -578,10 +578,9 @@ func testDgeev(t *testing.T, impl Dgeever, tc string, test dgeevTest, jobvl lapa
 	case mediumWork:
 		work := make([]float64, 1)
 		impl.Dgeev(jobvl, jobvr, n, nil, 1, nil, nil, nil, 1, nil, 1, work, -1)
-		lwork = int(work[0])
```
kortschak: Now it's clearer. Thanks.

PR #196, opened 2016-10-12
cgo/lapacke: replace genLapacke.pl

kortschak:
Woot!


cgo/lapacke/generate_lapacke.go
```diff
@@ -0,0 +1,554 @@
...
+	"float":          "float32",
+	"double":         "float64",
+	"float complex":  "complex64",
+	"double complex": "complex128",
+}
+
+var cgoEnums = map[string]*template.Template{}
+
+var byteTypes = map[string]string{
+	"compq": "lapack.Comp",
```
vladimir-ch: It does not belong into this PR but I'll forget otherwise: I may have mentioned it before but I've been thinking that we could avoid referencing `lapack` from `lapacke` by simply using `byte` for these parameters. This PR enables to fine-tune better the type of the parameters but often we have to do a type conversion anyway. Moreover, for example in case of `side` it is more complicated by the fact that the type depends also on the routine (for example, I use `side lapack.EVSide` in `native.Dgebak`, not `blas.Side`).

cgo/lapacke/generate_lapacke.go
```diff
@@ -0,0 +1,554 @@
...
+	if err != nil {
+		log.Fatal(err)
+	}
+	err = ioutil.WriteFile(target, b, 0664)
+	if err != nil {
+		log.Fatal(err)
+	}
+}
+
+// This remove select and selctg parameterised functions.
```
vladimir-ch: removes?

cgo/lapacke/generate_lapacke.go
```diff
@@ -0,0 +1,554 @@
...
+	"text/template"
+
+	"github.com/cznic/cc"
+
+	"github.com/gonum/internal/binding"
+)
+
+const (
+	header        = "lapacke.h"
+	documentation = "../native"
```
vladimir-ch: It would be nice to have documentation generation in the cgo package. I'm not sure if native's documentation couldn't be occasionally misleading here, for example about when a function panics, etc.

vladimir-ch: Ah, it seems that this constant is not used?

cgo/lapacke/generate_lapacke.go
kortschak: We can go either way, `byte` or fine tuned type - configuration is much easier now. Will you file an issue please?

cgo/lapacke/generate_lapacke.go
kortschak: Fixed.

Issue #142, opened 2016-06-10
cgo: add missing methods provided by lapacke


vladimir-ch:
- [x] Dlarfb
- [x] Dlarfg
- [x] Dlarft
- [x] Dlacn2
- [x] Dlapy2
- [x] Dlascl
- [x] Dlaset
- [x] Dlasrt
- [x] Dlaswp
- [x] Dorgql
- [x] Dorgtr
- [ ] Dsteqr
- [x] Dsterf
- [x] Dsytrd

Issue #205, opened 2016-11-17
native: Dlaswp panics when incX != 1 and -1


vladimir-ch:
Reference LAPACK allows any value of INCX.

vladimir-ch:
While it is true that the reference implementation allows any INCX, it does not seem beneficial and reasonable to do so (I think this is what led @btracey do disallow other values that 1 and -1). I could even imagine departing here from the reference and hope that they eventually do the same. For some background see https://github.com/Reference-LAPACK/lapack/issues/92 and https://github.com/Reference-LAPACK/lapack/commit/2ceda6db0f046b1b676030f9b6d92541e6222e3c

btracey:
If `|incx| != 1` there is no way to satisfy the behavior of the method.

It's unclear to me how much we want to conform to the "spec" and how much we want to prevent bad ideas.

vladimir-ch:
> If |incx| != 1 there is no way to satisfy the behavior of the method.

Recently I submitted a patch to Reference LAPACK to fix a bug in DLASWP and they also updated the docs. I think it is possible to satisfy the described behavior but it really does not make much sense to have `|incX| != 1`. I did implement it but then when writing a test for it, I gave up, so much effort and it is simply not worth it.

To sum it up, I would vote in this case to prevent bad ideas and not conform to the "spec", and close this issue.

kortschak:
I also vote for not following the current LAPACK reference in this case. From what @langou wrote in the commit, there is every possibility that the behaviour will change to forbid non-unit increments. We don't need to add to any inertia that slows that change.

vladimir-ch:
Thanks for the input.

PR #206, opened 2016-11-17
cgo: add Dlapy2, Dlascl, Dlaset, Dlasrt, Dlaswp

vladimir-ch:
Do we want tests for these routines? I plan to add one for Dlaswp, what about the rest?

Updates #142

### DIFF
--- a/cgo/lapack.go
+++ b/cgo/lapack.go
@@ -163,6 +163,13 @@ func (impl Implementation) Dlacpy(uplo blas.Uplo, m, n int, a []float64, lda int
 	lapacke.Dlacpy(uplo, m, n, a, lda, b, ldb)
 }
 
+// Dlapy2 is the LAPACK version of math.Hypot.
+//
+// Dlapy2 is an internal routine. It is exported for testing purposes.
+func (Implementation) Dlapy2(x, y float64) float64 {
+	return lapacke.Dlapy2(x, y)
+}
+
 // Dlarfb applies a block reflector to a matrix.
 //
 // In the call to Dlarfb, the mxn c is multiplied by the implicitly defined matrix h as follows:
--- a/cgo/lapack.go
+++ b/cgo/lapack.go
@@ -422,7 +422,10 @@ func (impl Implementation) Dlarfx(side blas.Side, m, n int, v []float64, tau flo
 	lapacke.Dlarfx(side, m, n, v, tau, c, ldc, work)
 }
 
-// Dlascl multiplies a rectangular matrix by a scalar.
+// Dlascl multiplies an m×n matrix by the scalar cto/cfrom.
+//
+// cfrom must not be zero, and cto and cfrom must not be NaN, otherwise Dlascl
+// will panic.
 //
 // Dlascl is an internal routine. It is exported for testing purposes.
 func (impl Implementation) Dlascl(kind lapack.MatrixType, kl, ku int, cfrom, cto float64, m, n int, a []float64, lda int) {
--- a/cgo/lapack.go
+++ b/cgo/lapack.go
@@ -6,6 +6,8 @@
 package cgo
 
 import (
+	"math"
+
 	"github.com/gonum/blas"
 	"github.com/gonum/lapack"
 	"github.com/gonum/lapack/cgo/lapacke"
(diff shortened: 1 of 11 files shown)


btracey:
If you'd like to add them it would be a good idea.

vladimir-ch:
I'd like to add the tests in separate PRs. Please take a look.

vladimir-ch:
In the end I ended up adding the tests here except for Dlaswp which needs to be fixed first #205.

vladimir-ch:
Ugh, `LAPACKE_dlascl_work` is completely broken.

vladimir-ch:
It's broken only in OpenBLAS that did not incorporate changes to `LAPACKE_dlascl_work` made earlier this year.

btracey:
Open an issue? They've been really responsive

btracey:
LGTM otherwise

vladimir-ch:
I'll merge when OpenBLAS has updated its copy of LAPACK to 3.6.1 (or at least `LAPACKE_dlascl_work`).

cgo/lapack.go
```diff
@@ -413,6 +425,63 @@ func (impl Implementation) Dlarfx(side blas.Side, m, n int, v []float64, tau flo
...
+	if cfrom == 0 {
+		panic(zeroDiv)
+	}
+	if math.IsNaN(cfrom) || math.IsNaN(cto) {
+		panic(nanScale)
+	}
+	lapacke.Dlascl(byte(kind), kl, ku, cfrom, cto, m, n, a, lda)
+}
+
+// Dlaset sets the off-diagonal elements of A to alpha, and the diagonal
```
kortschak: This doesn't currently match the behaviour of lapacke.Dlaset which panics is uplo is not blas.Upper or blas.Lower. To fix this, you need to add the following and regenerate lapacke:

```diff
diff --git a/cgo/lapacke/generate_lapacke.go b/cgo/lapacke/generate_lapacke.go
index 9110c9a..02432c1 100644
--- a/cgo/lapacke/generate_lapacke.go
+++ b/cgo/lapacke/generate_lapacke.go
@@ -64,6 +64,7 @@ var needsInt = map[string]bool{
 // The list keys are truncated by one character to cover all four numeric types.
 var allUplo = map[string]bool{
        "lacpy": true,
+       "laset": true,
 }
```

Note that even then, the comment is not true. The uplo check checks for U, L, A and panics if it is not one of those. This could be relaxed I guess - ?lacpy has the same horrible behaviour.

cgo/lapack.go
```diff
@@ -413,6 +425,63 @@ func (impl Implementation) Dlarfx(side blas.Side, m, n int, v []float64, tau flo
...
+//
+// Dlaset is an internal routine. It is exported for testing purposes.
+func (impl Implementation) Dlaset(uplo blas.Uplo, m, n int, alpha, beta float64, a []float64, lda int) {
+	checkMatrix(m, n, a, lda)
+	lapacke.Dlaset(uplo, m, n, alpha, beta, a, lda)
+}
+
+// Dlasrt sorts the numbers in the input slice d. If sort == lapack.SortIncreasing,
+// the elements are sorted in increasing order. If sort == lapack.SortDecreasing,
+// the elements are sorted in decreasing order.
```
kortschak: Note that this panics for s values other than lister here.

cgo/lapack.go
kortschak: This would require

```diff
diff --git a/cgo/lapacke/generate_lapacke.go b/cgo/lapacke/generate_lapacke.go
index 9110c9a..f082d16 100644
--- a/cgo/lapacke/generate_lapacke.go
+++ b/cgo/lapacke/generate_lapacke.go
@@ -421,10 +422,8 @@ func uplo(buf *bytes.Buffer, d binding.Declaration, p binding.Parameter) bool {
                ul = 'U'
        case blas.Lower:
                ul = 'L'
-       case blas.All:
-               ul = 'A'
        default:
-               panic("lapack: illegal triangle")
+               ul = 'A'
        }
 `)
```

in addition to the above changes.

cgo/lapack.go

cgo/lapack.go
kortschak: The change in #216 doesn't match the Fortran docs and is more invasive. The documented [behaviour](http://www.netlib.org/cgi-bin/netlibfiles.txt?format=txt&filename=/lapack/lapack_routine/dlacpy.f) of ?lacpy is to use all when not U or L. I would do what is suggested above.

```
*> DLACPY copies all or part of a two-dimensional matrix A to another
*> matrix B.
*> \endverbatim
*
*  Arguments:
*  ==========
*
*> \param[in] UPLO
*> \verbatim
*>          UPLO is CHARACTER*1
*>          Specifies the part of the matrix A to be copied to B.
*>          = 'U':      Upper triangular part
*>          = 'L':      Lower triangular part
*>          Otherwise:  All of the matrix A
*> \endverbatim
```

PR #209, opened 2016-11-18
cgo: add Dorgql, rewrite its test and update its docs

vladimir-ch:
I had to rewrite the test because LAPACKE does not provide the routines that it had been using. It still relies on Dlarfg which could be avoided if desired but I think it's ok like this.

Updates #142

### DIFF
--- a/native/dorgql.go
+++ b/native/dorgql.go
@@ -11,14 +11,22 @@ import (
 
 // Dorgql generates the m×n matrix Q with orthonormal columns defined as the
 // last n columns of a product of k elementary reflectors of order m
-//  Q = H_{k-1} * ... * H_1 * H_0
-// as returned by Dgelqf. See Dgelqf for more information.
+//  Q = H_{k-1} * ... * H_1 * H_0.
+//
+// It must hold that
+//  0 <= k <= n <= m,
+// and Dorgql will panic otherwise.
+//
+// On entry, the (n-k+i)-th column of A must contain the vector which defines
+// the elementary reflector H_i, for i=0,...,k-1, and tau[i] must contain its
+// scalar factor. On return, a contains the m×n matrix Q.
 //
 // tau must have length at least k, and Dorgql will panic otherwise.
 //
-// work is temporary storage, and lwork specifies the usable memory length. At minimum,
-// lwork >= n, and Dorgql will panic otherwise. The amount of blocking is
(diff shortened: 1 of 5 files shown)


native/dorgql.go
```diff
@@ -11,33 +11,56 @@ import (
...
+	case k > n:
+		panic(kGTN)
+	case lwork < max(1, n) && lwork != -1:
 		panic(badWork)
+	case len(work) < lwork:
+		panic(shortWork)
 	}
+	if lwork != -1 {
+		checkMatrix(m, n, a, lda)
+		if len(tau) < k {
```
vladimir-ch: I wanted to make this stricter and require `len(tau) == k` but it would require modifications in unrelated methods. Postponed to a future PR.

vladimir-ch:
I added a comparison with `Dorg2l` in order to keep the essence of the previous test and because  it is a good check to have. PTAL

btracey:
I don't think I like the test rewrite. There's too much overlap with Dorg2l. I think I prefer my previous solution, which was to write a wrapper interface to call the base method. See  for example blockedTranslate with Dgebd2 and Dgebrd

vladimir-ch:
With `blockedTranslate` we can't test for varying `lwork`, for example. Is that ok? In the past `work` and `lwork` caused trouble. More importantly, `Dorg2lTest` needs `Dgeql2` which is only implemented in native (not provided by LAPACKE) but whose blocked  version `Dgeqlf` we don't have even in native. I can rewrite the part that calls `Dgeql2` with what I have here but what about the work part?

btracey:
We could run both tests, right? Use the wrapper and run TestDorg2l, and then additionally run the Dorglq test which would test the work values. Could we use the same shim to wrap Dgeqlf and call it Dgeql2?

vladimir-ch:
> Could we use the same shim to wrap Dgeqlf and call it Dgeql2?

We could (if someone ported Dgeqlf to native and cgo) but we don't need Dgeqlf nor Dgeql2 to test Dorgql. We use them to create the elementary reflectors in A and tau but anything that resembles them will do.

> ... then additionally run the Dorglq test which would test the work values.

What do you mean by "the Dorglq test"? The test in this PR?

btracey:
Okay, this is fine then.

LGTM

PR #217, opened 2016-12-04
Update go 1.6.3 to 1.6.4, 1.7.3 to 1.7.4, and add go 1.8beta1 and remove 1.5.4 in the testing matrix

jonlawlor:


kortschak:
I think throughout the removal of 1.5 is premature. We have maintained 3 go release versions in the past, 1.8 is not yet release, so 1.5.4 should stay.

jonlawlor:
I was using @evertlammerts hub script with a small modification.  It is hard to get the sed script to add newlines in the -e field, but much easier to replace one line with another.  One way around would be to put a placeholder comment for where beta builds would go.

```bash
bash update.sh -g gonum -e "/usr/bin/sed -i \"\" -e 's/1\.5\.4/1\.6\.4/g' .travis.yml && /usr/bin/sed -i \"\" -e 's/1\.6\.3/1\.7\.4/g' .travis.yml && /usr/bin/sed -i \"\" -e 's/1\.7\.3/1\.8beta1/g' .travis.yml" -r jonlawlor -c '1.6.3->1.6.4, 1.7.3->1.7.4, +1.8beta1, -1.5.4' -p 'Update go 1.6.3 to 1.6.4, 1.7.3 to 1.7.4, and add go 1.8beta1 and remove 1.5.4 in the testing matrix' lapack stat plot graph integrate matrix mathext diff floats unit optimize blas internal
```
That requires a modification to https://gist.github.com/evertlammerts/733b6282fc59b95ca1ae5e107c70b0a5#file-update-sh-L109 - the `$EXEC` needs to be replaced with an `eval ${EXEC}`

jonlawlor:
Added 1.5.4 manually.  When 1.8 becomes stable I think we should put a comment placeholder there instead of deleting the line as it will make running the update script much easier.

sbinet:
(drive by comment, feel free to ignore)
Alternatively, as it's just yaml, we could just have a go struct type to (de)serialize using gopkg.in/yaml.v2 so the modifications would be simpler...

jonlawlor:
I checked that but the library didn't seem to handle comments?  Maybe I missed it though.

evertlammerts:
This works fine for me using BSD and GNU sed (note the escaped hard newline after 1.7.4... `\n` is a bit of a hassle to get to work):

```
sed -i '' -e '/1\.5\.4/d' -e 's/1\.6\.3/1\.6\.4/g' -e 's/1\.7\.3/1\.7\.4\
 - 1\.8beta1/g' .travis.yml
```

- delete 1.5.4
- replace 1.6.3 with 1.6.4
- replace 1.7.3 with 1.7.4 and add 1.8beta1

Also note that nothing prevents you from putting your sed (and other) commands into a bash script. Ie you could put the above in `myscript.sh` and run `update.sh` something like:

```
bash update.sh -e "bash $PWD/myscript.sh" ...
```

Personally, I wouldn't add a placeholder for the sake of `update.sh`.

kortschak:
@evertlammerts, yes. Sed has problems finding new lines, but not problems adding them.

```
~ $ echo "  - 1.7.3" | sed 's/1\.7\.3/1.7.4\n  - 1.8beta1/'
  - 1.7.4
  - 1.8beta1
```

I agree that adding a place holder to make up for limitations in tooling is a bad idea. If there does come a time when new lines need to be found, use perl as a RE engine.

If these PRs are merged, they need to be squashed and have their commit messages and PR titles corrected. In light of that, I would suggest that this family of PRs be abandoned and redone.

jonlawlor:
The newline problem isn't with sed, it was with the bash string used in the -e flag.  I'll redo the pulls.

PR #219, opened 2017-01-18
testlapack: add test for Dlaswp

### LINKED ISSUE #214: testlapack: add test for Dlaswp

vladimir-ch:
Closes #214 

The test failure comes from a bug in LAPACKE, see https://github.com/Reference-LAPACK/lapack/issues/110
Fix the reference, then get the change to OpenBLAS ... it will take some time. PTAL in the meantime.


vladimir-ch:
A fix has been merged into Reference LAPACK and OpenBLAS. I'll wait for travis to finish and then I'll merge this.

Issue #218, opened 2017-01-18
lapack/native: add dggsvd3 implementation


kortschak:
dggsvd3 implements Generalised Singular Value Decomposition that has uses in signal processing and genomics.

- [x] dggsvd3
- [x] - dtgsja
- [x] -- dlapll
- [x] -- dlags2
- [x] - dggsvp3
- [x] -- dlapmt
- [x] -- dgeqp3
- [x] --- dlaqps
- [x] --- dlaqp2
- [x] -- dgerq2
- [x] -- dormr2
- [x] --- dgerqf (required for testing)

kortschak:
Collecting implementations [here](https://github.com/gonum/lapack/tree/dggsvd3).

btracey:
Sorry, what do you mean by "Collecting implementations"?

kortschak:
The dggsvg3 branch will become the PR at some stage - the three checked routines above are committed on that branch.

btracey:
In the past we've just submitted each function individually as a PR. I don't see any gain in keeping them on a branch. I do see a gain in submitting them to master, for instance if other high-level routines share lower ones.

kortschak:
This doesn't preclude that. This approach allows me to keep my different development environments synced while I don't have a PR unit (at least one tested routine). At the moment I have three untested routines.

PR #224, opened 2017-01-23
cgo,native: implement dgeqp3 and dependencies

kortschak:

Documentation probably needs work and tests also.

Updates #218.

### DIFF
--- a/cgo/lapack.go
+++ b/cgo/lapack.go
@@ -116,6 +116,70 @@ type Implementation struct{}
 
 var _ lapack.Float64 = Implementation{}
 
+// Dgeqp3 computes a QR factorization with column pivoting of the
+// m×n matrix A: A*P = Q*R using Level 3 BLAS.
+//
+// The matrix Q is represented as a product of elementary reflectors
+//  Q = H_0 H_1 . . . H_{k-1}, where k = min(m,n).
+// Each H_i has the form
+//  H_i = I - tau * v * v^T
+// where tau and v are real vectors with v[0:i-1] = 0 and v[i] = 1;
+// v[i:m] is stored on exit in A[i:m, i], and tau in tau[i].
+//
+// jpvt specifies a column pivot to be applied to A. If
+// jpvt[j] is at least zero, the jth column of A is permuted
+// to the front of A*P (a leading column), if jpvt[j] is -1
+// the jth column of A is a free column. If jpvt[j] < -1, Dgeqp3
+// will panic. On return, jpvt holds the permutation that was
+// applied; the jth column of A*P was the jpvt[j] column of A.
+// jpvt must have length n or Dgeqp3 will panic.
+//
+// tau holds the scalar factors of the elementary reflectors.
+// It must have length min(m, n), otherwise Dgeqp3 will panic.
+//
+// work must have length at least max(1,lwork), and lwork must be at least
+// 3*n+1, otherwise Dgeqp3 will panic. For optimal performance lwork must
+// be at least 2*n+(n+1)*nb, where nb is the optimal blocksize. On return,
+// work[0] will contain the optimal value of lwork.
+//
+// If lwork == -1, instead of performing Dgeqp3, only the optimal value of lwork
+// will be stored in work[0].
+//
+// Dgeqp3 is an internal routine. It is exported for testing purposes.
+func (impl Implementation) Dgeqp3(m, n int, a []float64, lda int, jpvt []int, tau, work []float64, lwork int) {
+	checkMatrix(m, n, a, lda)
+	if len(jpvt) != n {
+		panic(badIpiv)
+	}
+	if len(tau) != min(m, n) {
+		panic(badTau)
+	}
+	if len(work) < max(1, lwork) {
+		panic(badWork)
+	}
+
+	// Don't update jpvt if querying lwkopt.
+	if lwork == -1 {
+		lapacke.Dgeqp3(m, n, a, lda, nil, nil, work, -1)
+		return
+	}
+
+	jpvt32 := make([]int32, len(jpvt))
+	for i, v := range jpvt {
+		v++
+		if v != int(int32(v)) || v < 0 || n < v {
+			panic("lapack: jpvt element out of range")
+		}
+		jpvt32[i] = int32(v)
+	}
+
+	lapacke.Dgeqp3(m, n, a, lda, jpvt32, tau, work, lwork)
+
+	for i, v := range jpvt32 {
+		jpvt[i] = int(v - 1)
+	}
+}
+
 // Dlacn2 estimates the 1-norm of an n×n matrix A using sequential updates with
 // matrix-vector products provided externally.
 //
--- a/cgo/lapack.go
+++ b/cgo/lapack.go
@@ -16,6 +16,7 @@ import (
 // Copied from lapack/native. Keep in sync.
 const (
 	absIncNotOne    = "lapack: increment not one or negative one"
+	badAuxv         = "lapack: auxv has insufficient length"
 	badD            = "lapack: d has insufficient length"
 	badDecompUpdate = "lapack: bad decomp update"
 	badDiag         = "lapack: bad diag"
@@ -34,6 +35,7 @@ const (
 	badK2           = "lapack: k2 out of range"
 	badKperm        = "lapack: incorrect permutation length"
 	badLdA          = "lapack: index of a out of range"
+	badNb           = "lapack: nb out of range"
 	badNorm         = "lapack: bad norm"
 	badPivot        = "lapack: bad pivot"
 	badS            = "lapack: s has insufficient length"
@@ -46,6 +48,8 @@ const (
 	badTauQ         = "lapack: tauQ has insufficient length"
 	badTauP         = "lapack: tauP has insufficient length"
 	badTrans        = "lapack: bad trans"
+	badVn1          = "lapack: vn1 has insufficient length"
+	badVn2          = "lapack: vn2 has insufficient length"
 	badUplo         = "lapack: illegal triangle"
 	badWork         = "lapack: insufficient working memory"
 	badWorkStride   = "lapack: insufficient working array stride"
@@ -60,6 +64,7 @@ const (
 	negZ            = "lapack: negative z value"
 	nLT0            = "lapack: n < 0"
 	nLTM            = "lapack: n < m"
+	offsetGTM       = "lapack: offset > m"
 	shortWork       = "lapack: working array shorter than declared"
 	zeroDiv         = "lapack: zero divisor"
 )
--- a/native/dgeqp3.go
+++ b/native/dgeqp3.go
@@ -1,0 +1,173 @@
+// Copyright ©2017 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package native
+
+import (
+	"github.com/gonum/blas"
+	"github.com/gonum/blas/blas64"
+)
+
+// Dgeqp3 computes a QR factorization with column pivoting of the
+// m×n matrix A: A*P = Q*R using Level 3 BLAS.
+//
+// The matrix Q is represented as a product of elementary reflectors
+//  Q = H_0 H_1 . . . H_{k-1}, where k = min(m,n).
+// Each H_i has the form
+//  H_i = I - tau * v * v^T
+// where tau and v are real vectors with v[0:i-1] = 0 and v[i] = 1;
+// v[i:m] is stored on exit in A[i:m, i], and tau in tau[i].
+//
+// jpvt specifies a column pivot to be applied to A. If
+// jpvt[j] is at least zero, the jth column of A is permuted
+// to the front of A*P (a leading column), if jpvt[j] is -1
+// the jth column of A is a free column. If jpvt[j] < -1, Dgeqp3
+// will panic. On return, jpvt holds the permutation that was
+// applied; the jth column of A*P was the jpvt[j] column of A.
+// jpvt must have length n or Dgeqp3 will panic.
+//
+// tau holds the scalar factors of the elementary reflectors.
+// It must have length min(m, n), otherwise Dgeqp3 will panic.
+//
+// work must have length at least max(1,lwork), and lwork must be at least
+// 3*n+1, otherwise Dgeqp3 will panic. For optimal performance lwork must
+// be at least 2*n+(n+1)*nb, where nb is the optimal blocksize. On return,
+// work[0] will contain the optimal value of lwork.
+//
+// If lwork == -1, instead of performing Dgeqp3, only the optimal value of lwork
+// will be stored in work[0].
+//
+// Dgeqp3 is an internal routine. It is exported for testing purposes.
+func (impl Implementation) Dgeqp3(m, n int, a []float64, lda int, jpvt []int, tau, work []float64, lwork int) {
+	const (
+		inb    = 1
+		inbmin = 2
+		ixover = 3
+	)
+	checkMatrix(m, n, a, lda)
+
+	if len(jpvt) != n {
+		panic(badIpiv)
+	}
+	for _, v := range jpvt {
+		if v < -1 || n <= v {
+			panic("lapack: jpvt element out of range")
+		}
+	}
+	minmn := min(m, n)
+	if len(tau) != minmn {
+		panic(badTau)
+	}
+	if len(work) < max(1, lwork) {
+		panic(badWork)
+	}
+
+	var iws, lwkopt, nb int
+	if minmn == 0 {
+		iws = 1
+		lwkopt = 1
+	} else {
+		iws = 3*n + 1
+		nb = impl.Ilaenv(inb, "DGEQRF", " ", m, n, -1, -1)
+		lwkopt = 2*n + (n+1)*nb
+	}
+	work[0] = float64(lwkopt)
+
+	if lwork == -1 {
+		return
+	}
+
+	bi := blas64.Implementation()
+
+	// Move initial columns up front.
+	var nfxd int
+	for j := 0; j < n; j++ {
+		if jpvt[j] == -1 {
+			jpvt[j] = j
+			continue
+		}
+		if j != nfxd {
+			bi.Dswap(m, a[j:], lda, a[nfxd:], lda)
+			jpvt[j], jpvt[nfxd] = jpvt[nfxd], j
+		} else {
+			jpvt[j] = j
+		}
+		nfxd++
+	}
+
+	// Factorize nfxd columns.
+	//
+	// Compute the QR factorization of nfxd columns and update remaining columns.
+	if nfxd > 0 {
+		na := min(m, nfxd)
+		impl.Dgeqrf(m, na, a, lda, tau, work, lwork)
+		iws = max(iws, int(work[0]))
+		if na < n {
+			impl.Dormqr(blas.Left, blas.Trans, m, n-na, na, a, lda, tau[:na], a[na:], lda,
+				work, lwork)
+			iws = max(iws, int(work[0]))
+		}
+	}
+
+	if nfxd >= minmn {
+		work[0] = float64(iws)
+		return
+	}
+
+	// Factorize free columns.
+	sm := m - nfxd
+	sn := n - nfxd
+	sminmn := minmn - nfxd
+
+	// Determine the block size.
+	nb = impl.Ilaenv(inb, "DGEQRF", " ", sm, sn, -1, -1)
+	nbmin := 2
+	nx := 0
+
+	if 1 < nb && nb < sminmn {
+		// Determine when to cross over from blocked to unblocked code.
+		nx = max(0, impl.Ilaenv(ixover, "DGEQRF", " ", sm, sn, -1, -1))
+
+		if nx < sminmn {
+			// Determine if workspace is large enough for blocked code.
+			minws := 2*sn + (sn+1)*nb
+			iws = max(iws, minws)
+			if lwork < minws {
+				// Not enough workspace to use optimal nb. Reduce
+				// nb and determine the minimum value of nb.
+				nb = (lwork - 2*sn) / (sn + 1)
+				nbmin = max(2, impl.Ilaenv(inbmin, "DGEQRF", " ", sm, sn, -1, -1))
+			}
+		}
+	}
+
+	// Initialize partial column norms.
+	// The first n elements of work store the exact column norms.
+	for j := nfxd; j < n; j++ {
+		work[j] = bi.Dnrm2(sm, a[nfxd*lda+j:], lda)
+		work[n+j] = work[j]
+	}
+	j := nfxd
+	if nbmin <= nb && nb < sminmn && nx < sminmn {
+		// Use blocked code initially.
+
+		// Compute factorization.
+		var fjb int
+		for topbmn := minmn - nx; j < topbmn; j += fjb {
+			jb := min(nb, topbmn-j)
+
+			// Factorize jb columns among columns j:n.
+			fjb = impl.Dlaqps(m, n-j, j, jb, a[j:], lda, jpvt[j:], tau[j:],
+				work[j:n], work[j+n:2*n], work[2*n:2*n+jb], work[2*n+jb:], jb)
+		}
+	}
+
+	// Use unblocked code to factor the last or only block.
+	if j < minmn {
+		impl.Dlaqp2(m, n-j, j, a[j:], lda, jpvt[j:], tau[j:],
+			work[j:n], work[j+n:2*n], work[2*n:])
+	}
+
+	work[0] = float64(iws)
+}
--- a/native/dlaqp2.go
+++ b/native/dlaqp2.go
@@ -1,0 +1,111 @@
+// Copyright ©2017 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package native
+
+import (
+	"math"
+
+	"github.com/gonum/blas"
+	"github.com/gonum/blas/blas64"
+)
+
+// Dlaqp2 computes a QR factorization with column pivoting of the block A[offset:m, 0:n]
+// of the m×n matrix A. The block A[0:offset, 0:n] is accordingly pivoted, but not factorized.
+//
+// On exit, the upper triangle of block A[offset:m, 0:n] is the triangular factor obtained.
+// The elements in block A[offset:m, 0:n] below the diagonal, together with tau, represent
+// the orthogonal matrix Q as a product of elementary reflectors.
+//
+// offset is number of rows of the matrix A that must be pivoted but not factorized.
+// offset must not be negative otherwise Dlaqp2 will panic.
+//
+// On exit, jpvt holds the permutation that was applied; the jth column of A*P was the
+// jpvt[j] column of A. jpvt must have length n, otherwise Dlaqp2 will panic.
+//
+// On exit tau holds the scalar factors of the elementary reflectors. It must have length
+// at least min(m-offset, n) otherwise Dlaqp2 will panic.
+//
+// vn1 and vn2 hold the partial and complete column norms respectively. They must have length n,
+// otherwise Dlaqp2 will panic.
+//
+// work must have length n, otherwise Dlaqp2 will panic.
+//
+// Dlaqp2 is an internal routine. It is exported for testing purposes.
+func (impl Implementation) Dlaqp2(m, n, offset int, a []float64, lda int, jpvt []int, tau, vn1, vn2, work []float64) {
+	checkMatrix(m, n, a, lda)
+	if len(jpvt) != n {
+		panic(badIpiv)
+	}
+	mn := min(m-offset, n)
+	if len(tau) < mn {
+		panic(badTau)
+	}
+	if len(vn1) < n {
+		panic(badVn1)
+	}
+	if len(vn2) < n {
+		panic(badVn2)
+	}
+	if len(work) < n {
+		panic(badWork)
+	}
+
+	tol3z := math.Sqrt(dlamchE)
+
+	bi := blas64.Implementation()
+
+	// Compute factorization.
+	for i := 0; i < mn; i++ {
+		offpi := offset + i
+
+		// Determine ith pivot column and swap if necessary.
+		p := i + bi.Idamax(n-i, vn1[i:], 1)
+		if p != i {
+			bi.Dswap(m, a[p:], lda, a[i:], lda)
+			jpvt[p], jpvt[i] = jpvt[i], jpvt[p]
+			vn1[p] = vn1[i]
+			vn2[p] = vn2[i]
+		}
+
+		// Generate elementary reflector H_i.
+		if offpi < m-1 {
+			a[offpi*lda+i], tau[i] = impl.Dlarfg(m-offpi, a[offpi*lda+i], a[(offpi+1)*lda+i:], lda)
+		} else {
+			tau[i] = 0
+		}
+
+		if i < n-1 {
+			// Apply H_i^T to A[offset+i:m, i:n] from the left.
+			aii := a[offpi*lda+i]
+			a[offpi*lda+i] = 1
+			impl.Dlarf(blas.Left, m-offpi, n-i-1, a[offpi*lda+i:], lda, tau[i], a[offpi*lda+i+1:], lda, work)
+			a[offpi*lda+i] = aii
+		}
+
+		// Update partial column norms.
+		for j := i + 1; j < n; j++ {
+			if vn1[j] == 0 {
+				continue
+			}
+
+			// The following marked lines follow from the
+			// analysis in Lapack Working Note 176.
+			r := math.Abs(a[offpi*lda+j]) / vn1[j] // *
+			temp := math.Max(0, 1-r*r)             // *
+			r = vn1[j] / vn2[j]                    // *
+			temp2 := temp * r * r                  // *
+			if temp2 < tol3z {
+				var v float64
+				if offpi < m-1 {
+					v = bi.Dnrm2(m-offpi-1, a[(offpi+1)*lda+j:], lda)
+				}
+				vn1[j] = v
+				vn2[j] = v
+			} else {
+				vn1[j] *= math.Sqrt(temp) // *
+			}
+		}
+	}
+}
--- a/native/dlaqps.go
+++ b/native/dlaqps.go
@@ -1,0 +1,217 @@
+// Copyright ©2017 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package native
+
+import (
+	"math"
+
+	"github.com/gonum/blas"
+	"github.com/gonum/blas/blas64"
+)
+
+// Dlaqps computes a step of QR factorization with column pivoting
+// of an m×n matrix A by using Blas-3. It tries to factorize nb
+// columns from A starting from the row offset, and updates all
+// of the matrix with Dgemm.
+//
+// In some cases, due to catastrophic cancellations, it cannot
+// factorize nb columns. Hence, the actual number of factorized
+// columns is returned in kb.
+//
+// Dlaqps computes a QR factorization with column pivoting of the
+// block A[offset:m, 0:nb] of the m×n matrix A. The block
+// A[0:offset, 0:n] is accordingly pivoted, but not factorized.
+//
+// On exit, the upper triangle of block A[offset:m, 0:kb] is the
+// triangular factor obtained. The elements in block A[offset:m, 0:n]
+// below the diagonal, together with tau, represent the orthogonal
+// matrix Q as a product of elementary reflectors.
+//
+// offset is number of rows of the matrix A that must be pivoted but
+// not factorized. offset must not be negative otherwise Dlaqps will panic.
+//
+// On exit, jpvt holds the permutation that was applied; the jth column
+// of A*P was the jpvt[j] column of A. jpvt must have length n,
+// otherwise Dlapqs will panic.
+//
+// On exit tau holds the scalar factors of the elementary reflectors.
+// It must have length nb, otherwise Dlapqs will panic.
+//
+// vn1 and vn2 hold the partial and complete column norms respectively.
+// They must have length n, otherwise Dlapqs will panic.
+//
+// auxv must have length nb, otherwise Dlaqps will panic.
+//
+// f and ldf represent an n×nb matrix F that is overwritten during the
+// call.
+//
+// Dlaqps is an internal routine. It is exported for testing purposes.
+func (impl Implementation) Dlaqps(m, n, offset, nb int, a []float64, lda int, jpvt []int, tau, vn1, vn2, auxv, f []float64, ldf int) (kb int) {
+	checkMatrix(m, n, a, lda)
+	checkMatrix(n, nb, f, ldf)
+	if offset > m {
+		panic(offsetGTM)
+	}
+	if n < 0 || nb > n {
+		panic(badNb)
+	}
+	if len(jpvt) != n {
+		panic(badIpiv)
+	}
+	if len(tau) < nb {
+		panic(badTau)
+	}
+	if len(vn1) < n {
+		panic(badVn1)
+	}
+	if len(vn2) < n {
+		panic(badVn2)
+	}
+	if len(auxv) < nb {
+		panic(badAuxv)
+	}
+
+	lastrk := min(m, n+offset)
+	lsticc := -1
+	tol3z := math.Sqrt(dlamchE)
+
+	bi := blas64.Implementation()
+
+	var k, rk int
+	for ; k < nb && lsticc == -1; k++ {
+		rk = offset + k
+
+		// Determine kth pivot column and swap if necessary.
+		p := k + bi.Idamax(n-k, vn1[k:], 1)
+		if p != k {
+			bi.Dswap(m, a[p:], lda, a[k:], lda)
+			bi.Dswap(k, f[p*ldf:], 1, f[k*ldf:], 1)
+			jpvt[p], jpvt[k] = jpvt[k], jpvt[p]
+			vn1[p] = vn1[k]
+			vn2[p] = vn2[k]
+		}
+
+		// Apply previous Householder reflectors to column K:
+		//
+		// A[rk:m, k] = A[rk:m, k] - A[rk:m, 0:k-1]*F[k, 0:k-1]^T.
+		if k > 0 {
+			bi.Dgemv(blas.NoTrans, m-rk, k, -1,
+				a[rk*lda:], lda,
+				f[k*ldf:], 1,
+				1,
+				a[rk*lda+k:], lda)
+		}
+
+		// Generate elementary reflector H_k.
+		if rk < m-1 {
+			a[rk*lda+k], tau[k] = impl.Dlarfg(m-rk, a[rk*lda+k], a[(rk+1)*lda+k:], lda)
+		} else {
+			tau[k] = 0
+		}
+
+		akk := a[rk*lda+k]
+		a[rk*lda+k] = 1
+
+		// Compute kth column of F:
+		//
+		// Compute F[k+1:n, k] = tau[k]*A[rk:m, k+1:n]^T*A[rk:m, k].
+		if k < n-1 {
+			bi.Dgemv(blas.Trans, m-rk, n-k-1, tau[k],
+				a[rk*lda+k+1:], lda,
+				a[rk*lda+k:], lda,
+				0,
+				f[(k+1)*ldf+k:], ldf)
+		}
+
+		// Padding F[0:k, k] with zeros.
+		for j := 0; j < k; j++ {
+			f[j*ldf+k] = 0
+		}
+
+		// Incremental updating of F:
+		//
+		// F[0:n, k] := F[0:n, k] - tau[k]*F[0:n, 0:k-1]*A[rk:m, 0:k-1]^T*A[rk:m,k].
+		if k > 0 {
+			bi.Dgemv(blas.Trans, m-rk, k, -tau[k],
+				a[rk*lda:], lda,
+				a[rk*lda+k:], lda,
+				0,
+				auxv, 1)
+			bi.Dgemv(blas.NoTrans, n, k, 1,
+				f, ldf,
+				auxv, 1,
+				1,
+				f[k:], ldf)
+		}
+
+		// Update the current row of A:
+		//
+		// A[rk, k+1:n] = A[rk, k+1:n] - A[rk, 0:k]*F[k+1:n, 0:k]^T.
+		if k < n-1 {
+			bi.Dgemv(blas.NoTrans, n-k-1, k+1, -1,
+				f[(k+1)*ldf:], ldf,
+				a[rk*lda:], 1,
+				1,
+				a[rk*lda+k+1:], 1)
+		}
+
+		// Update partial column norms.
+		if rk < lastrk-1 {
+			for j := k + 1; j < n; j++ {
+				if vn1[j] == 0 {
+					continue
+				}
+
+				// The following marked lines follow from the
+				// analysis in Lapack Working Note 176.
+				r := math.Abs(a[rk*lda+j]) / vn1[j] // *
+				temp := math.Max(0, 1-r*r)          // *
+				r = vn1[j] / vn2[j]                 // *
+				temp2 := temp * r * r               // *
+				if temp2 < tol3z {
+					// vn2 is used here as a collection of
+					// indices into vn2 and also a collection
+					// of column norms.
+					vn2[j] = float64(lsticc)
+					lsticc = j
+				} else {
+					vn1[j] *= math.Sqrt(temp) // *
+				}
+			}
+		}
+
+		a[rk*lda+k] = akk
+	}
+	kb = k
+	rk = offset + kb
+
+	// Apply the block reflector to the rest of the matrix:
+	//
+	// A[offset+kb+1:m, kb+1:n] := A[offset+kb+1:m, kb+1:n] - A[offset+kb+1:m, 1:kb]*F[kb+1:n, 1:kb]^T.
+	if kb < min(n, m-offset) {
+		bi.Dgemm(blas.NoTrans, blas.Trans,
+			m-rk, n-kb, kb, -1,
+			a[rk*lda:], lda,
+			f[kb*ldf:], ldf,
+			1,
+			a[rk*lda+kb:], lda)
+	}
+
+	// Recomputation of difficult columns.
+	for lsticc >= 0 {
+		itemp := int(vn2[lsticc])
+
+		// NOTE: The computation of vn1[lsticc] relies on the fact that
+		// Dnrm2 does not fail on vectors with norm below the value of
+		// sqrt(dlamchS)
+		v := bi.Dnrm2(m-rk, a[rk*lda+lsticc:], lda)
+		vn1[lsticc] = v
+		vn2[lsticc] = v
+
+		lsticc = itemp
+	}
+
+	return kb
+}
--- a/testlapack/dgeqp3.go
+++ b/testlapack/dgeqp3.go
@@ -1,0 +1,135 @@
+// Copyright ©2015 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package testlapack
+
+import (
+	"math"
+	"math/rand"
+	"testing"
+
+	"github.com/gonum/blas"
+	"github.com/gonum/blas/blas64"
+)
+
+type Dgeqp3er interface {
+	Dlapmter
+	Dgeqp3(m, n int, a []float64, lda int, jpvt []int, tau, work []float64, lwork int)
+}
+
+func Dgeqp3Test(t *testing.T, impl Dgeqp3er) {
+	rnd := rand.New(rand.NewSource(1))
+	for c, test := range []struct {
+		m, n, lda int
+	}{
+		{1, 1, 0},
+		{2, 2, 0},
+		{3, 2, 0},
+		{2, 3, 0},
+		{1, 12, 0},
+		{2, 6, 0},
+		{3, 4, 0},
+		{4, 3, 0},
+		{6, 2, 0},
+		{12, 1, 0},
+		{1, 1, 20},
+		{2, 2, 20},
+		{3, 2, 20},
+		{2, 3, 20},
+		{1, 12, 20},
+		{2, 6, 20},
+		{3, 4, 20},
+		{4, 3, 20},
+		{6, 2, 20},
+		{12, 1, 20},
+		{129, 256, 0},
+		{256, 129, 0},
+		{129, 256, 266},
+		{256, 129, 266},
+	} {
+		n := test.n
+		m := test.m
+		lda := test.lda
+		if lda == 0 {
+			lda = test.n
+		}
+		const (
+			all = iota
+			some
+			none
+		)
+		for _, free := range []int{all, some, none} {
+			a := make([]float64, m*lda)
+			for i := range a {
+				a[i] = rnd.Float64()
+			}
+			aCopy := make([]float64, len(a))
+			copy(aCopy, a)
+			jpvt := make([]int, n)
+			for j := range jpvt {
+				switch free {
+				case all:
+					jpvt[j] = -1
+				case some:
+					jpvt[j] = rnd.Intn(2) - 1
+				case none:
+					jpvt[j] = 0
+				default:
+					panic("bad freedom")
+				}
+			}
+			k := min(m, n)
+			tau := make([]float64, k)
+			for i := range tau {
+				tau[i] = rnd.Float64()
+			}
+			work := make([]float64, 1)
+			impl.Dgeqp3(m, n, a, lda, jpvt, tau, work, -1)
+			lwork := int(work[0])
+			work = make([]float64, lwork)
+			for i := range work {
+				work[i] = rnd.Float64()
+			}
+			impl.Dgeqp3(m, n, a, lda, jpvt, tau, work, lwork)
+
+			// Test that the QR factorization has completed successfully. Compute
+			// Q based on the vectors.
+			q := constructQ("QR", m, n, a, lda, tau)
+
+			// Check that q is orthonormal
+			for i := 0; i < m; i++ {
+				nrm := blas64.Nrm2(m, blas64.Vector{Inc: 1, Data: q.Data[i*m:]})
+				if math.Abs(nrm-1) > 1e-13 {
+					t.Errorf("Case %v, q not normal", c)
+				}
+				for j := 0; j < i; j++ {
+					dot := blas64.Dot(m, blas64.Vector{Inc: 1, Data: q.Data[i*m:]}, blas64.Vector{Inc: 1, Data: q.Data[j*m:]})
+					if math.Abs(dot) > 1e-14 {
+						t.Errorf("Case %v, q not orthogonal", c)
+					}
+				}
+			}
+			// Check that A * P = Q * R
+			r := blas64.General{
+				Rows:   m,
+				Cols:   n,
+				Stride: n,
+				Data:   make([]float64, m*n),
+			}
+			for i := 0; i < m; i++ {
+				for j := i; j < n; j++ {
+					r.Data[i*n+j] = a[i*lda+j]
+				}
+			}
+			got := nanGeneral(m, n, lda)
+			blas64.Gemm(blas.NoTrans, blas.NoTrans, 1, q, r, 0, got)
+
+			want := blas64.General{Rows: m, Cols: n, Stride: lda, Data: aCopy}
+			impl.Dlapmt(true, want.Rows, want.Cols, want.Data, want.Stride, jpvt)
+			if !equalApproxGeneral(got, want, 1e-13) {
+				t.Errorf("Case %v,  Q*R != A*P\nQ*R=%v\nA*P=%v", c, got, want)
+			}
+		}
+	}
+}
--- a/testlapack/dlaqp2.go
+++ b/testlapack/dlaqp2.go
@@ -1,0 +1,115 @@
+// Copyright ©2017 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package testlapack
+
+import (
+	"fmt"
+	"math"
+	"testing"
+
+	"github.com/gonum/blas"
+	"github.com/gonum/blas/blas64"
+)
+
+type Dlaqp2er interface {
+	Dlapmter
+	Dlaqp2(m, n, offset int, a []float64, lda int, jpvt []int, tau, vn1, vn2, work []float64)
+}
+
+func Dlaqp2Test(t *testing.T, impl Dlaqp2er) {
+	for ti, test := range []struct {
+		m, n, offset int
+	}{
+		{m: 4, n: 3, offset: 0},
+		{m: 4, n: 3, offset: 2},
+		{m: 4, n: 3, offset: 4},
+		{m: 3, n: 4, offset: 0},
+		{m: 3, n: 4, offset: 1},
+		{m: 3, n: 4, offset: 2},
+		{m: 8, n: 3, offset: 0},
+		{m: 8, n: 3, offset: 4},
+		{m: 8, n: 3, offset: 8},
+		{m: 3, n: 8, offset: 0},
+		{m: 3, n: 8, offset: 1},
+		{m: 3, n: 8, offset: 2},
+		{m: 10, n: 10, offset: 0},
+		{m: 10, n: 10, offset: 5},
+		{m: 10, n: 10, offset: 10},
+	} {
+		m := test.m
+		n := test.n
+		jpiv := make([]int, n)
+
+		for _, extra := range []int{0, 11} {
+			a := zeros(m, n, n+extra)
+			c := 1
+			for i := 0; i < m; i++ {
+				for j := 0; j < n; j++ {
+					a.Data[i*a.Stride+j] = float64(c)
+					c++
+				}
+			}
+			aCopy := cloneGeneral(a)
+			for j := range jpiv {
+				jpiv[j] = j
+			}
+
+			tau := make([]float64, n)
+			vn1 := columnNorms(m, n, a.Data, a.Stride)
+			vn2 := columnNorms(m, n, a.Data, a.Stride)
+			work := make([]float64, n)
+
+			impl.Dlaqp2(m, n, test.offset, a.Data, a.Stride, jpiv, tau, vn1, vn2, work)
+
+			prefix := fmt.Sprintf("Case %v (offset=%t,m=%v,n=%v,extra=%v)", ti, test.offset, m, n, extra)
+			if !generalOutsideAllNaN(a) {
+				t.Errorf("%v: out-of-range write to A", prefix)
+			}
+
+			if test.offset == m {
+				continue
+			}
+
+			mo := m - test.offset
+			q := constructQ("QR", mo, n, a.Data[test.offset*a.Stride:], a.Stride, tau)
+			// Check that q is orthonormal
+			for i := 0; i < mo; i++ {
+				nrm := blas64.Nrm2(mo, blas64.Vector{Inc: 1, Data: q.Data[i*mo:]})
+				if math.Abs(nrm-1) > 1e-13 {
+					t.Errorf("Case %v, q not normal", ti)
+				}
+				for j := 0; j < i; j++ {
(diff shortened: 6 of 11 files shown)


cgo/lapack.go
```diff
@@ -108,6 +111,51 @@ type Implementation struct{}
 
 var _ lapack.Float64 = Implementation{}
 
+// Dgeqp3 computes a QR factorization with column pivoting of the
+// m×n matrix A: A*P = Q*R using Level 3 BLAS.
+//
+// The matrix Q is represented as a product of elementary reflectors
+//  Q = H[0] H[1] . . . H[k-1], where k = min(m,n).
```
vladimir-ch: I think we have settled on using `[ ]` for indexing of vectors and matrices, and `_` for subscripting (as here).

cgo/lapack.go
```diff
@@ -108,6 +111,51 @@ type Implementation struct{}
...
+// Each H[i] has the form
+//  H[i] = I - tau * v * v**T
```
vladimir-ch: `^` instead of `**`

cgo/lapack.go
```diff
@@ -108,6 +111,51 @@ type Implementation struct{}
...
+// where tau is a real scalar, and v is a real/complex vector
```
vladimir-ch: v is a real vector

cgo/lapack.go
```diff
@@ -108,6 +111,51 @@ type Implementation struct{}
...
+// with v[0:i-1] = 0 and v[i] = 1; v[i:m] is stored on exit in
+// A[i:m, i], and tau in tau[i].
+//
+// jpvt specifies a column pivot to be applied to A. If
+// jpvt[j] is at least zero, the jth column of A is permuted
+// to the front of A*P (a leading column), otherwise the jth
+// column of A is a free column. On return, jpvt holds the
+// permutation that were applied; if jpvt[j] == k, then the jth
```
vladimir-ch: that was applied

cgo/lapack.go
```diff
@@ -108,6 +111,51 @@ type Implementation struct{}
...
+// column of A*P was the the kth column of A.
+// jpvt must have length n or Dgeqp3 will panic.
+//
+// On exit tau holds the scalar factors of the elementary reflectors.
+// It must have length n, otherwise Dgeqp3 will panic.
+//
+// work must have length max(1, lwork). otherwise Dgeqp3 will panic.
```
vladimir-ch: What value must `lwork` have? Can it be -1?

cgo/lapack.go
```diff
@@ -108,6 +111,51 @@ type Implementation struct{}
...
+	if len(tau) != min(m, n) {
+		panic(badTau)
+	}
+	if len(work) != max(1, lwork) {
+		panic(badWork)
+	}
+
+	jpvt32 := make([]int32, len(jpvt))
+	for i, v := range jpvt {
+		jpvt32[i] = int32(v + 1)
```
vladimir-ch: In the other PR you check for overflow. Should we do it here, too?

cgo/lapack.go
```diff
@@ -108,6 +111,51 @@ type Implementation struct{}
...
+	}
+
+	lapacke.Dgeqp3(m, n, a, lda, jpvt32, tau, work, lwork)
+}
```
vladimir-ch: I think you have to reflect jpvt32 vack into jpvt.

native/dgeqp3.go
```diff
@@ -0,0 +1,163 @@
...
+	sm := m - nxfd
+	sn := n - nxfd
+	sminmn := minmn - nxfd
+
+	// Determine the block size.
+	nb = impl.Ilaenv(inb, "DGEQRF", " ", sm, sn, -1, -1)
+	nbmin := 2
+	nx := 0
+
+	if nb > 1 && nb < sminmn {
```
vladimir-ch: `1 < nb` ?

native/dlaqp2.go
```diff
@@ -0,0 +1,112 @@
...
+)
+
+// Dlaqp2 computes a QR factorization with column pivoting of the block A[offset:m, 0:n]
+// of the m×n matrix A. The block A[0:offset, 0:n] is accordingly pivoted, but not factorized.
+//
+// On exit, the upper triangle of block A[offset:m, 0:n] is the triangular factor obtained.
+// The elements in block A[offset:m, 0:n] below the diagonal, together with tau, represent
+// the orthogonal matrix Q as a product of elementary reflectors.
+//
+// offset is number of rows of the matrix A that must be pivoted but no factorized.
```
vladimir-ch: not factorized

native/dlaqp2.go
```diff
@@ -0,0 +1,112 @@
...
+// On exit, if jpvt[i] = k, then the ith column of A*P was the kth column of A.
+// jpvt must have length n, otherwise Dlapq2 will panic.
+//
+// On exit tau holds the scalar factors of the elementary reflectors. It must have length
+// at least min(m, n) otherwise Dlapq2 will panic.
+//
+// vn1 and vn2 hold the partial and complete column norms respectively. They must have length n,
+// otherwise Dlapq2 will panic.
+//
+// work contains temporary storage and must have length m. Dlapq2 will panic if there is
```
vladimir-ch: In my PRs I stopped writing that work is a temporary memory, and just write how long it must be.

cgo/lapack.go
kortschak: Done.

cgo/lapack.go
kortschak: Done.

cgo/lapack.go
kortschak: Done.

cgo/lapack.go
kortschak: Done.

cgo/lapack.go
kortschak: Done.

There are a few other places that do not do that as well.

cgo/lapack.go
kortschak: Done. Thanks.

native/dgeqp3.go
kortschak: Not according to the Fortran.

cgo/lapack.go
kortschak: Yes. Thinking about how to word this. Will come back to it.

native/dlaqp2.go
kortschak: Done.

cgo/lapack.go
kortschak: Done.

native/dgeqp3.go
vladimir-ch: I just meant to cosmetically rearrange the condition to put `nb` in the middle.

kortschak: Ugh. Sorry. Fixed.

(I should try reading).

native/dgeqp3.go
```diff
@@ -0,0 +1,170 @@
...
+		nxfd++
+	}
+
+	// Factorize fixed columns.
+	//
+	// Compute the QR factorization of fixed columns and update remaining columns.
+	if nxfd >= 0 {
+		na := min(m, nxfd)
+		impl.Dgeqrf(m, na, a, lda, tau, work, lwork)
+		iws = max(iws, int(work[0]))
```
vladimir-ch: Our native implementation does not set work[0] to optimum lwork. That should be fixed. Issue filed.

native/dgeqp3.go
```diff
@@ -0,0 +1,170 @@
...
+		nx = max(0, impl.Ilaenv(ixover, "DGEQRF", " ", sm, sn, -1, -1))
+
+		if nx < sminmn {
+			// Determine if workspace is large enough for blocked code.
+			minws := 2*sn + (sn+1)*nb
+			iws = max(iws, minws)
+			if lwork < minws {
+				// Not enough workspace to use optimal nb. Reduce
+				// nb and determine the minimum value of nb.
+				nb = (lwork - 2*sn) / (sm + 1)
```
vladimir-ch: Fortran has `/ (sn + 1)`.

native/dlaqp2.go
```diff
@@ -0,0 +1,111 @@
...
+// On exit tau holds the scalar factors of the elementary reflectors. It must have length
+// at least min(m, n) otherwise Dlapq2 will panic.
+//
+// vn1 and vn2 hold the partial and complete column norms respectively. They must have length n,
+// otherwise Dlapq2 will panic.
+//
+// work must have length m, otherwise Dlapq2 will panic.
+//
+// Dlapq2 is an internal routine. It is exported for testing purposes.
+func (impl Implementation) Dlapq2(m, n, offset int, a []float64, lda int, jpvt []int, tau, vn1, vn2, work []float64) {
```
vladimir-ch: Dlaqp2 :-)
Also in the docs.

native/dlaqp2.go
```diff
@@ -0,0 +1,111 @@
...
+		if p != i {
+			bi.Dswap(m, a[p:], lda, a[i:], lda)
+			jpvt[p], jpvt[i] = jpvt[i], jpvt[p]
+			vn1[p] = vn1[i]
+			vn2[p] = vn2[i]
+		}
+
+		// Generate elementary reflector H[i].
+		if offpi < m {
+			_, tau[i] = impl.Dlarfg(m-offpi, a[offpi*lda+i], a[offpi*lda+i:], lda)
```
vladimir-ch: 1) Fortran's DLARFG modifies the second argument, so I think this should not ignore the first return value.
2) One of the indices into a seems to be wrong, compare with Fortran.

native/dlaqp2.go
```diff
@@ -0,0 +1,111 @@
...
+		p := i + bi.Idamax(n-i, vn1[i:], 1)
+		if p != i {
+			bi.Dswap(m, a[p:], lda, a[i:], lda)
+			jpvt[p], jpvt[i] = jpvt[i], jpvt[p]
+			vn1[p] = vn1[i]
+			vn2[p] = vn2[i]
+		}
+
+		// Generate elementary reflector H[i].
+		if offpi < m {
```
vladimir-ch: Could you please recheck that this condition is correct? I honestly don't know, so I prefer to be careful. `offpi` is compared to number of rows, and is used both as a number and as an index.

native/dlaqp2.go
```diff
@@ -0,0 +1,111 @@
...
+			_, tau[i] = impl.Dlarfg(m-offpi, a[offpi*lda+i], a[offpi*lda+i:], lda)
+		} else {
+			_, tau[i] = impl.Dlarfg(1, a[m*lda+i], a[m*lda+i:], lda)
```
vladimir-ch: This should definitely be:
```
a[m*lda+i], tau[i] = impl.Dlarfg(1, a[m*lda+i], a[m*lda+i:], lda)
```
This is a nice example where strict checking in Dlarfg would raise alarm and draw attention to this. For `n=1` the slice is not referenced. On the other hand, is it a reason to panic if the caller unnecessarily passes a slices that is not modified? In fact, we could even not call Dlarfg, and just set tau[i] to 0.

cgo/lapack.go
```diff
@@ -108,6 +111,65 @@ type Implementation struct{}
 
 var _ lapack.Float64 = Implementation{}
 
+// Dgeqp3 computes a QR factorization with column pivoting of the
+// m×n matrix A: A*P = Q*R using Level 3 BLAS.
+//
+// The matrix Q is represented as a product of elementary reflectors
+//  Q = H_0 H_1 . . . H_k-1, where k = min(m,n).
+// Each H[i] has the form
```
kortschak: Done.

cgo/lapack.go
```diff
@@ -108,6 +111,65 @@ type Implementation struct{}
...
+//  H[i] = I - tau * v * v^T
+// where tau is a real vector, and v is a real/complex vector
```
kortschak: Done.

cgo/lapack.go
```diff
@@ -108,6 +111,65 @@ type Implementation struct{}
...
+// jpvt specifies a column pivot to be applied to A. If
+// jpvt[j] is at least zero, the jth column of A is permuted
+// to the front of A*P (a leading column), otherwise the jth
+// column of A is a free column. On return, jpvt holds the
+// permutation that was applied; if jpvt[j] == k, then the jth
+// column of A*P was the the kth column of A.
+// jpvt must have length n or Dgeqp3 will panic.
+//
+// On exit tau holds the scalar factors of the elementary reflectors.
+// It must have length n, otherwise Dgeqp3 will panic.
```
kortschak: Done.

cgo/lapack.go
```diff
@@ -108,6 +111,65 @@ type Implementation struct{}
...
+//
+// work must have length max(1, lwork). otherwise Dgeqp3 will panic.
+// work must have length at least max(1,lwork), and lwork must be at least
+// 3*n+1. For optimal performance lwork must be at least 2*n+(n+1)*nb, where
+// nb is the optimal blocksize.On return, work[0] will contain the optimal
```
kortschak: Done.

cgo/lapack.go
```diff
@@ -108,6 +111,65 @@ type Implementation struct{}
...
+// value of lwork.
+//
+// If lwork == -1, instead of performing Dgeqp3, only the optimal value of lwork
+// will be stored in work[0].
+//
+// Dgeqp3 is an internal routine. It is exported for testing purposes.
+func (impl Implementation) Dgeqp3(m, n int, a []float64, lda int, jpvt []int, tau, work []float64, lwork int) {
+	checkMatrix(m, n, a, lda)
```
kortschak: The rational I have for checking early is to fail if the parameters are wrong up front. I'm happy to change this, but I don't see a problem with this approach (I prefer it).

cgo/lapack.go
```diff
@@ -108,6 +111,65 @@ type Implementation struct{}
...
+	if len(jpvt) != n {
+		panic(badIpiv)
+	}
+	if len(tau) < min(m, n) {
```
kortschak: The issue here is that `badTau` talks about "insufficient" length. That should be fixed.

#226

kortschak: Done.

cgo/lapack.go
```diff
@@ -108,6 +111,65 @@ type Implementation struct{}
...
+	if len(work) < max(1, lwork) {
+		panic(badWork)
+	}
+
+	jpvt32 := make([]int32, len(jpvt))
+	for i, v := range jpvt {
+		if v+1 != int(int32(v+1)) {
+			panic("lapack: jpvt element out of range")
+		}
+		jpvt32[i] = int32(v + 1)
```
kortschak: What would you prefer?

The options as I see them are:
1. Change the Go code to only accept -1. This has two possible index error handling approaches: either allow runtime panics for negatives != -1 or precheck for those and give a lapack panic.
2. Massage the jpvt parameter during the cgo call so that all negatives are passed to lapacke as 0.
3. Leave the native implementation as it is, but specify the current Fortran behaviour in both.

native/dgeqp3.go
```diff
@@ -0,0 +1,170 @@
...
+// license that can be found in the LICENSE file.
+
+package native
+
+import (
+	"github.com/gonum/blas"
+	"github.com/gonum/blas/blas64"
+)
+
+// Dgeqp3 computes a QR factorization with column pivoting of the
```
kortschak: Done.

native/dgeqp3.go
```diff
@@ -0,0 +1,170 @@
...
+	work[0] = float64(lwkopt)
+
+	if lwork == -1 {
+		return
+	}
+
+	bi := blas64.Implementation()
+
+	// Move initial columns up front.
+	nxfd := 0
```
kortschak: Ahh, thank you.

Done.

native/dgeqp3.go
kortschak: OK

native/dgeqp3.go
kortschak: Done.

native/dlaqp2.go
kortschak: Done.

native/dlaqp2.go
kortschak: I think it is. The changes that I have made bring `offpi` into zero-based indexing, so the number and index are now in the same space. The key is looking at the case where the condition is true - the check is looking for bounds.

Having said all that, changing it to `offpi <= m` makes no apparent difference to the test status.

@btracey Your thoughts?

native/dlaqp2.go
kortschak: Done.

native/dlaqp2.go
kortschak: 1. Fixed.
2. I don't think it is. This is the result of changing from 1-based/counting to just 0-based. If I have the +1 offset the code panics.

native/dlaqp2.go
vladimir-ch: If `offset` is 0 and `m <= n`, then `i == offpi` and it goes from 0 to `m-1`, so it is always less than `m`. The code must handle differently the case when there are no more rows below in the column `i` because then there is nothing to zero out for Dlarfg and `A( OFFPI+1, I )` would index outside of the matrix.

Please recheck but I _think_ that it _should_ be something like:
```
if offpi < m-1 {
    // Note the (offpi+1) in the third argument.
    a[offpi*lda+i], tau[i] = impl.Dlarfg(m-offpi, a[offpi*lda+i], a[(offpi+1)*lda+i:], lda)
} else {
    // Last row, we can call Dlarfg:
    // a[(m-1)*lda+i], tau[i] = impl.Dlarfg(1, a[(m-1)*lda+i], nil, lda)
    // or we can just equivalently do:
    tau[i] = 0
}
```

native/dlaqp2.go
```diff
@@ -0,0 +1,111 @@
...
+			a[offpi*lda+i], tau[i] = impl.Dlarfg(m-offpi, a[offpi*lda+i], a[offpi*lda+i:], lda)
+		} else {
+			a[m*lda+i], tau[i] = impl.Dlarfg(1, a[m*lda+i], a[m*lda+i:], lda)
+		}
+
+		if i < n {
+			// Apply H[i]**T to A[offset+i:m, i:n] from the left.
+			aii := a[offpi*lda+i]
+			a[offpi*lda+i] = 1
+			impl.Dlarf(blas.Left, m-offpi, n-i, a[offpi*lda+i:], lda, tau[i], a[offpi*lda+i:], lda, work)
```
vladimir-ch: 1) `i` is zero indexed, so the number of columns after the `i`th column is `n-i-1`.
2) The second slice should be `a[offpi*lda+i+1:]`

native/dlaqp2.go
```diff
@@ -0,0 +1,111 @@
...
+		}
+
+		// Generate elementary reflector H[i].
+		if offpi < m {
+			a[offpi*lda+i], tau[i] = impl.Dlarfg(m-offpi, a[offpi*lda+i], a[offpi*lda+i:], lda)
+		} else {
+			a[m*lda+i], tau[i] = impl.Dlarfg(1, a[m*lda+i], a[m*lda+i:], lda)
+		}
+
+		if i < n {
```
vladimir-ch: Similarly here. Unlike Fortran, `i` is always less than `n`. You are applying `H[i]^T` to the columns _after_ `i`, which makes sense only if there are any. So the test should be `if i < n-1`.

native/dlaqp2.go
```diff
@@ -0,0 +1,111 @@
...
+			// The following marked lines follow from the
+			// analysis in Lapack Working Note 176.
+			r := math.Abs(a[offpi*lda+j]) / vn1[j] // *
+			temp := math.Max(0, 1-r*r)             // *
+			r = vn1[j] / vn2[j]                    // *
+			temp2 := temp * r * r                  // *
+			if temp2 < tol3z {
+				var v float64
+				if offpi < m {
+					v = bi.Dnrm2(m-offpi, a[(offpi+1)*lda+j:], lda)
```
vladimir-ch: `m-offpi-1`

native/dlaqp2.go
```diff
@@ -0,0 +1,111 @@
...
+
+			// The following marked lines follow from the
+			// analysis in Lapack Working Note 176.
+			r := math.Abs(a[offpi*lda+j]) / vn1[j] // *
+			temp := math.Max(0, 1-r*r)             // *
+			r = vn1[j] / vn2[j]                    // *
+			temp2 := temp * r * r                  // *
+			if temp2 < tol3z {
+				var v float64
+				if offpi < m {
```
vladimir-ch: `if offpi < m-1`

testlapack/dgeqp3.go
```diff
@@ -0,0 +1,118 @@
...
+type Dgeqp3er interface {
+	Dgeqp3(m, n int, a []float64, lda int, jpvt []int, tau, work []float64, lwork int)
+}
+
+func Dgeqp3Test(t *testing.T, impl Dgeqp3er) {
+	rnd := rand.New(rand.NewSource(1))
+	for c, test := range []struct {
+		m, n, lda int
+	}{
+		{1, 1, 0},
```
vladimir-ch: Skipping Dlaqps for the moment, I think that these matrix sizes are too small to trigger the blocked code. Have you checked the coverage if you test only Dgeqp3, and Ilaenv for the minimum block size?

cgo/lapack.go
vladimir-ch: Does any of these options fit better into what the calling routine does?

native/dgeqp3.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+
+	// Initialize partial column norms.
+	// The first n elements of work store the exact column norms.
+	for j := nfxd; j < n; j++ {
+		work[j] = bi.Dnrm2(sm, a[nfxd*lda+j:], lda)
+		work[n+j] = work[j]
+	}
+	j := nfxd
+	if nb >= nbmin && nb < sminmn && nx < sminmn {
```
vladimir-ch: Put `nb` in the middle?

native/dlaqps.go
```diff
@@ -0,0 +1,208 @@
...
+import (
+	"math"
+
+	"github.com/gonum/blas"
+	"github.com/gonum/blas/blas64"
+)
+
+// Dlaqps computes a step of QR factorization with column pivoting
+// of an m×n matrix A by using Blas-3.  It tries to factorize
+// NB columns from A starting from the row OFFSET, and updates all
```
vladimir-ch: Lower case NB and OFFSET.

kortschak: Done.

native/dgeqp3.go
kortschak: Done.

native/dlaqp2.go
kortschak: Done.

native/dlaqp2.go
kortschak: Done.

native/dlaqp2.go
kortschak: 1. Yes.
2. Done.

Thanks.

native/dlaqp2.go
kortschak: Thanks, you are right.

Done.

native/dlaqp2.go
kortschak: That (either of those - so I chose the cheaper) works (in that it doesn't panic).

testlapack/dgeqp3.go
kortschak: DGEQRF has a minimum block size of 2. DGEQP3 does not define that for itself.

native/dlaqps.go
```diff
@@ -0,0 +1,208 @@
...
+			_, tau[k] = impl.Dlarfg(1, a[rk*lda+k], a[rk*lda+k:], lda)
+		}
+
+		akk := a[rk*lda+k]
+		a[rk*lda+k] = 1
+
+		// Compute kth column of F:
+		//
+		// Compute F[k+1:n, k] = tau[k]*A[rk:m, k+1:n]**T*A[rk:m, k].
+		if k < n {
```
vladimir-ch: `k < n-1`

native/dlaqps.go
```diff
@@ -0,0 +1,208 @@
...
+			bi.Dgemv(blas.Trans, m-rk, n-k, tau[k],
```
vladimir-ch: `n-k-1`

native/dlaqps.go
```diff
@@ -0,0 +1,208 @@
...
+		}
+
+		// Padding F[0:k, k] with zeros.
+		for j := 0; j < k; j++ {
+			f[j*ldf+k] = 0
+		}
+
+		// Incremental updating of F:
+		//
+		// F[0:n, k] := F[0:n, k] - tau[k]*F[0:n, 0:k-1]*A[rk:m, 0:k-1]**T*A[rk:m,k].
```
vladimir-ch: `s/**/^`

native/dlaqps.go
```diff
@@ -0,0 +1,208 @@
...
+		if k > 0 {
+			bi.Dgemv(blas.Trans, m-rk, k-1, -tau[k],
```
vladimir-ch: s/k-1/k

native/dlaqps.go
```diff
@@ -0,0 +1,208 @@
...
+				a[rk*lda:], lda,
+				a[rk*lda+k:], lda,
+				0,
+				auxv, 1)
+			bi.Dgemv(blas.NoTrans, n, k-1, 1,
```
vladimir-ch: s/k-1/k

native/dlaqps.go
```diff
@@ -0,0 +1,208 @@
...
+				f, ldf,
+				auxv, 1,
+				1,
+				f[k:], ldf)
+		}
+
+		// Update the current row of A:
+		//
+		// A[rk, k+1:n] = A[rk, k+1:n] - A[rk, 0:k]*F[k+1:n, 0:k]**T.
```
vladimir-ch: `s/**/^`

kortschak:
I have pushed failing tests for Dgeqp3. The free column factorisation was not being exercised. It is clearly incorrect. What is more interesting is that it looks like the cgo code is also incorrect in how it returns the jpvt elements (or I have misunderstood the comments); in two cases the application of the pivot gives a permutation of the expected answer.

native/dlaqps.go
```diff
@@ -0,0 +1,208 @@
...
+		if k < n {
```
vladimir-ch: `k < n-1`

native/dlaqps.go
```diff
@@ -0,0 +1,208 @@
...
+			bi.Dgemv(blas.NoTrans, n-k, k, -1,
```
vladimir-ch: `n-k-1, k+1,`

native/dlaqps.go
```diff
@@ -0,0 +1,208 @@
...
+				f[(k+1)*ldf:], ldf,
+				a[rk*lda:], lda,
```
vladimir-ch: s/lda/1

native/dlaqps.go
```diff
@@ -0,0 +1,208 @@
...
+				1,
+				a[rk*lda+k+1:], 1)
+		}
+
+		// Update partial column norms.
+		if rk < lastrk {
```
vladimir-ch: I think that `rk < lastrk-1` but please check.

native/dlaqps.go
```diff
@@ -0,0 +1,208 @@
...
+// They must have length n, otherwise Dlapq2 will panic.
+//
+// auxv must have length nb, otherwise Dlaqps will panic.
+//
+// f and ldf represent an nb×nb matrix F such that F**T = L*Y**T*A. ldf
+// must be max(1, n), otherwise Dlaqps will panic.
+//
+// Dlaqps is an internal routine. It is exported for testing purposes.
+func (impl Implementation) Dlaqps(m, n, offset, nb int, a []float64, lda int, jpvt []int, tau, vn1, vn2, auxv, f []float64, ldf int) (kb int) {
+	checkMatrix(m, n, a, lda)
```
vladimir-ch: It could be a good idea to also check that offset and nb are sane.

native/dlaqps.go
```diff
@@ -0,0 +1,208 @@
...
+					lsticc = j
+				} else {
+					vn1[j] = vn1[j] * math.Sqrt(temp) // *
+				}
+			}
+		}
+
+		a[rk*lda+k] = akk
+	}
+	kb = k
```
vladimir-ch: Judging from the code below (you omit +1 consistently everywhere) I think you are aware of this but anyway: when the above loop terminates, in Fortran k will point to the last processed column (which is in the matrix), while here k will point to the next one (which may be outside). Here I assume that you will move incrementing k to the end of the loop body.

native/dlaqps.go
```diff
@@ -0,0 +1,208 @@
...
+		bi.Dgemm(blas.NoTrans, blas.Trans,
+			m-rk, n-kb, kb, -1,
+			a[rk*lda:], lda,
+			f[kb*ldf:], ldf,
+			1,
+			a[rk*lda+kb:], lda)
+	}
+
+	// Recomputation of difficult columns.
+	for lsticc >= 0 {
```
vladimir-ch: How can ever `lsticc` fall below zero?

cgo/lapack.go
kortschak: The only caller is dggsvp3 which explicitly sets elements that are free to 0. Do one, but do we just allow the runtime to panic, or explicitly check and panic ourselves?

native/dlaqps.go
```diff
@@ -0,0 +1,208 @@
...
+// Dlaqps computes a step of QR factorization with column pivoting
+// of an m×n matrix A by using Blas-3.  It tries to factorize
+// NB columns from A starting from the row OFFSET, and updates all
+// of the matrix with Dgemm.
+//
+// In some cases, due to catastrophic cancellations, it cannot
+// factorize nb columns. Hence, the actual number of factorized
+// columns is returned in kb.
+//
+// Dlaqp2 computes a QR factorization with column pivoting of the
```
kortschak: Sorry.

Done.

native/dlaqps.go
```diff
@@ -0,0 +1,208 @@
...
+// block A[offset:m, 0:n] of the m×n matrix A. The block
+// A[0:offset, 0:n] is accordingly pivoted, but not factorized.
+//
+// On exit, the upper triangle of block A[offset:m, 0:n] is the
+// triangular factor obtained. The elements in block A[offset:m, 0:n]
+// below the diagonal, together with tau, represent the orthogonal
+// matrix Q as a product of elementary reflectors.
+//
+// offset is number of rows of the matrix A that must be pivoted but
+// no factorized. offset must not be negative otherwise Dlaqp2 will panic.
```
kortschak: Done... and the rest.

native/dlaqps.go
```diff
@@ -0,0 +1,208 @@
...
+//
+// On exit tau holds the scalar factors of the elementary reflectors.
+// It must have length n, otherwise Dlapq2 will panic.
+//
+// vn1 and vn2 hold the partial and complete column norms respectively.
+// They must have length n, otherwise Dlapq2 will panic.
+//
+// auxv must have length nb, otherwise Dlaqps will panic.
+//
+// f and ldf represent an nb×nb matrix F such that F**T = L*Y**T*A. ldf
```
kortschak: I can find no description anywhere for these.

Done

native/dlaqps.go
kortschak: Done.

native/dlaqps.go
```diff
@@ -0,0 +1,208 @@
...
+// must be max(1, n), otherwise Dlaqps will panic.
+//
+// Dlaqps is an internal routine. It is exported for testing purposes.
+func (impl Implementation) Dlaqps(m, n, offset, nb int, a []float64, lda int, jpvt []int, tau, vn1, vn2, auxv, f []float64, ldf int) (kb int) {
+	checkMatrix(m, n, a, lda)
+	checkMatrix(nb, nb, f, ldf)
+	if len(jpvt) != n {
+		panic(badIpiv)
+	}
+	if len(tau) < n {
```
kortschak: At the moment the panic is "lapack: tau has insufficient length". These are part of #228.

native/dlaqps.go
```diff
@@ -0,0 +1,208 @@
...
+		panic(badTau)
+	}
+	if len(vn1) < n {
```
kortschak: #228

native/dlaqps.go
```diff
@@ -0,0 +1,208 @@
...
+		panic(badVn1)
+	}
+	if len(vn2) < n {
```
kortschak: #228

native/dlaqps.go
```diff
@@ -0,0 +1,208 @@
...
+		panic(badVn2)
+	}
+	if len(auxv) < nb {
```
kortschak: #228

native/dlaqps.go
```diff
@@ -0,0 +1,208 @@
...
+	}
+
+	lastrk := min(m, n+offset)
+	lsticc := 0
+	k := 0
+	tol3z := math.Sqrt(dlamchE)
+
+	bi := blas64.Implementation()
+
+	var rk int
```
kortschak: Done.

native/dlaqps.go
```diff
@@ -0,0 +1,208 @@
...
+	for k < nb && lsticc == 0 {
+		k++
```
kortschak: Done.

native/dlaqps.go
```diff
@@ -0,0 +1,208 @@
...
+			bi.Dswap(m, a[p:], lda, a[k:], lda)
+			bi.Dswap(k, f[p*ldf:], 1, f[k*ldf:], 1)
+			jpvt[p], jpvt[k] = jpvt[k], jpvt[p]
+			vn1[p] = vn1[k]
+			vn2[p] = vn2[k]
+		}
+
+		// Apply previous Householder reflectors to column K:
+		//
+		// A[rk:m, k] = A[rk:m, k] - A[rk:m, 0:k-1]*F[k, 0:k-1]**T.
```
kortschak: Done.

native/dlaqps.go
```diff
@@ -0,0 +1,208 @@
...
+		if k > 0 {
+			bi.Dgemv(blas.NoTrans, m-rk, k-1, -1,
+				a[rk*lda:], lda,
+				f[k*ldf:], ldf,
```
kortschak: Thanks. Done.

native/dlaqps.go
```diff
@@ -0,0 +1,208 @@
...
+				1,
+				a[rk*lda+k:], 1)
+		}
+
+		// Generate elementary reflector H[k].
+		if rk < m {
```
kortschak: Done.

native/dlaqps.go
kortschak: Done.

native/dlaqps.go
kortschak: Done.

native/dlaqps.go
kortschak: Done, and above.

native/dlaqps.go
kortschak: Done.

native/dlaqps.go
kortschak: Done.

native/dlaqps.go
kortschak: Done.

native/dlaqps.go
kortschak: Done.

native/dlaqps.go
kortschak: Done.

native/dlaqps.go
kortschak: Done.

native/dlaqps.go
kortschak: That is now how it is.

native/dlaqps.go
kortschak: In the fortran it is initialised to 0 on the basis that that is an illegal index. I have changed the initialisation to -1.

native/dlaqps.go
kortschak: That is unclear to me either way. I will make the change and see how we go.

native/dlaqp2.go
```diff
@@ -0,0 +1,111 @@
...
+
+		// Generate elementary reflector H[i].
+		if offpi < m-1 {
+			a[offpi*lda+i], tau[i] = impl.Dlarfg(m-offpi-1, a[offpi*lda+i], a[(offpi+1)*lda+i:], lda)
+		} else {
+			tau[i] = 0
+		}
+
+		if i < n-1 {
+			// Apply H[i]^T to A[offset+i:m, i:n] from the left.
```
kortschak: Done.

native/dlaqp2.go
```diff
@@ -0,0 +1,111 @@
...
+		if p != i {
+			bi.Dswap(m, a[p:], lda, a[i:], lda)
+			jpvt[p], jpvt[i] = jpvt[i], jpvt[p]
+			vn1[p] = vn1[i]
+			vn2[p] = vn2[i]
+		}
+
+		// Generate elementary reflector H[i].
+		if offpi < m-1 {
+			a[offpi*lda+i], tau[i] = impl.Dlarfg(m-offpi-1, a[offpi*lda+i], a[(offpi+1)*lda+i:], lda)
```
kortschak: Done.

kortschak:
I'm at a loss for how to sort this out, so I'll leave it for the weekend. The main problem I'm having is that I don't see a way to know what the `ldf` should be - though there are clearly other issues.

btracey:
There's been a lot of messages on this. Should I be reviewing this?

vladimir-ch:
@btracey I went through the code completely (not the tests). I haven't checked yet the current status after my last night's batch of comments but I think you can definitely start reviewing it. What's needed now are solid tests, the implementation itself should not be far from correct.

native/dgeqp3.go
```diff
@@ -0,0 +1,168 @@
...
+			jb := min(nb, topbmn-j)
+
+			// Factorize jb columns among columns j:n.
+			fjb = impl.Dlaqps(m, n-j, j, jb, a[j:], lda, jpvt[j:], tau[j:],
+				work[j:j+n], work[j+n:j+2*n], work[2*n:2*n+jb], work[2*n+jb:], n-j)
+		}
+	}
+
+	// Use unblocked code to factor the last or only block.
+	if j <= minmn {
```
vladimir-ch: `j < minmn`

native/dgeqp3.go
```diff
@@ -0,0 +1,168 @@
...
+		impl.Dlaqp2(m, n-j, j, a[j:], lda, jpvt[j:], tau[j:],
```
vladimir-ch: The current test will complain that tau here is too short. I see the problem but not the solution right now.

native/dgeqp3.go
```diff
@@ -0,0 +1,168 @@
...
+		// Use blocked code initially.
+
+		// Compute factorization.
+		var fjb int
+		for topbmn := minmn - nx; j <= topbmn; j += fjb {
+			jb := min(nb, topbmn-j)
+
+			// Factorize jb columns among columns j:n.
+			fjb = impl.Dlaqps(m, n-j, j, jb, a[j:], lda, jpvt[j:], tau[j:],
+				work[j:j+n], work[j+n:j+2*n], work[2*n:2*n+jb], work[2*n+jb:], n-j)
```
vladimir-ch: Last argument should be `jb` (because F is (n-j)xjb and we are row-major).

cgo/lapack.go
```diff
@@ -108,6 +113,63 @@ type Implementation struct{}
 
 var _ lapack.Float64 = Implementation{}
 
+// Dgeqp3 computes a QR factorization with column pivoting of the
+// m×n matrix A: A*P = Q*R using Level 3 BLAS.
+//
+// The matrix Q is represented as a product of elementary reflectors
+//  Q = H_0 H_1 . . . H_k-1, where k = min(m,n).
```
kortschak: Done.

cgo/lapack.go
```diff
@@ -108,6 +113,63 @@ type Implementation struct{}
...
+// Each H_i has the form
+//  H_i = I - tau * v * v^T
+// where tau and v are real vectors with v[0:i-1] = 0 and v[i] = 1;
+// v[i:m] is stored on exit in A[i:m, i], and tau in tau[i].
+//
+// jpvt specifies a column pivot to be applied to A. If
+// jpvt[j] is at least zero, the jth column of A is permuted
+// to the front of A*P (a leading column), otherwise the jth
+// column of A is a free column. On return, jpvt holds the
+// permutation that was applied; if jpvt[j] == k, then the jth
```
kortschak: That's much nicer. Done.

native/dgeqp3.go
kortschak: I'm doing this as I go, so perhaps "Doing." would be better.

cgo/lapack.go
kortschak: OK

native/dgeqp3.go
kortschak: I thought that too, but work is not long enough then. I seem to be unable to do the simple arithmetic to resolve that.

native/dgeqp3.go
```diff
@@ -0,0 +1,168 @@
...
+		work[j] = bi.Dnrm2(sm, a[nfxd*lda+j:], lda)
+		work[n+j] = work[j]
+	}
+	j := nfxd
+	if nbmin <= nb && nb < sminmn && nx < sminmn {
+		// Use blocked code initially.
+
+		// Compute factorization.
+		var fjb int
+		for topbmn := minmn - nx; j <= topbmn; j += fjb {
```
kortschak: Yes. Thanks.

Done.

native/dgeqp3.go
kortschak: Done.

native/dgeqp3.go
kortschak: It does - I explored that and I also don't see the solution.

native/dlaqps.go
```diff
@@ -0,0 +1,212 @@
...
+//
+// auxv must have length nb, otherwise Dlaqps will panic.
+//
+// f and ldf represent an nb×nb matrix F that is overwritten during the
+// call. ldf must be max(nb, n), otherwise Dlaqps will panic.
+//
+// Dlaqps is an internal routine. It is exported for testing purposes.
+func (impl Implementation) Dlaqps(m, n, offset, nb int, a []float64, lda int, jpvt []int, tau, vn1, vn2, auxv, f []float64, ldf int) (kb int) {
+	checkMatrix(m, n, a, lda)
+	checkMatrix(nb, nb, f, ldf)
```
kortschak: Done.

native/dgeqp3.go
kortschak: That works now - presumably other changes you have suggested here not in place previously have fixed what I was bumping into.

native/dgeqp3.go
kortschak: The fix was to change the incorrect documentation in Dlaqp2 which was too strict. The index into tau in Dlaqp2 never goes above `min(m-offset, n)`, so make that claim on the docs and fix the check and we now pass.

native/dgeqp3.go
```diff
@@ -0,0 +1,173 @@
...
+			// Factorize jb columns among columns j:n.
+			fjb = impl.Dlaqps(m, n-j, j, jb, a[j:], lda, jpvt[j:], tau[j:],
+				work[j:j+n], work[j+n:j+2*n], work[2*n:2*n+jb], work[2*n+jb:], jb)
+		}
+	}
+
+	// Use unblocked code to factor the last or only block.
+	if j < minmn {
+		impl.Dlaqp2(m, n-j, j, a[j:], lda, jpvt[j:], tau[j:],
+			work[j:j+n], work[j+n:j+2*n], work[2*n:])
```
vladimir-ch: I think the work slices are wrong here. First, the last two slices overlap whenever `j!=0`. Second, the subslices are too long, they only need to have length `n-j` because that's the value we pass for `n` to Dlaqp2. The partition is important, it is used to hold the (partial) column norms. I think this line should be:
```
work[j:n], work[n+j:2*n], work[2*n:])
```
but please recheck.

Is this if branch covered by the tests? Perhaps you have to add more matrix sizes.

native/dgeqp3.go
```diff
@@ -0,0 +1,173 @@
...
+		// Use blocked code initially.
+
+		// Compute factorization.
+		var fjb int
+		for topbmn := minmn - nx; j < topbmn; j += fjb {
+			jb := min(nb, topbmn-j)
+
+			// Factorize jb columns among columns j:n.
+			fjb = impl.Dlaqps(m, n-j, j, jb, a[j:], lda, jpvt[j:], tau[j:],
+				work[j:j+n], work[j+n:j+2*n], work[2*n:2*n+jb], work[2*n+jb:], jb)
```
vladimir-ch: I think this has the same overlap problem as the call to Dlaqp2 below:
```
work[j:n], work[n+j:2*n], work[2*n:2*n+jb], work[2*n+jb:], jb)
```

native/dlaqps.go
```diff
@@ -0,0 +1,212 @@
...
+//
+// On exit tau holds the scalar factors of the elementary reflectors.
+// It must have length nb, otherwise Dlapqs will panic.
+//
+// vn1 and vn2 hold the partial and complete column norms respectively.
+// They must have length n, otherwise Dlapqs will panic.
+//
+// auxv must have length nb, otherwise Dlaqps will panic.
+//
+// f and ldf represent an nb×nb matrix F that is overwritten during the
```
vladimir-ch: Please fix the docs here about the size of `F` and `ldf`.

testlapack/dgeqp3.go
```diff
@@ -0,0 +1,152 @@
...
+		}
+		for _, free := range []bool{false, true} {
+			a := make([]float64, m*lda)
+			for i := range a {
+				a[i] = rnd.Float64()
+			}
+			aCopy := make([]float64, len(a))
+			copy(aCopy, a)
+			jpvt := make([]int, n)
+			for j := range jpvt {
```
vladimir-ch: It would be nice to have a test that has a mix of free and fixed columns ...

testlapack/dgeqp3.go
```diff
@@ -0,0 +1,152 @@
...
+					r.Data[i*n+j] = a[i*lda+j]
+				}
+			}
+			atmp := blas64.General{
+				Rows:   m,
+				Cols:   n,
+				Stride: lda,
+				Data:   make([]float64, m*lda),
+			}
+			copy(atmp.Data, a)
```
vladimir-ch: Isn't this copy unnecessary? You overwrite `atmp` in Gemm on the next line.

testlapack/dgeqp3.go
```diff
@@ -0,0 +1,152 @@
...
+			blas64.Gemm(blas.NoTrans, blas.NoTrans, 1, q, r, 0, atmp)
+
+			// Unpivot the result.
+			sort.Sort(byPivotIndex{rows: m, pvt: jpvt, data: atmp.Data, inc: lda})
```
vladimir-ch: For consistency and safety: `inc: atmp.Stride`

testlapack/dgeqp3.go
```diff
@@ -0,0 +1,152 @@
...
+
+			if !floats.EqualApprox(atmp.Data, aCopy, 1e-10) {
```
vladimir-ch: Can't you make the tolerance tighter?

testlapack/dlaqp2.go
```diff
@@ -0,0 +1,60 @@
...
+import (
+	"fmt"
+	"testing"
+)
+
+type Dlaqp2er interface {
+	Dlaqp2(m, n, offset int, a []float64, lda int, jpvt []int, tau, vn1, vn2, work []float64)
+}
+
+// TODO(kortschak) Make this a test of more than just indexing failure and "the vibe".
```
vladimir-ch: Are you working on this and `DlaqpsTest`?

testlapack/dgeqp3.go
```diff
@@ -0,0 +1,152 @@
...
+				a[i] = rnd.Float64()
+			}
+			aCopy := make([]float64, len(a))
+			copy(aCopy, a)
+			jpvt := make([]int, n)
+			for j := range jpvt {
+				if free {
+					jpvt[j] = -1
+				} else {
+					jpvt[j] = j
```
vladimir-ch: According to the docs and the code the actual value should not matter, so making it a constant or random value should work. I would change this especially if you are not going to mix free and fixed columns (I would mix them).

native/dlaqps.go
kortschak: Done - simplified to "f and ldf represent an n×nb matrix F that is overwritten during the call."

cgo/lapack.go
```diff
@@ -108,6 +113,70 @@ type Implementation struct{}
...
+		return
+	}
+
+	jpvt32 := make([]int32, len(jpvt))
+	for i, v := range jpvt {
+		v++
+		if v != int(int32(v)) || v < 0 || n < v {
+			panic("lapack: jpvt element out of range")
+		}
+		jpvt32[i] = int32(v + 1)
```
kortschak: Wow. That explains the weirdness I was seeing before. Yes, as you note in the test below, this doesn't really care.

native/dgeqp3.go
```diff
@@ -0,0 +1,173 @@
...
+		} else {
+			jpvt[j] = j
+		}
+		nfxd++
+	}
+
+	// Factorize nfxd columns.
+	//
+	// Compute the QR factorization of nfxd columns and update remaining columns.
+	if nfxd >= 0 {
```
kortschak: Done.

testlapack/dlaqp2.go
kortschak: I'm not sure what to do here. I was hoping some inspiration would come to me. That has not happened yet.

native/dgeqp3.go
kortschak: I think you are right, though also that it doesn't matter. The Fortran doesn't define the lengths (obviously) and sets the starts where they are here. But as you say they over provide so the bounds won't matter. I've fixed them though so that if things do go out of range we don't overwrite data. (The first upper bound I think is wrong in the line above thought, given the second parameter's lower bound).

native/dgeqp3.go
kortschak: Done, though `work[j:j+n]`.

testlapack/dgeqp3.go
kortschak: Done.

testlapack/dgeqp3.go
kortschak: Done.

testlapack/dgeqp3.go
kortschak: That uncovers a failure in the cases where `m=256, n=129`. The failure is also present with the cgo implementation, though it is a Heisenbug - changing the set of test cases changes the failure in a non-linear way. This needs some thought.

testlapack/dgeqp3.go
kortschak: Done.

testlapack/dgeqp3.go
vladimir-ch: Ah, you copy because of the out-of-matrix elements, don't you?

native/dgeqp3.go
vladimir-ch: Initially I also thought that the starts are where they are but then I realized there is a reason. In a for loop above work is initialized to column norms. There you can see that [0:n] and [n:2*n] hold the exact and partial column norms which is exactly what we want to pass to Dlaqp2 but only the remaining subslices starting at column j. If Dlaqp2 is correct, then it does not matter but we shouldn't really mix these two parts together. So I think my original suggestion is correct.

native/dlaqps.go
```diff
@@ -0,0 +1,212 @@
...
+				f[(k+1)*ldf:], ldf,
+				a[rk*lda:], 1,
+				1,
+				a[rk*lda+k+1:], 1)
+		}
+
+		// Update partial column norms.
+		if rk < lastrk-1 {
+			for j := k + 1; j < n; j++ {
+				if vn1[j] == -1 {
```
vladimir-ch: `== 0`

It's not about detecting invalid index (that comes later), rather avoiding division by zero.

native/dlaqp2.go
```diff
@@ -0,0 +1,111 @@
...
+			temp2 := temp * r * r                  // *
+			if temp2 < tol3z {
+				var v float64
+				if offpi < m-1 {
+					v = bi.Dnrm2(m-offpi-1, a[(offpi+1)*lda+j:], lda)
+				}
+				vn1[j] = v
+				vn2[j] = v
+			} else {
+				vn1[j] = vn1[j] * math.Sqrt(temp) // *
```
vladimir-ch: `*=`

native/dlaqps.go
```diff
@@ -0,0 +1,212 @@
...
+				// analysis in Lapack Working Note 176.
+				r := math.Abs(a[rk*lda+j]) / vn1[j] // *
+				temp := math.Max(0, 1-r*r)          // *
+				r = vn1[j] / vn2[j]                 // *
+				temp2 := temp * r * r               // *
+				if temp2 < tol3z {
+					vn2[j] = float64(lsticc)
+					lsticc = j
+				} else {
+					vn1[j] = vn1[j] * math.Sqrt(temp) // *
```
vladimir-ch: `*=`

native/dlaqps.go
vladimir-ch: I'm still afraid here that Fortan might use zero both as an invalid index and also as a true zero column norm. Is my fear void? Could you please check?

vladimir-ch: Ok, after checking the code I think that this is fine. But I'm still confused by the rounding. I know that Fortran does it but why? We start at the last difficult column, and step back until we find an invalid index. Wouldn't `int(vn2[lsticc])` work just as well? Does the rounding even work with our invalid index of -1?

native/dlaqps.go
vladimir-ch: Did you mean #226 ?

testlapack/dgeqp3.go
kortschak: Done.

native/dlaqps.go
kortschak: I think I should just go home.

Done.

native/dlaqp2.go
kortschak: Thanks.

Done.

native/dlaqps.go
kortschak: Done.

testlapack/dgeqp3.go
kortschak: That was what I was thinking, but that doesn't explain the current failures since now I'm using `equalApproxGeneral`. At some point the matrices are not being updated properly (they keep zeros, so I might use `got := nanGeneral(m, n, n)` instead of what I have now.

kortschak: Ahhh, it's a tolerance problem. 1e-14 is too tight. 1e-13 is OK.

native/dgeqp3.go
kortschak: OK.

Done.

native/dlaqps.go
kortschak: Arrrgh. Yes, yes I do. Sorry.

native/dlaqps.go
kortschak: I was really struggling with this, so I'm not sure I can give a good answer.

kortschak: When/if we do sort this out, I think it is worth expanding on the commentary here to make it clear what is happening.

kortschak: I am not sure how to get [that part of the code exercised](https://coveralls.io/builds/9935115/source?filename=native%2Fdlaqps.go#L199). We need to [get `temp2 < tol3z`](https://coveralls.io/builds/9935115/source?filename=native%2Fdlaqps.go#L172) so that `lsticc` is raised above -1. The [working note](www.netlib.org/lapack/lawnspdf/lawn176.pdf) presumably would help here. However, the details are beyond me.

vladimir-ch:
I'll try to look into it.

native/dlaqps.go
vladimir-ch: I'm still reading the paper but it seems to me that the rounding has nothing to do with the computation of the partial column norms. I think it is just to convert the column index from double to integer when we walk down the stack of difficult columns. I don't know enough Fortran to tell whether rounding is the right way to do it, but at least it works for them. On the other hand, I don't think the rounding works for us when we get to `lsticc == -1` because `int(-1+0.5)` is 0.

I'll finish the paper and one of the references to see if I can find some nice difficult testing matrix.

vladimir-ch: They generate special random matrices for which they can control the condition number. When the condition number is sufficiently high, the exceptional code path to compute exact column norms (and to terminate the computation of the current block) is more likely to be taken. It would be nice to replicate what they do but it requires some work, so it can be postponed to another PR. I am still convinced that the rounding has nothing to do with the working note, it is just to get back the integers stored in `vn2` when `temp2 < tol3z`. You can leave it in with a note and later test it with a properly bad matrix. Or you could raise `lsticc` above -1 artificially and see whether the `for lsticc >= 0` is an infinite loop or not.

native/dlaqps.go
```diff
@@ -0,0 +1,212 @@
...
+// of an m×n matrix A by using Blas-3.  It tries to factorize
+// nb columns from A starting from the row offset, and updates all
+// of the matrix with Dgemm.
+//
+// In some cases, due to catastrophic cancellations, it cannot
+// factorize nb columns. Hence, the actual number of factorized
+// columns is returned in kb.
+//
+// Dlaqps computes a QR factorization with column pivoting of the
+// block A[offset:m, 0:n] of the m×n matrix A. The block
```
vladimir-ch: In the first paragraph it says that only `nb` columns are factorized, so here it should be `A[offset:m, 0:nb]`.

native/dlaqps.go
```diff
@@ -0,0 +1,212 @@
...
+// A[0:offset, 0:n] is accordingly pivoted, but not factorized.
+//
+// On exit, the upper triangle of block A[offset:m, 0:n] is the
```
vladimir-ch: Only `kb` block is actually factorized, so in this paragraph it should say `A[offset:m, 0:kb]`

testlapack/dlaqp2.go
vladimir-ch: You could check that `A[offset:m, 0:n]` contains its QR decomposition. Similarly for Dlaqps but only for the block `A[offset:m, 0:kb]`.  It's not so much work, just copy, paste and adjust the test for Dgeqp3. It's up to you.

native/dlaqps.go
kortschak: Done.

native/dlaqps.go
kortschak: Done.

testlapack/dlaqp2.go
kortschak: I'll look at this later.

native/dlaqps.go
kortschak: >I am still convinced that the rounding has nothing to do with the working note, it is just to get back the integers stored in `vn2` when `temp2 < tol3z`.

Yes, sorry. What I meant there was that in order to be able to test this section the kind of matrix they are discussing seemed to be necessary.

Bumping it to greater than 1 does indeed result in an infinite loop.

kortschak:
f885a332860e95683118e6799651f4bed7b5d6e2 uncovers indexing errors in Dlaqps that were missed. I need to leave this for a while to let the frustration subside.

testlapack/dlaqp2.go
kortschak: Done for Dlaqp2, though that fails for `offset != 0`.

testlapack/dlaqp2.go
```diff
@@ -0,0 +1,132 @@
...
+				}
+				aCopy := cloneGeneral(a)
+				for j := range jpiv {
+					switch free {
+					case all:
+						jpiv[j] = -1
+					case some:
+						jpiv[j] = rnd.Intn(2) - 1
+					case none:
+						jpiv[j] = j
```
vladimir-ch: Since `Dlaqp2` works only on columns that are swappable and only updates jpiv, I think you can keep only this case and remove the others together with the corresponding `for` loop.

testlapack/dlaqp2.go
```diff
@@ -0,0 +1,132 @@
...
+					default:
+						panic("bad freedom")
+					}
+				}
+
+				tau := make([]float64, n)
+				vn1 := make([]float64, n)
```
vladimir-ch: Leaving `vn1` and `vn2` zeroed will not trigger the column norm updating branch in `Dlaqp2`. I think it would be good to initialize them to column norms using `blas64.Norm` as is done in `Dgeqp3`. I don't know if we can use the output for some kind of test check.

native/dlaqp2.go
```diff
@@ -0,0 +1,111 @@
...
+// On exit, if jpvt[i] = k, then the ith column of A*P was the kth column of A.
+// jpvt must have length n, otherwise Dlaqp2 will panic.
+//
+// On exit tau holds the scalar factors of the elementary reflectors. It must have length
+// at least min(m-offset, n) otherwise Dlaqp2 will panic.
+//
+// vn1 and vn2 hold the partial and complete column norms respectively. They must have length n,
+// otherwise Dlaqp2 will panic.
+//
+// work must have length m, otherwise Dlaqp2 will panic.
```
vladimir-ch: work must have length `n`

native/dlaqp2.go
```diff
@@ -0,0 +1,111 @@
...
+	if len(tau) < mn {
+		panic(badTau)
+	}
+	if len(vn1) < n {
+		panic(badVn1)
+	}
+	if len(vn2) < n {
+		panic(badVn2)
+	}
+	if len(work) < m {
```
vladimir-ch: `< n`

testlapack/dlaqp2.go
```diff
@@ -0,0 +1,132 @@
...
+	rnd := rand.New(rand.NewSource(1))
+	for ti, test := range []struct {
+		m, n, offset int
+	}{
+		{m: 4, n: 3, offset: 0},
+		// {m: 4, n: 3, offset: 2},
+		// {m: 4, n: 3, offset: 4},
+		{m: 10, n: 10, offset: 0},
+		// {m: 10, n: 10, offset: 5},
+		// {m: 10, n: 10, offset: 10},
```
vladimir-ch: I found the work slice issue with a `m < n` matrix, could you please add such a test case?

testlapack/dlaqp2.go
```diff
@@ -0,0 +1,132 @@
...
+				// Check that A * P = Q * R
+				r := blas64.General{
+					Rows:   mo,
+					Cols:   n,
+					Stride: n,
+					Data:   make([]float64, mo*n),
+				}
+				for i := 0; i < mo; i++ {
+					for j := i; j < n; j++ {
+						r.Data[i*n+j] = a.Data[i*a.Stride+j]
```
vladimir-ch: `r.Data[i*n+j] = a.Data[(test.offset+i)*a.Stride+j]`
Then loosen the tolerance below to 1e-12 and voilà.

kortschak: !

Done (and in dlaqps without the increase in tolerance)

native/dlaqp2.go
kortschak: Done.

native/dlaqp2.go
kortschak: Done.

testlapack/dlaqp2.go
kortschak: Done (also for Dlaqps)

testlapack/dlaqp2.go
kortschak: Done (also must be `m < n-1` otherwise `Dlarf` doesn't notice.

testlapack/dlaqp2.go
kortschak: Done.

The situation for Dlaqps is more subtle. I can exercise the code for difficult columns, but it results in a panic because the norms are outside the bounds of `vn2`.

vladimir-ch: What do you mean that the norms are outside the bounds? It should not happen that the norms are used as indices.

kortschak: `lsticc` is assigned values from `vn2`. If these values are at least `len(vn2)` we get a bounds panic when we do `itemp := int(vn2[lsticc] + 0.5)`. So the partial norms are constrained to be less than `float64(len(vn2)) - 0.5`.

You can see this when you instrument so:

```diff
diff --git a/native/dlaqps.go b/native/dlaqps.go
index ce8cdb9..647e5e1 100644
--- a/native/dlaqps.go
+++ b/native/dlaqps.go
@@ -5,6 +5,7 @@
 package native
 
 import (
+       "fmt"
        "math"
 
        "github.com/gonum/blas"
@@ -201,7 +202,9 @@ func (impl Implementation) Dlaqps(m, n, offset, nb int, a []float64, lda int, jp
        //
        // Recomputation of difficult columns.
        for lsticc >= 0 {
+               fmt.Println(vn2, lsticc)
                itemp := int(vn2[lsticc] + 0.5)
+               fmt.Println(itemp)
 
                // NOTE: The computation of vn1[lsticc] relies on the fact that
                // Dnrm2 does not fail on vectors with norm below the value of
```
and set `vn2` to `columnNorms(m, n, a.Data, a.Stride)` in the tests as I've done for dlaqp2.go
```
$ go test -run Dlaqps
[12.884098726725126 14.628738838327795 -1] 2
0
[12.884098726725126 14.628738838327795 0.6175483551795395] 0
13
[0.5670198402783161 14.628738838327795 0.6175483551795395] 13
--- FAIL: TestDlaqps (0.00s)
panic: runtime error: index out of range [recovered]
	panic: runtime error: index out of range
```

kortschak: I think I have figured this out. One part was an incorrect `incX`, but the other part is a result of the way the Fortran was hiding indices into the `vn2` array in the `vn2` array. There is horrible faking code doing that now which makes the tests pass and does exercise the difficult columns loop. The code that exists can be rewritten to use pure 0-based indexing, but I'm leaving it as it is for the moment for reading purposes.

native/dlaqps.go
```diff
@@ -0,0 +1,218 @@
...
+			m-rk, n-kb, kb, -1,
+			a[rk*lda:], lda,
+			f[kb*ldf:], ldf,
+			1,
+			a[rk*lda+kb:], lda)
+	}
+
+	// Recomputation of difficult columns.
+	for lsticc >= 0 {
+		itemp := int(vn2[lsticc]+0.5) - 1 // Convert vn2 value back to 0-based index.
```
vladimir-ch: I still don't understand the need for the rounding here. This loop only accesses elements of `vn2` where integers have been stored above. Why not just recover them by `int(vn2[lsticc])`?

kortschak: I don't think there is any need. Maybe the author of the Fortran just habitually reaches for nint when converting floats to ints. I can remove it.

vladimir-ch: I think that if you removed it, you could also remove the not-so-nice 0->1->0-based index manipulation.

kortschak: You are right.

Done.

btracey:
I didn't check the implementations, but the tests LGTM

kortschak:
*ping*

native/dlaqp2.go
```diff
@@ -0,0 +1,111 @@
...
+// of the m×n matrix A. The block A[0:offset, 0:n] is accordingly pivoted, but not factorized.
+//
+// On exit, the upper triangle of block A[offset:m, 0:n] is the triangular factor obtained.
+// The elements in block A[offset:m, 0:n] below the diagonal, together with tau, represent
+// the orthogonal matrix Q as a product of elementary reflectors.
+//
+// offset is number of rows of the matrix A that must be pivoted but not factorized.
+// offset must not be negative otherwise Dlaqp2 will panic.
+//
+// On exit, if jpvt[i] = k, then the ith column of A*P was the kth column of A.
```
vladimir-ch: `On exit, jpvt holds the permutation that was applied; the jth column of A*P was the jpvt[j] column of A.`

native/dlaqps.go
```diff
@@ -0,0 +1,216 @@
...
+//
+// On exit, the upper triangle of block A[offset:m, 0:kb] is the
+// triangular factor obtained. The elements in block A[offset:m, 0:n]
+// below the diagonal, together with tau, represent the orthogonal
+// matrix Q as a product of elementary reflectors.
+//
+// offset is number of rows of the matrix A that must be pivoted but
+// not factorized. offset must not be negative otherwise Dlaqps will panic.
+//
+// On exit, if jpvt[i] = k, then the ith column of A*P was the kth
```
vladimir-ch: `On exit, jpvt holds the permutation that was applied; the jth column of A*P was the jpvt[j] column of A.`

native/dlaqp2.go
kortschak: Done.

native/dlaqps.go
kortschak: Done.

PR #227, opened 2017-01-25
cgo,native: fix optimal work length return

### LINKED ISSUE #225: native: Dgeqrf does not return optimum lwork in work[0]

kortschak:

Fixes #225.

### DIFF
--- a/cgo/lapack.go
+++ b/cgo/lapack.go
@@ -894,18 +894,18 @@ func (impl Implementation) Dgeqr2(m, n int, a []float64, lda int, tau, work []fl
 //
 // The C interface does not support providing temporary storage. To provide compatibility
 // with native, lwork == -1 will not run Dgeqrf but will instead write the minimum
-// work necessary to work[0]. If len(work) < lwork, Dgeqrf will panic.
+// work necessary to work[0]. If len(work) < max(1, lwork), Dgeqrf will panic.
 //
 // tau must have length at least min(m,n), and this function will panic otherwise.
 func (impl Implementation) Dgeqrf(m, n int, a []float64, lda int, tau, work []float64, lwork int) {
+	if len(work) < max(1, lwork) {
+		panic(shortWork)
+	}
 	if lwork == -1 {
 		work[0] = float64(n)
 		return
 	}
 	checkMatrix(m, n, a, lda)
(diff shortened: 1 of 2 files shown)


native/dgeqrf.go
```diff
@@ -14,7 +14,7 @@ import (
 // parameters at entry and exit.
 //
 // work is temporary storage, and lwork specifies the usable memory length.
-// At minimum, lwork >= m and this function will panic otherwise.
+// At minimum, lwork >= max(1, m) and this function will panic otherwise.
 // Dgeqrf is a blocked QR factorization, but the block size is limited
```
vladimir-ch: `max(1,n)`

cgo/lapack.go
```diff
@@ -894,18 +894,18 @@ func (impl Implementation) Dgeqr2(m, n int, a []float64, lda int, tau, work []fl
 //
 // The C interface does not support providing temporary storage. To provide compatibility
```
vladimir-ch: We should remove these comments, they are not true anymore.

native/dgeqrf.go
```diff
@@ -25,14 +25,14 @@ func (impl Implementation) Dgeqrf(m, n int, a []float64, lda int, tau, work []fl
 	nb := impl.Ilaenv(1, "DGEQRF", " ", m, n, -1, -1)
 	lworkopt := n * max(nb, 1)
 	lworkopt = max(n, lworkopt)
+	if len(work) < max(1, lwork) {
```
vladimir-ch: You could do this check even before calling Ilaenv.

cgo/lapack.go
kortschak: Please file an issue.

native/dgeqrf.go
kortschak: Done.

I'm going to file an issue to consider moving checkMatrix calls higher universally as well. The decision may be not to, but I think it's something that should be discussed: #228.

native/dgeqrf.go
kortschak: Done. This required a moderate rewording since that would have been incorrect as well.

kortschak:
PTAL

vladimir-ch:
LGTM

PR #242, opened 2017-02-09
native,testlapack: some small cleanup in the eigensolver

vladimir-ch:
This has been on my todo list since I finished Dgeev. It is not essential but at least I can check it as done. PTAL

### DIFF
--- a/native/dlaqr04.go
+++ b/native/dlaqr04.go
@@ -419,9 +419,9 @@ func (impl Implementation) Dlaqr04(wantt, wantz bool, n, ilo, ihi int, h []float
 			// Shuffle shifts into pairs of real shifts and pairs of
 			// complex conjugate shifts using the fact that complex
 			// conjugate shifts are already adjacent to one another.
-			// TODO(vladimir-ch): Dlaqr5 also does this shuffling.
-			// It seems that the one in Dlaqr5 is redundant but
-			// check.
+			// TODO(vladimir-ch): The shuffling here could probably
+			// be removed but I'm not sure right now and it's safer
+			// to leave it.
 			for i := kbot; i > ks+1; i -= 2 {
 				if wi[i] == -wi[i-1] {
 					continue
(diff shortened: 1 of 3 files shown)


native/dlaqr04.go
```diff
@@ -419,9 +419,6 @@ func (impl Implementation) Dlaqr04(wantt, wantz bool, n, ilo, ihi int, h []float
 			// Shuffle shifts into pairs of real shifts and pairs of
 			// complex conjugate shifts using the fact that complex
 			// conjugate shifts are already adjacent to one another.
-			// TODO(vladimir-ch): Dlaqr5 also does this shuffling.
```
vladimir-ch: In fact, the shuffling here could _probably_ be removed but I'm not sure right now and it's safer to leave it.

native/dlaqr5.go
```diff
@@ -124,9 +124,8 @@ func (impl Implementation) Dlaqr5(wantt, wantz bool, kacc22 int, n, ktop, kbot,
 	}
 
 	// Note: lapack says that nshfts must be even but allows it to be odd
-	// anyway. At the moment, we panic above if nshfts is not even, so
-	// reducing it by one is unnecessary (it seems that Dlaqr0 and Dlaqr4
-	// indeed use only even nshfts).
+	// anyway. We panic above if nshfts is not even, so reducing it by one
```
vladimir-ch: We could remove the `nshfts` parameter, in Go it is not really needed.

kortschak: How would that work? It seems pretty integral from a brief reading here.

vladimir-ch: We could do `nshfts := len(sr)`. But but this is a very minor point not worth changing.

native/dlaqr04.go
btracey: Maybe change the TODO to say that instead?

btracey:
This LGTM

native/dlaqr04.go
vladimir-ch: Done.

PR #243, opened 2017-02-10
native: fix wrong stride in Dsteqr

vladimir-ch:
In the call to Dlascl we pretend that the diagonals d and e are 1-column
wide matrices. Reference LAPACK is column-major, so the value of the
stride should not matter, which is demonstrated by the (probably
arbitrary) value of n. We are row-major, so the stride should be set to
the number of columns, that is, 1.


btracey:
LGTM

PR #245, opened 2017-02-17
testlapack: add Dlatm1

vladimir-ch:
This is a first step in my effort on #232 . In Reference LAPACK, `testing/MATGEN` has routines that probably contain what we need. I'm trying to follow their interface although not as strictly as in native, we can deviate here and be more Go-like. Let me know if you think the random number part could be handled better. I also thought it might be a good idea to export these matrix generation functions.

### DIFF
--- a/testlapack/matgen.go
+++ b/testlapack/matgen.go
@@ -9,30 +9,30 @@ import (
 	"math/rand"
 )
 
-// Dlatm1 computes the entries of d as specified by mode, cond and rsign.
+// Dlatm1 computes the entries of dst as specified by mode, cond and rsign.
 //
-// mode describes how d will be computed:
-//  |mode| == 1: d[0] = 1 and d[1:n] = 1/cond
-//  |mode| == 2: d[:n-1] = 1/cond and d[n-1] = 1
-//  |mode| == 3: d[i] = cond^{-i/(n-1)}, i=0,...,n-1
-//  |mode| == 4: d[i] = 1 - i*(1-1/cond)/(n-1)
-//  |mode| == 5: d[i] = random number in the range (1/cond, 1) such that
+// mode describes how dst will be computed:
+//  |mode| == 1: dst[0] = 1 and dst[1:n] = 1/cond
+//  |mode| == 2: dst[:n-1] = 1/cond and dst[n-1] = 1
+//  |mode| == 3: dst[i] = cond^{-i/(n-1)}, i=0,...,n-1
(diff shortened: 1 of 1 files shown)


testlapack/matgen.go
```diff
@@ -0,0 +1,118 @@
...
+// or -1 with probability 0.5
+//
+// dist specifies the type of distribution to be used when mode == ±6:
+//  dist == 1: Uniform(0,1)
+//  dist == 2: Uniform(-1,1)
+//  dist == 3: Normal(0,1)
+// For other values of dist Dlatm1 will panic.
+//
+// rnd is used as a source of random numbers.
+func Dlatm1(mode int, cond float64, rsign bool, dist int, rnd *rand.Rand, d []float64) {
```
kortschak: If we're not following the LAPACK API tightly, move `d` to the front and rename to `dst`?

testlapack/matgen.go
```diff
@@ -0,0 +1,118 @@
...
+	if amode < 0 {
+		amode = -amode
+	}
+	if amode < 1 || 6 < amode {
+		panic("testlapack: invalid mode")
+	}
+	if cond < 1 {
+		panic("testlapack: cond < 1")
+	}
+	if dist != 1 && dist != 2 && dist != 3 {
```
kortschak: `amode == 6 && (dist < 1 || 3 < dist)`

testlapack/matgen.go
```diff
@@ -0,0 +1,118 @@
...
+				d[i] = 2*rnd.Float64() - 1
+			}
+		case 3:
+			for i := range d {
+				d[i] = rnd.NormFloat64()
+			}
+		}
+	}
+
+	if rsign && mode != -6 && mode != 6 {
```
kortschak: `rsign && amode != 0 && amode != 6`

testlapack/matgen.go

testlapack/matgen.go

testlapack/matgen.go

kortschak: Sorry, I missed that you don't have a valid mode for 0. Is there a reason for that?

vladimir-ch: Yea, I omitted it so that it doesn't complicate the docs. If mode == 0 does not modify dst, I don't see any value in having it except for following the reference.

kortschak: There may be a reason to add it later, but leaving it out now seems fine.

PR #251, opened 2017-03-02
native,testlapack: fix Dtrtri and tests for Dtrtri and Dtrti2

vladimir-ch:


kortschak:
Note missing word in second commit message.

vladimir-ch:
And I'll have to loosen the tolerance because of cgo.

PR #252, opened 2017-03-02
native: fix bad lwork in Dorgql

vladimir-ch:


kortschak:
This is related in form to the fix you identified for Dgerqf?

vladimir-ch:
Yes, and also elsewhere, but this I found independently when trying to increase coverage.

PR #253, opened 2017-03-03
native: add blas.Lower case to Dpotrf test

vladimir-ch:


testlapack/dpotrf.go
```diff
@@ -18,55 +18,107 @@ type Dpotrfer interface {
...
-		// Multiply u^T * u
-		ans := make([]float64, len(a))
-		bi.Dsyrk(blas.Upper, blas.Trans, n, n, 1, a, n, 0, ans, n)
 
-		match := true
-		for i := 0; i < n; i++ {
-			for j := i; j < n; j++ {
-				if !floats.EqualWithinAbsOrRel(ans[i*n+j], aCopy[i*n+j], 1e-14, 1e-14) {
-					match = false
+			if uplo == blas.Upper {
```
kortschak: A switch with a default panicking case here? Not necessary, but more elegant (?).

PR #254, opened 2017-03-03
testlapack: add Dlagsy

vladimir-ch:
Also, "test" the new function Dlagsy by using it to generate a not positive definite matrix for an expected-failure test case for Dpotrf. PTAL


vladimir-ch:
I would add it if and when needed. Now I'd like to get to Dlatms (and maybe Dlatme) as soon as possible to see how this experiment works out.

kortschak:
Maybe add the `k` parameter as a place holder then, but mark it as unused?

PR #255, opened 2017-03-04
testlapack: add unused parameter k to Dlagsy for future implementation

vladimir-ch:
@kortschak PTAL


kortschak:
Sorry, would you add the bounds checks for k that exist in the Fortran docs.

PR #256, opened 2017-03-04
testlapack: add Dlagge

vladimir-ch:


testlapack/matgen.go
```diff
@@ -183,6 +183,92 @@ func Dlagsy(n int, d []float64, a []float64, lda int, rnd *rand.Rand, work []flo
...
+				bi.Dscal(m-i-1, 1/wb, work[1:m-i], 1)
+				work[0] = 1
+				tau = wb / wa
+			}
+
+			// Multiply A[i:m,i:n] by random reflection from the left.
+			bi.Dgemv(blas.Trans, m-i, n-i,
+				1, a[i*lda+i:], lda, work[:m-i], 1,
+				0, work[m:m+n-i], 1)
+			bi.Dger(m-i, n-i,
```
kortschak: Can we have more consistent/meaningful line breaks here and below please.

testlapack/matgen.go
```diff
@@ -183,6 +183,92 @@ func Dlagsy(n int, d []float64, a []float64, lda int, rnd *rand.Rand, work []flo
...
 }
 
+// Dlagge generates a real general m×n matrix A, by pre- and post-multiplying
+// a real diagonal matrix D with random orthogonal matrices:
+//  A = U*D*V.
+//
+// d must have length min(m,n), and work must have length m+n, otherwise Dlagge
+// will panic.
+//
+// The parameters ku and kl are unused.
```
kortschak: Please document the bounds for `ku` and `kl` and check for them in the code.

testlapack/matgen.go
vladimir-ch: Do you have any suggestion? I'm trying to be consistent here: 1st line - matrix size, 2nd line - operation (e.g., scalar * mat * vec or scalar * vec * vec^T), 3rd line - destination.

vladimir-ch:
Regarding testing, I will add a test with D=I and m==n, and checking that the result is an orthogonal matrix. Not very general because it will not exercise all possible states but I can't think of anything better.

testlapack/matgen.go
kortschak: Ahh, OK. That works for me.

kortschak:
I think that's fine. It can be extended in the future.

vladimir-ch:
@kortschak PTAL. I smuggled in also a test for Dlagsy.

testlapack/matgen_test.go
```diff
@@ -0,0 +1,90 @@
...
+				Rows:   n,
+				Cols:   n,
+				Stride: lda,
+				Data:   nanSlice(n * lda),
+			}
+			work := make([]float64, a.Rows+a.Cols)
+
+			Dlagsy(a.Rows, 0, d, a.Data, a.Stride, rnd, work)
+
+			isIdentity := true
```
kortschak: Did you want to use this for a `t.Errorf` later?

vladimir-ch: Of course, thanks. Fixed and test passing.

PR #257, opened 2017-03-08
native: fix safe minimum and maximum values in Dlartg

vladimir-ch:
Also, modify Dlartg test to use tiny and huge input values to cover all code paths.

### DIFF
--- a/native/dlartg.go
+++ b/native/dlartg.go
@@ -17,7 +17,7 @@ import "math"
 func (impl Implementation) Dlartg(f, g float64) (cs, sn, r float64) {
 	safmin := dlamchS
 	eps := dlamchE
-	safmn2 := math.Pow(float64(dlamchB), math.Log(safmin/eps))
+	safmn2 := math.Pow(dlamchB, math.Trunc(math.Log(safmin/eps)/math.Log(dlamchB)/2))
 	safmx2 := 1 / safmn2
 	if g == 0 {
 		cs = 1
--- a/testlapack/dlartg.go
+++ b/testlapack/dlartg.go
@@ -8,6 +8,8 @@ import (
 	"math"
 	"math/rand"
 	"testing"
+
+	"github.com/gonum/floats"
 )
 
 type Dlartger interface {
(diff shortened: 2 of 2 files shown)


native/dlartg.go
```diff
@@ -17,7 +17,8 @@ import "math"
 func (impl Implementation) Dlartg(f, g float64) (cs, sn, r float64) {
 	safmin := dlamchS
 	eps := dlamchE
-	safmn2 := math.Pow(float64(dlamchB), math.Log(safmin/eps))
+	exp := int(math.Log(safmin/eps) / math.Log(float64(dlamchB)) / 2)
+	safmn2 := math.Pow(float64(dlamchB), float64(exp))
```
kortschak: `safmn2 := math.Pow(dlamchB, math.Floor(math.Log(safmin/eps) / math.Log(dlamchB) / 2)`

`dlamchB` is an untyped number, so it will be float64 for this signature.

Is there any reason not to have values like this as package global?

testlapack/dlartg.go
```diff
@@ -8,29 +8,81 @@ import (
...
 )
 
 type Dlartger interface {
 	Dlartg(f, g float64) (cs, sn, r float64)
 }
 
 func DlartgTest(t *testing.T, impl Dlartger) {
+	const (
+		tol   = 1e-14
+		safmn = 2.002083095183101e-146
```
kortschak: Can you say how these were calculated?

testlapack/dlartg.go
```diff
@@ -8,29 +8,81 @@ import (
...
+		p := rnd.Float64()
+		switch {
+		case p < 1/3.0:
+			// Huge f.
+			fHuge = true
+			f = math.Pow(10, 10-20*rnd.Float64()) * safmx
+			if rnd.Float64() < 0.5 {
+				f *= -1
+			}
+		case p < 2/3.0:
```
kortschak: Not that this matters, but do you mean `p > 2/3.0`?

native/dlartg.go
vladimir-ch: `dlamchB` is an int variable but yes, it would be nice if it were a const.
I don't know where else this value is used to make it package global. Unfortunately, this kind of floating point manipulation is mostly magic to me.

testlapack/dlartg.go
vladimir-ch: Just by printing `safmn2` and `safmx2` in Dlartg ... Should I add a comment here?

testlapack/dlartg.go
vladimir-ch: It would be probably slightly clearer if I wrote what you suggest. I will change it.

native/dlartg.go
kortschak: >`dlamchB` is an int variable but yes, it would be nice if it were a const.

I've fixed that in #258. If you allow me to merge that, the expression above will work.

testlapack/dlartg.go
kortschak: Maybe just copy-paste the expressions since they will be simpler in a few minutes.

testlapack/dlartg.go
kortschak: ... or just

```
switch rnd.Intn(3) {
case 0:
    // stuff
case 1:
    // more stuff
default:
    // other stuff
}
```

native/dlartg.go
vladimir-ch: With the simplified form it's clearer what the formula does, it's not so magical:
```
Log(safmin/eps) / Log(dlamchB) / 2 = Log_{dlamchB}(safmin/eps) / 2
                                   = Log_{dlamchB}(Sqrt(safmin/eps))
```
Using that expression as an exponent for `dlamchB`:
```
dlamchB^{Log_{dlamchB}(Sqrt(safmin/eps))} = Sqrt(safmin/eps))
```
So basically
```
safmn2 := math.Sqrt(safmin/eps))
```
I write "basically" because above I omitted the `INT` function used in Fortran. I don't know whether including it is important or not. Also note that `math.Floor` is not the same as `float64(int())` for negative numbers, which is the case here.
I cannot explain completely why they compute `safmn2` as they do. I think that the square is there to make sure that the numbers don't get too small when `f1*f1` or `g1*g1` are computed. I don't know why divide by eps, and also why go through the exponentiation of `dlamchB`. The only reason to do that seems to be the possibility to use the `INT` function whose objective is to make the exponentiation in the form `2^{negative integer}`. Again, I don't know why.

kortschak: `math.Trunc`

testlapack/dlartg.go
```diff
@@ -8,29 +8,74 @@ import (
...
 )
 
 type Dlartger interface {
 	Dlartg(f, g float64) (cs, sn, r float64)
 }
 
 func DlartgTest(t *testing.T, impl Dlartger) {
+	const (
+		tol   = 1e-14
+		safmn = 2.002083095183101e-146 // math.Pow(dlamchB, math.Floor(math.Log(safmin/eps)/math.Log(dlamchB)/2))
```
kortschak: Make these vars and then just use the expressions I think.

vladimir-ch:
The coverage is back to normal, I think that the high percentage was a glitch perhaps triggered by adding  first tests to the testlapack package. It might be that the coverage is low in general because the `testlapack` package is included in the total count of lines but is not covered by the tests. I think we should ignore the testlapack directory when calling goveralls. Unrelatedly, we should add 1.8 to the test matrix and remove the old version.

PTAL, @kortschak

testlapack/dlartg.go
vladimir-ch: But `dlamchB`, `dlamchS`, and `dlamchE` are not in this package.

native/dlartg.go
```diff
@@ -17,7 +17,7 @@ import "math"
 func (impl Implementation) Dlartg(f, g float64) (cs, sn, r float64) {
 	safmin := dlamchS
 	eps := dlamchE
-	safmn2 := math.Pow(float64(dlamchB), math.Log(safmin/eps))
+	safmn2 := math.Pow(dlamchB, math.Floor(math.Log(safmin/eps)/math.Log(dlamchB)/2))
```
vladimir-ch: We could save one call to `Log` here if we admitted that `dlamchB` is always 2.

native/dlartg.go
vladimir-ch: Oh, thanks.

testlapack/dlartg.go
kortschak: Yes, sorry. Maybe we should consider exporting those?

PR #260, opened 2017-03-16
travis: remove testlapack from coverage testing

vladimir-ch:
Adding two tests in testlapack package recently triggered the whole testlapack package to be included in coverage testing. As a result, our official coverage dropped from about 80% to about 30%. This PR removes testlapack package from coverage testing. @jonlawlor @kortschak PTAL

I also remove two unnecessary calls to go get mat64. Also, native package is tested twice, once with BLAS_LIB=gonum and once with BLAS_LIB=OpenBLAS (and then both again when doing coverage). However, I don't think that this changes the BLAS library used because in native we use blas64, its default implementation is blas/native and this is not changed by linking against OpenBLAS. Am I missing something or could we speed up the travis tests by not testing native twice?

### DIFF
--- a/.travis/linux/OpenBLAS/install.sh
+++ b/.travis/linux/OpenBLAS/install.sh
@@ -51,10 +51,9 @@ fi
 sudo cp -r ${CACHE_DIR}/* /usr/
 
 # install gonum/blas against OpenBLAS
-# fetch and install gonum/blas and gonum/matrix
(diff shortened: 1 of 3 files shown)


kortschak:
You are right that we test native twice and that that is a waste. I think it's like that just out of ease at the time the harness was written.

Issue #228, opened 2017-01-26
cgo,native: consider early checkMatrix and checkVector


kortschak:
In some cgo and native implementations, checkMatrix and checkVector occur after work length query. I have been writing them very early under the Principle of Repair, but we should be behaving consistently.

This issue is to discuss which way we should change the existing code; either to allow work length queries without checks or require all parameters be correct under all conditions.

vladimir-ch:
My intention with skipping the checks when `lwork == -1` was to make possible to call the function with nil slices, so that it is explicit in the caller that the optimal workspace size is determined only based on the other parameters, not the slice length. Since the caller may still provide that slices, currently the code is not consistent in this. Requiring all parameters be correct under all conditions is simpler and will inevitably lead to consistency. This seems like a better option.

vladimir-ch:
A case against requiring all parameters be correct under all conditions is `Dgesvd`. There we need to allocate an mxn slice just for the work query.

kortschak:
True. Similarly in `Dggsvp3` I check `tau` late. I think this is something that should not be rigid, but should happen as early as possible.

vladimir-ch:
I think we should get rid of that allocation in Dgesvd but it seems that all the functions used in the work query would have to be modified. Or do you see another way?
Speaking of allocations, there is another `make` in `Dormlq`, it would be nice to remove that too.

kortschak:
A quick look at `Dgeqrf`. If they are all like that, I think we can weaken the requirements to how they are in `Dggsvp3`. In the case of `Dgeqrf`, `tau` is not checked until after `lwork != -1` and the length of work need only be at least 1 until then.

I can go through and do those.

kortschak:
I don't see a way to avoid the allocation in `Dormlq`.

PR #262, opened 2017-03-16
native: reduce allocations in Dgesvd

kortschak:
@vladimir-ch @btracey  Please take a look.

Updates #228.

### DIFF
--- a/cgo/lapack.go
+++ b/cgo/lapack.go
@@ -859,10 +859,11 @@ func (impl Implementation) Dbdsqr(uplo blas.Uplo, n, ncvt, nru, ncc int, d, e, v
 //  [v1  v2  v3   e   d  u5]
 //
 // d, tauQ, and tauP must all have length at least min(m,n), and e must have
-// length min(m,n) - 1.
+// length min(m,n) - 1, unless lwork is -1 when there is no check except for
+// work which must have a length of at least one.
 //
 // work is temporary storage, and lwork specifies the usable memory length.
-// At minimum, lwork >= max(m,n) and this function will panic otherwise.
+// At minimum, lwork >= max(1,m,n) or be -1 and this function will panic otherwise.
 // Dgebrd is blocked decomposition, but the block size is limited
 // by the temporary space available. If lwork == -1, instead of performing Dgebrd,
 // the optimal work length will be stored into work[0].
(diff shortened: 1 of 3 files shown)


cgo/lapack.go
```diff
@@ -871,6 +871,11 @@ func (impl Implementation) Dbdsqr(uplo blas.Uplo, n, ncvt, nru, ncc int, d, e, v
 func (impl Implementation) Dgebrd(m, n int, a []float64, lda int, d, e, tauQ, tauP, work []float64, lwork int) {
 	checkMatrix(m, n, a, lda)
 	minmn := min(m, n)
+	ws := max(m, n)
+	if lwork == -1 {
+		work[0] = float64(ws)
```
vladimir-ch: If `m` and `n` are both 0, then the optimal work will also be 0. Could you please add something in respect here and in Dgebrd?

kortschak: I'm not clear what you mean. I've added a check that work is at least 1 element long when lwork is -1, but a zero length work is otherwise OK.

vladimir-ch: `work` must have always length at least 1, the reference always stores the optimum length (we don't but we should). Therefore it is strange to return work[0] = 0 if both m and n are 0 even though the callers should not assume work[0] >= 1.

kortschak: Ah, OK - the obligitory storing of lwkopt changes this. I'll fix that.

kortschak: Done.

cgo/lapack.go
```diff
@@ -859,17 +859,25 @@ func (impl Implementation) Dbdsqr(uplo blas.Uplo, n, ncvt, nru, ncc int, d, e, v
...
 // the optimal work length will be stored into work[0].
 //
 // Dgebrd is an internal routine. It is exported for testing purposes.
 func (impl Implementation) Dgebrd(m, n int, a []float64, lda int, d, e, tauQ, tauP, work []float64, lwork int) {
 	checkMatrix(m, n, a, lda)
+	if lwork == -1 {
+		if len(work) < 1 {
+			panic(badWork)
+		}
+		work[0] = float64(max(1, max(m, n)))
```
kortschak: Done. Also fixed the documentation which was understandable but wrong.

PR #263, opened 2017-03-17
Add installation instructions for OS X

jbochi:


btracey:
Thanks for the contribution.

I wonder if the better approach is to just totally rewrite the readme. We've had a lot of questions about clapack recently, and the way this is currently written emphasizes lapack. In fact, we've spent a lot of effort creating a native lapack so that people don't have to have these issues. We want clapack to be possible and useful, but we also don't want to be the support for everyone's issues with C. What do you think, @kortschak ?

kortschak:
I agree, the cgo reference documentation could go easy down the bottom as an alternative to the native implementation.

PR #265, opened 2017-03-22
native,testlapack: fix Dlatrs and increase its coverage

vladimir-ch:
Bigger PR than planned but one thing leads to another ...

If you wish I can remove the last commit and submit it separately. I reduced the indentation to make it easier for me to compare with the reference.

It would be nice to export dlattr later but I don't think it's ready.

@kortschak  @btracey  PTAL

### DIFF
--- a/native/dlatrs.go
+++ b/native/dlatrs.go
@@ -85,11 +85,11 @@ func (impl Implementation) Dlatrs(uplo blas.Uplo, trans blas.Transpose, diag bla
 	if noTrans {
 		if upper {
 			jfirst = n - 1
-			jlast = 0
+			jlast = -1
 			jinc = -1
 		} else {
 			jfirst = 0
-			jlast = n - 1
+			jlast = n
 			jinc = 1
 		}
 		// Compute the growth in A * x = b.
@@ -125,11 +125,11 @@ func (impl Implementation) Dlatrs(uplo blas.Uplo, trans blas.Transpose, diag bla
 	} else {
 		if upper {
 			jfirst = 0
-			jlast = n - 1
+			jlast = n
 			jinc = 1
 		} else {
 			jfirst = n - 1
-			jlast = 0
+			jlast = -1
 			jinc = -1
 		}
 		if tscal != 1 {
@@ -176,16 +176,16 @@ Finish:
 		if noTrans {
 			for j := jfirst; j != jlast; j += jinc {
 				xj := math.Abs(x[j])
-				var tjjs float64
+				var tjj, tjjs float64
 				if nonUnit {
 					tjjs = a[j*lda+j] * tscal
 				} else {
 					tjjs = tscal
 					if tscal == 1 {
-						break
+						goto Skip1
 					}
 				}
-				tjj := math.Abs(tjjs)
+				tjj = math.Abs(tjjs)
 				if tjj > smlnum {
 					if tjj < 1 {
 						if xj > tjj*bignum {
@@ -218,6 +218,7 @@ Finish:
 					scale = 0
 					xmax = 0
 				}
+			Skip1:
 				if xj > 1 {
 					rec := 1 / xj
 					if cnorm[j] > (bignum-xmax)*rec {
@@ -294,7 +295,7 @@ Finish:
 					} else {
 						tjjs = tscal
 						if tscal == 1 {
-							goto Out2
+							goto Skip2
 						}
 					}
 					tjj := math.Abs(tjjs)
@@ -327,7 +328,7 @@ Finish:
 				} else {
 					x[j] = x[j]/tjjs - sumj
 				}
-			Out2:
+			Skip2:
 				xmax = math.Max(xmax, math.Abs(x[j]))
 			}
 		}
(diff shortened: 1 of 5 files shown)


testlapack/matgen.go
```diff
@@ -283,6 +283,31 @@ func Dlagge(m, n, kl, ku int, d []float64, a []float64, lda int, rnd *rand.Rand,
 	// superdiagonals to ku.
 }
 
+// dlarnv fills dst with random numbers from a uniform or normal distribution
+// specified by dist:
+//  dist=1: uniform[0,1),
```
kortschak: Not that it probably matters that much, but it's not clear from the Fortran docs whether the distributions are e.g. [0,1) or (0,1).

testlapack/matgen.go
```diff
@@ -12,6 +12,15 @@ import (
 	"github.com/gonum/blas/blas64"
 )
 
+const (
```
kortschak: `dlamchE` is used elsewhere than just `matgen.go`, so these should probably be in `general.go`.

native/dlatrs.go
```diff
@@ -176,16 +176,16 @@ Finish:
 		if noTrans {
 			for j := jfirst; j != jlast; j += jinc {
 				xj := math.Abs(x[j])
-				var tjjs float64
+				var tjj, tjjs float64
```
kortschak: Why change this line?

testlapack/matgen.go
vladimir-ch: I wanted the number to be together, I'll move them to general.go

testlapack/matgen.go
vladimir-ch: Yea, it's not exactly a useful or necessary piece of information. I will change it to parentheses

native/dlatrs.go
vladimir-ch: Because the new `goto Skip1` would skip over the declaration of of `tjj`.

vladimir-ch:
> 7989362 confuses me. Can you explain what is going on there?

It actualy fixes two bugs.

One is the for loop bounds. The end condition is `j != jend` so `jend` must be one index past the last column if we want to include `jend`.

The other is that breaking out of the loop was skipping part of the loop body and the remaining iterations. I replaced it with `goto Skip1` .

testlapack/matgen.go

testlapack/matgen.go

native/dlatrs.go
kortschak: Ah, things in the spec that get so rarely exercised in my mind.

testlapack/general.go
```diff
@@ -17,8 +17,14 @@ import (
...
-// dlamchE is the machine epsilon. For IEEE this is 2^-53.
-const dlamchE = 1.0 / (1 << 53)
+const (
+	// dlamchE is the machine epsilon. For IEEE this is 2^-53.
+	dlamchE = 1.0 / (1 << 53)
+	dlamchP = 2 * dlamchE
+	dlamchB = 2
+)
+
+var dlamchS = math.Nextafter((4 / math.MaxFloat64), 0)
```
vladimir-ch: Would there be agreement to replace this with
```
const dlamchS = 2.22507385850720138309023271733240406421922e-308 // 2^{-1022}
```
? Also in native.

kortschak: I'd rather not, though that number can be expressed (horribly) as `1.0 / (1 << 256) / (1 << 256) / (1 << 256) / (1 << 254)`. So maybe use that, with the comment. At least it's easy to add the numbers together and confirm that they are 1022. Then that const can go in the block above.

testlapack/matgen.go
```diff
@@ -308,6 +317,323 @@ func dlarnv(dst []float64, dist int, rnd *rand.Rand) {
 	}
 }
 
+// dlattr generates an n×n triangular test matrix A with its properties uniquely
+// determined by imat and uplo, and returns whether A has unit diagonal. If diag
+// is blas.Unit, the diagonal elements are set so that A[k,k]=k.
+//
+// trans specifies whether the matrix A or its transpose will be used.
```
kortschak: Note that trans is not currently used or add the final transpose when asked for.

vladimir-ch: Thanks, I completely forgot about it. I will add it.

testlapack/matgen.go
```diff
@@ -308,6 +317,323 @@ func dlarnv(dst []float64, dist int, rnd *rand.Rand) {
...
+	ulp := dlamchE * dlamchB
+	smlnum := dlamchS
+	bignum := (1 - ulp) / smlnum
+
+	bi := blas64.Implementation()
+
+	switch imat {
+	default:
+		// TODO(vladimir-ch): Implement the remaining cases.
+		panic("testlapack: invalid or not implemented imat")
```
vladimir-ch: Thanks. Done.

testlapack/general.go
vladimir-ch: I changed it to const using the expression you suggested. Later we can do the same change in native.

testlapack/matgen.go

PR #269, opened 2017-03-29
native: avoid allocation in Dormlq

vladimir-ch:
@btracey @kortschak PTAL

### DIFF
--- a/native/dormlq.go
+++ b/native/dormlq.go
@@ -60,10 +60,14 @@ func (impl Implementation) Dormlq(side blas.Side, trans blas.Transpose, m, n, k 
 		return
 	}
 
-	const nbmax = 64
+	const (
+		nbmax = 64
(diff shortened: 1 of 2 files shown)


kortschak:
Is it worth sending the equivalent change to LAPACK?

vladimir-ch:
No, this is just what the reference is doing.

kortschak:
I completely missed that. When I had a look I went [here](ftp://ftp2.uk.freebsd.org/sites/ftp.netlib.org/lapack/explore-html/de/dff/dormlq_8f_source.html) instead of NETLIB (I have no idea why). It is at 3.4.2 instead of 3.7.0 and uses a temporary, `t( ldt, nbmax )`.

vladimir-ch:
Ah, now I understand where the previous code was coming from.

PR #271, opened 2017-05-14
native: remove unnecessary parameters of Dlasq5 and Dlasq6

vladimir-ch:
Are we still accepting PRs or better after the merge?


btracey:
LGTM and agree with @kortschak

